PR #780, opened 2015-03-03
Update cms-tools-vm-image.xml

pherterich:
Needs files attached as FFT

### DIFF
--- a/invenio_opendata/testsuite/data/cms/cms-tools-vm-image.xml
+++ b/invenio_opendata/testsuite/data/cms/cms-tools-vm-image.xml
@@ -43,4 +43,43 @@
       <subfield code="a">/tmp/opendata.cern.ch-fft-file-cache/cernvm-files/CMS-OpenData-1.0.0-rc7.ova</subfield>
     </datafield>
   </record>
+  <record>
+  <controlfield tag="001">249</controlfield>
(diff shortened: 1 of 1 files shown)


tiborsimko:
> Needs files attached as FFT

@pherterich From where do I get contextualisation script files to be attached as FFT?

pherterich:
They should be in your inbox. E-mail from 3rd of march titled "CODP - record 249 - Files"

tiborsimko:
@pherterich Got it, thanks. :smile:

tiborsimko:
Thanks, amended commit log, added FFT tags and files, and merged.

tiborsimko:
@pherterich @katilp Some comments to the record.  I've added all the files literally, i.e. we have three attached files:

   - `cernvm-script`
   - `cms-user-data.txt`
   - `opendata-desktop-settings`

but I think we may want to:  (1) consider renaming some files perhaps (so that scripts end with `.sh`); 
(2) check the contents (e.g. `cernvm-script` uses `https://github.com/cernvm/cernvm-config/blob/master/usr/bin/cvm2ova` which is a moving target due to `master` branch; could we take some tagged released version?); (3) check whether a file is needed (`opendata-desktop-settings` with only `[Desktop Entry]` which is kind of user dependent thing not needed for preservation?).

PR #798, opened 2015-07-15
Add glossary alts for plural terms

RaoOfPhysics:
I think this is good to go, @tiborsimko!


invenio_opendata/base/templates/helpers/text/glossary.json
```diff
@@ -1,405 +0,0 @@
-[
```
tiborsimko: WRT the removal of the old glossary file, it is still used by https://github.com/cernopendata/opendata.cern.ch/blob/master/invenio_opendata/base/views.py#L566

RaoOfPhysics: Hmm. So, what should I do at the moment?

tiborsimko: Nothing on your side, me or @pamfilos shall look at fixing this while merging your PR.

RaoOfPhysics:
@katilp: Please check the definitions for the new terms added.

katilp:
@RaoOfPhysics These are not yet visible in http://opendatadev.cern.ch/glossary, will they appear there?

RaoOfPhysics:
@katilp: I believe the PR needs to be merged before it'll appear there. For now, can you check the JSON file? It's not pretty, but…

tiborsimko:
Thanks, squashed and merged as cb0b467b6e130f346949f4ae2473be4436144b2c and c023406291d118a37da45aad751f5c290eca2208, and quick-fixed escaping in 44d40d14e4f334b16e9a638ac116df04e518091c.

PR #831, opened 2015-11-16
Update cms-derived-pattuples-ana.xml

AnxhelaDani:
I added the 264 field


tiborsimko:
Thanks, I fixed the second indicator as in PR #833, and merged as 4e36e15765f017bdcbdfd06dc07390b78018834c.

PR #834, opened 2015-11-16
Update cms-hamburg-files.xml

pherterich:
Added 264_0 c as per discussion


tiborsimko:
Thanks, squashed and merged as 9be1e6e.

Issue #820, opened 2015-10-28
New 2011 CMS records 


katilp:
How do we plan to proceed with the new records? I remember discussing whether moving away from the current scheme to something closer to CAP. Is this still an option?

Independently from that choice, from the CMS  side, we could start providing metadata for the collision primary datasets (24 in all)? Should we look how to do it in automated way, or rather do it by hand?
NB that the number of collision primary datasets will not be much larger than this for any release,
so it would not be a big problem to do it by hand. The efforts could then go in thinking how to extract the metadata for the MC (400 datasets), for which the information maybe slightly different than that of the collision data

pherterich:
With the time we have left I don't see us moving to something like CAP. @tiborsimko  might have ideas on how to automate the procedure as much as possible (primary datasets and MC, especially getting metadata out of DAS) before @AnxhelaDani and me do the rest manually. Maybe you can start thinking already about the additional information we will need for MCs and where to get it and how much texts on limitations etc. have to be created manually and then we can have a short meeting on that?

katilp:
The basic info for MC comes from DAS. I'm finding out how to extract the additional info for MC (generator settings i.e. the input parameters for the MC generator step). They are certainly extracted from two different sources: for pythia (the most commonly used generator program), they can be extracted from the CMS-internal McM-database, for other generators, they have to be extracted from the data file itself. I would like to do the latter step when the files have been transferred to eospublic, as the script that reads the event header from the file is actually a cmsRun job, which can then be used to test that everything is in order and the files are readable.

We could meet at the end of the week or early next week to discuss more in detail.

katilp:
BTW for the move to another schema: if you think that you should do it, there is no pressure from our side not to do it (unless it takes months...). I do not know whether it has impact at all, but thinking simple-minded it looks easier to make this move before we load the system with hundreds of MC records and tens of other stuff :-)

katilp:
@pherterich Could we meet to discuss?

pherterich:
@tiborsimko we ran out of sensible fields, so we suggest a local note field. After consulting the Invenio MARC bible, we'd go for 593$a generator $b Global Tag. If it's fine to conflict with the other Invenio MARC, 938 analog to 538 would be nicer, I leave the decision to you. Other suggestions are welcome, too, obviously.

tiborsimko:
Notes from live discussion: (can be spit away into separate tickets)
- data: create new collection "CMS Simulated Datasets" for MC
[...]

PR #835, opened 2015-11-17
Update cms-validated-runs.xml

AnxhelaDani:
Added the record https://github.com/cms-outreach/ispy-analyzers/blob/Run2011A/Cert_160404-180252_7TeV_ReRecoNov08_Collisions11_JSON.txt and renamed a collection #820


tiborsimko:
Thanks, cherry-picked, amended commit log message, added FFTs, and merged as aba344e and 99af466.

PR #846, opened 2015-11-23
Update cms-hamburg-files.xml

AnxhelaDani:
Added 264_0 c at record 207


tiborsimko:
Thanks, merged as 0a318fc.

Issue #847, opened 2015-11-23
Add Runs to metadata


AnxhelaDani:
Add to 942 field:
- [x] $r Run2010B
- [x] $r Run2011A

PR #849, opened 2015-11-24
Run field

AnxhelaDani:
Added to 942 field $r Run2011A
for issue #847


AnxhelaDani:
Update of 256 $e FIXME in the primary datasets metadata with the number of
events for #820

tiborsimko:
`942_r` is a new subfield.  I'll add it to the data model.

tiborsimko:
> 942_r is a new subfield. I'll add it to the data model.

Actually we call the whole field "collision energy"... not sure we want to have "run number" under the same heading.

tiborsimko:
Merged as 10a2be682e2ea462bf2920542b4ad96da016de27, b9787680bc804cf91d7bb33b555bab476b5c5e85 and 67836619dfb2dd7703afeb4676dbbbbbc49f1a55.

AnxhelaDani:
Regarding the discussion we had on Friday for the Run number, @pherterich and I suggest the 964_0 $c field, which follows the 264_0 $c (production date), as was mentioned that Run indicates a date of data collection. However, we saw that 964 and 960 (as another option) are used from Invenio as well http://demo.invenio-software.org/help/admin/howto-marc?ln=en . @tiborsimko can we find a solution for this?

AnxhelaDani:
After the discussion we had on our last meeting, @tiborsimko said that we can use the 964 field, so ```964_0 $c 
``` is the MARC field for the Run number

PR #854, opened 2015-11-26
Update cms-derived-pattuples-ana.xml

AnxhelaDani:
Added 942 $r Run2010B  - #847


tiborsimko:
Thanks, merged as 2cc2d83.

PR #870, opened 2015-12-03
Create cmssw-Run2011A.xml

AnxhelaDani:
CMSSW record for 2011


pherterich:
@tiborsimko to add a record number, all the other fixmes will be taken care of in January

PR #873, opened 2015-12-04
Update cms-validated-runs.xml

AnxhelaDani:


tiborsimko:
Thanks, merged as 20f6dae.

PR #878, opened 2015-12-08
Update cms-tools-vm-image.xml

AnxhelaDani:
Update of fields


tiborsimko:
Thanks, merged as 836c277.

PR #881, opened 2015-12-09
CMS: event display action image

### LINKED ISSUE #880: Event display image on education page
The current image here:
http://opendatadev.cern.ch/education
for the link to the event display needs to be replaced with an image from the new display

tpmccauley:
* Replace the old event image with one from the new display.

* Closes #880.


tiborsimko:
Thanks, merged as cf9cb6a.

PR #890, opened 2016-01-22
WIP CMS: create trigger2011_record.xml

AnxhelaDani:


tiborsimko:
I'll add the configuration files and merge "as is".  We can enrich the FIXME parts later.

PR #915, opened 2016-02-12
fixed the records for the derived datasets for ATLAS W-Path and Z-Path

fsocher:
I updated the records for the derived datasets for the ATLAS W-Path and Z-Path records as there were some small issues.


tiborsimko:
Note that the build failed due to XML errors. I'll cherry-pick the changes and fix the error during merge.

tiborsimko:
Merged as d3cf55a713c67148e92b77d41addb1c63e9d8c92 with fixes in e61bef93e15ba0ffc6ce143a166611215378b82b.

PR #930, opened 2016-03-03
Update event-display help info and files

### LINKED ISSUE #853: Help page update for event display
With the new event display the help page at http://opendata.cern.ch/visualise/events/CMS will have to be updated.

RaoOfPhysics:
Closes #853


tiborsimko:
Note that you deleted three images that the `visualise_events_help.html` template still uses:

- `help-small.png`
- `range-selector.png`
- `range.gif`

Will you amend the template or shall we keep these images?

RaoOfPhysics:
@tiborsimko: Those items aren't used in the file (and weren't earlier as well). I think it's ok to leave them in the documentation for now, commented out as they are, and add the images in if we use them in the future? Is that ok?

tiborsimko:
Yes, that's OK, though generally speaking, it is better to remove them altogether from the commented secions, and if we ever need to get back to those parts, we always have the version control to help us out.  

Currently, if somebody does:

```console
$ git grep -C 1 range.gif
invenio_opendata/base/templates/visualise_events_help.html-            <li>
invenio_opendata/base/templates/visualise_events_help.html:              <img src="../img/iSpy-Online-graphics/range-selector.png"> Select the energy range for the .<br /><img src="../img/iSpy-Online-graphics/range.gif">
invenio_opendata/base/templates/visualise_events_help.html-            </li>
```

then they would think there is a problem... as I did :smile:

tiborsimko:
Thanks, merged as 402caa9.

PR #937, opened 2016-03-12
Remove unnecessary horizontal scrolling on home page

1ace:
`.homepage-frame` has a `width: 100vw;`, which does not account for the with of the scrollbar, and `overflow: hidden;` seems to be ignored. This results in a few too many pixels and making the page scrollable horizontally, without having any actual content there.
I tried to quickly google for an explanation to this issue, but I didn't find any.

This could be fixed by always leaving the last 16px free, using calc: `width: calc(100vw - 16px);`
I don't know how far back you want to support old browsers, [Can I Use](http://caniuse.com/calc) might help you decide there.

_Disclaimer: I didn't go through the hoops to test this :]_

### DIFF
--- a/invenio_opendata/base/templates/index_scrollspy.html
+++ b/invenio_opendata/base/templates/index_scrollspy.html
@@ -85,7 +85,7 @@
 
     .homepage-frame {
       height: 100vh;
-      width: 100vw;
+      width: calc(100vw - 16px);
       float: left;
       color: #354654;
       overflow: hidden;


tiborsimko:
@pamfilos Can you please have a look?

tiborsimko:
@pamfilos Would you have time to look into this? Should we close this issue as per the forthcoming UI?

ioannistsanaktsidis:
Closing as not relevant anymore.

PR #976, opened 2016-03-22
CMS: fix absolute links

tiborsimko:
* Fixes absolute OPENDATADEV links that would be invalid on the QA and
  production sites.


tiborsimko:
@AnxhelaDani FYI here is the relative link style that we should be using everywhere...

PR #1064, opened 2016-04-15
CMS:link the author list to records

AnxhelaDani:
Addition of `110 $w` at:
- simulated datasets-Run2011A
- trigger path-Run2011A


tiborsimko:
Thanks, fixed the CI problem caused by unclosed subfield, and merged as c2992c424c1e69ebad9aa52f8c90885eeb5d7aeb.

PR #1068, opened 2016-04-16
Update Getting Started (Closes #969 and Closes #1015)

### LINKED ISSUE #969: Update CMS's "Getting Started" to point to Achim's di-muon examples
- [x] For 2010 data: explicitly point to the example code
- [x] For 2011 data: point to code saying that it is optimised to run on 2010 data, but can be run on 2011 data by:
    1. Changing the CMSSW version (appropriate VM version)
    2. Changing one line of code
    3. Changing the input files, in the same file as the one in ii.

The example record should also include this caveat that the code can be run on 2011 data but hasn't been optimised for it.

### LINKED ISSUE #1015: Add a mention about the use of Global tag in Getting started
We should mention the use of Global tag in Getting started.

This can be done together with #969, we need to see how to place this best in the text

"Some additional data are read from the CMS condition database when the job is running, for example, the constants for jet energy corrections. See the example in [pattuples2011] how to connect to the the condition database on cvfms. 
Note that the first time connection to the database may take long (40 mins on a 12Mbs line) as some Gb of data are loaded to the cache area of your CMS VM image. This happens before opening the event files and the event loop, and your job will produce no output during this time. This happens only the first time.

For simple analyses, where only objects needing no further data for corrections are used, you do not need to connect to the condition database. This is the case for [di-muon example]

RaoOfPhysics:
I think this addresses all the changes that need ton be made. @tiborsimko, if you can merge to `qa`, perhaps @katilp can review the changes as they would appear on the portal. If @pamfilos has added the `no-margin` style to the CSS, there should be no gap between the last and second-to-last text chunks on both pages.

### DIFF
--- a/invenio_opendata/base/templates/get_started_CMS_2010.html
+++ b/invenio_opendata/base/templates/get_started_CMS_2010.html
@@ -3,7 +3,7 @@
 <div class="gen-box">
   <h2>"I have installed the CERN Virtual Machine: now what?"</h2>
     <p>
-      To analyse CMS data collected in 2010, you need <strong>version 4.2.8</strong> of CMSSW, supported only on <strong>Scientific Linux 5</strong>. If you are unfamiliar with Linux, take a look at <a href="https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookBasicLinux">this short introduction to Linux</a>. Once you have installed the <a href="{{ url_for('VM/CMS/2010') }}">CMS-specific CERN Virtual Machine</a>, execute the following command in the terminal if you haven't done so before; it ensures that you have this version of CMSSW running:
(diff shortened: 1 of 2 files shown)


tiborsimko:
@RaoOfPhysics Thanks, merged as 593744b8db5e6c0738e09456135851d93a20fe07.

@katilp Deployed on DEV, you can have a look in the usual place:

-  http://opendatadev.cern.ch/getting-started/CMS

katilp:
@RaoOfPhysics Reading through now, it looks really nice! 

Maybe in the first example reading from AODs:
"However, there may be cases when you need to do so" -> "However, there may be cases when you can do so"
Maybe also remove "rare" from the next sentence. 

For inattentive readers, I would start the reduced dataset example by restating explicitly that this is now the second example for analysis in several steps.

katilp:
@RaoOfPhysics 

Note that the three links fo5 the github repo for 2011 are wrong, should be
https://github.com/katilp/pattuples2011
https://github.com/katilp/pattuples2011/blob/master/PAT_data_repo.py
https://github.com/katilp/OutreachExercise2011

For Global tag, I would write out the two lines from the code explicitly 
#globaltag
process.GlobalTag.connect = cms.string('sqlite_file:/cvmfs/cms-opendata-conddb.cern.ch/FT_53_LV5_AN1_RUNA.db')
process.GlobalTag.globaltag = 'FT_53_LV5_AN1::All'
(note that the lines are 46-58 for 2011)
and mention also that one has to set the symbolic link (write them out explicitly). E.g. Note that before running the job, you have to set the symbolic links to the condition data directory:
ln -sf /cvmfs/cms-opendata-conddb.cern.ch/FT_53_LV5_AN1_RUNA FT_53_LV5_AN1
ln -sf /cvmfs/cms-opendata-conddb.cern.ch/FT_53_LV5_AN1_RUNA.db FT_53_LV5_AN1_RUNA.db

katilp:
@RaoOfPhysics Ignore the previous about links and line, I was in the wrong page!!!

PR #1083, opened 2016-04-19
CMS: removal of ForwardTriggers Run2011A dataset

tiborsimko:
* Removes ForwardTriggers primary dataset record which was a technical
  dataset useful for calibration only.


tiborsimko:
@pherterich @AnxhelaDani If we have minted a DOI already, we may want to have a nice tombstone page for ForwardTriggers later perhaps...

pherterich:
Is that record 19? Because that now also looks weird in the results list...

pherterich:
Can you strip the record down just to the basic metadata and then put up a disclaimer saying that there's no data to download as it was only used internally for calibration and does not serve any research purpose or something like this?

tiborsimko:
@pherterich Thanks for the ping!  Yes, that's the record.  IIRC @AnxhelaDani has not registered it yet, so no tombstone needed.  The current visibility is just a temporary glitch due to cache invalidation... I just forced it so it should be gone now http://opendata.cern.ch/search?p=recid:19

Issue #779, opened 2015-02-27
data model: compare existing MARC tags with JSON data model


tiborsimko:
@pamfilos noticed one record containing `556 $u $y`, which is not present in the JSON data model configuration.

@tiborsimko should run comparison script to discover more like this...

pherterich:
It will be included in the JSON data model config. It is modeled on paper already and we're working on getting it from paper into the config file. Will happen next week. Should be the only field, seeing as we took the config file + my notes.

tiborsimko:
I've run the script and there are actually more "unknown" tags.  Here is the list:

- 037__a
- 041__a
- 269__a
- 269__b
- 269__c
- 300__a
- 516__y
- 538__u
- 556__u
- 556__y

This may be important if we would like to index them and/or display them properly.

@pherterich Are you going to amend `opendata.cfg` or shall I do that?

tiborsimko:
For reference, here is complete overview of which tags are used and how frequently:

```
# TAG     NB_RECORDS	# recID1 recID2 ... recID9 (example records)
005__     99 		 # 1 2 3 4 5 6 7 8 9
0247_2    58 		 # 1102 1100 1101 328 329 331 700 201 202
0247_a    58 		 # 1 2 3 4 5 6 7 8 9
037__a    1 		 # 410
041__a    11 		 # 55 330 50 59 61 57 58 410 411
100__a    57 		 # 50 52 55 56 57 58 61 101 200
100__h    1 		 # 450
100__i    1 		 # 450
100__u    6 		 # 329 330 450 61 56 410
110__a    46 		 # 1 2 3 4 5 6 7 8 9
110__g    2 		 # 53 54
245__a    99 		 # 1 2 3 4 5 6 7 8 9
246__a    17 		 # 1 2 3 4 5 6 7 8 9
256__a    62 		 # 1200 1201 1202 1203 331 212 200 101 553
256__b    47 		 # 1 2 3 4 5 6 7 8 9
256__e    52 		 # 1 2 3 4 5 6 7 8 9
256__f    47 		 # 1 2 3 4 5 6 7 8 9
256__t    62 		 # 1200 1201 1202 1203 328 331 203 204 205
260__b    87 		 # 1200 1201 1202 1203 1204 320 321 322 323
260__c    94 		 # 1200 1201 1202 1203 1102 1100 1101 203 204
264_0c    14 		 # 1 13 2 3 4 5 12 11 9
[...]

PR #1084, opened 2016-04-19
records: update of opendata.cfg

AnxhelaDani:
Addition of `520__u` and `520__w`
Deletion of `653__a` and use only 6531_a (I'll update the 4 records, where the field was included)
Addition of schema with forcing true for `964_0c`

[see also https://github.com/cernopendata/opendata.cern.ch/issues/779#issuecomment-207396704 ]


tiborsimko:
Thanks, merged as f4b02cee27eb30061326f68978165267af09496c.

PR #1087, opened 2016-04-20
CMS: MC pileup text

### LINKED ISSUE #1017: Mention pile-up for the simulated data
We could mention the pile-up in "How were these data generated"
Probably a generic statement after the processing step where the pile-up events were
added would be best. 

Note that this was mentioned in
https://github.com/cernopendata/opendata.cern.ch/issues/820#issuecomment-157777437
but we probably do not need a separate field for it.

tpmccauley:
* Add explanation of how pileup is simulated for MC records.

* Closes #1017.


katilp:
pile-up with hyphen!!!

katilp:
I think you could leave SimGeneral/MixingModule and TrueNumInteractions. If someone is reading this, they may be interested to have a look at the code and the configuration (which are both available)

tiborsimko:
Thanks, squashed and merged as 25bd3bf13492ad9578eb55456b6748ec552a1936.

tiborsimko:
@katilp @RaoOfPhysics @tpmccauley Deployed on DEV, we may want to amend the image size:

- http://opendatadev.cern.ch/about/CMS-Pileup-Simulation

PR #1096, opened 2016-04-20
Add content for Simulation-Dataset-Names (Closes #864)

### LINKED ISSUE #864: Add page on ProductionDataSetNames
From (internal link) https://twiki.cern.ch/twiki/bin/viewauth/CMS/ProductionDataSetNames

RaoOfPhysics:
I cannot overstate how much I love `pandoc`. :D

Good to go, I think, @tiborsimko!

### DIFF
--- a/invenio_opendata/base/templates/cms_simulated_dataset_names.html
+++ b/invenio_opendata/base/templates/cms_simulated_dataset_names.html
@@ -15,12 +15,222 @@
       <div class="row">
         <div class="col-sm-12">
           <h2>CMS Simulated Dataset Names</h2>
-          <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
-          tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam,
-          quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
-          consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
-          cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
-          proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>
+            <h3>What is the “simulated dataset” name?</h3>
+            <p>It’s the part of the dataset name before the second 'slash', i.e. for the dataset</p>
+            <p><kbd>/DiPhotonBorn_Pt10to25/Summer09-MC_31X_V3_7TeV_TrackingParticles-v1/GEN-SIM-RAW</kbd></p>
+            <p>the dataset name is</p>
+            <p><kbd>DiPhotonBorn_Pt10to25</kbd>.</p>
+            <h3>Convention</h3>
+            <p>CMS uses the following convention for this name:</p>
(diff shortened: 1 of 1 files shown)


tiborsimko:
Thanks, merged as 16237f8fae830af2575bd735f385a7ad7fed312e.

BTW I'd suggest to either modify `kbd` style to more cheerful colours, or else use the `<code>` markup instead, because the page looks rather sad with all those black boxes:

- http://opendatadev.cern.ch/about/CMS-Simulated-Dataset-Names

RaoOfPhysics:
@tiborsimko: Yup, it looks pretty weird on the page. Will replacing `<kbd>` with `<code>` suffice?

Also, the tables have column width set to the absolute maximum of the longest element in each column. Can this be changed in the CSS to give it a bit of padding? Looks pretty congested. @pamfilos?

katilp:
Maybe replace the example dataset name
/DiPhotonBorn_Pt10to25/Summer09-MC_31X_V3_7TeV_TrackingParticles-v1/GEN-SIM-RAW
 with a name from public data e.g.
GluGluToHToZZTo4L_M-550_7TeV-minloHJJ-pythia6-tauola/Summer11LegDR-PU_S13_START53_LV6-v1/AODSIM

tiborsimko:
Semantically speaking, the "black expressions" on that page are not really "things you type on your keyboard as input", rather "things that are code-like expressions", so `<code>` would seem more appropriate there. (As well as elsewhere in some of the other templates, but here it is striking the most.)  For the difference between `<code>` and `<kbd>`, see http://getbootstrap.com/css/#code

RaoOfPhysics:
@tiborsimko: OK, will fix it along with the edit to the example dataset name that @katilp proposes.

pamfilos:
@RaoOfPhysics for the tables you can add a "table" class on the `<table>` tag (e.g. `<table class='table'>`) and we can modify how it looks then if changes are needed

RaoOfPhysics:
@pamfilos: Can do. But shouldn't the default table CSS have some padding? Anyway, I'll do what you propose. :) Just `<table>`, correct?

tiborsimko:
Here's an example:

```
<table class="table">
<thead>
<tr><th>Foo</th><th>Bar</th><th>Baz</th></tr>
</thead>
<tbody>
<tr><td>1a</td><td>1b</td><td>1c</td></tr>
<tr><td>2a</td><td>2b</td><td>2c</td></tr>
<tr><td>3a</td><td>3b</td><td>3c</td></tr>
</tbody>
</table>
```

http://getbootstrap.com/css/#tables

RaoOfPhysics:
@tiborsimko: I've lost track of all the latest conventions. "Back in my day, you would hardcode your style in your HTML tags and you would like it!" :P

tiborsimko:
Yeah, some of the conventions may seem strange, but it's best to stick to plain Bootstrap whenever possible.  The `table` class adds some reasonable padding that should do, here is a live COD example: http://opendatadev.cern.ch/record/1700

RaoOfPhysics:
@tiborsimko: :+1: I'm happy to learn new things along the way. :)

PR #1098, opened 2016-04-21
Replace kbd with code in SimDataset Names (PR #1096)

RaoOfPhysics:


katilp:
Note a missing / in 
         <li><code>PtW-300<code> if there is a lower cut on the W pthat</li>

RaoOfPhysics:
@katilp: Thanks for the catch! Looks like it was a problem in the original edit too. :(

RaoOfPhysics:
@tiborsimko: Perhaps you can check if this looks ok and let me know if I need to make any further changes before merging. Thanks!

tiborsimko:
@RaoOfPhysics Yes, I'm always checking locally before merging :smile: The page was generally fine, so I preferred to merge and deploy yesterday, so that the content could be nailed down, treating the kbd/code suggestion as a simple further improvement...

tiborsimko:
Thanks, merged as 239cecbfc62978f61a0941aa1cf7ce68b65ef781.

PR #1111, opened 2016-04-25
CMS: create new records under Open Data Instructions

AnxhelaDani:
According to discussion with @katilp:
Two new records under collection: Open Data Instructions pointing to
"CMS Pile-up Simulation" and "CMS Simulated Dataset Names" pages,
respectively.


tiborsimko:
Note that there are duplicate record IDs...

AnxhelaDani:
@tiborsimko done, thanks!

tiborsimko:
Thanks, squashed and moved to the old file and merged into `production` as 0dec2cc2ebbf275e6529b12642f130e97075e2dc.

Issue #1125, opened 2016-06-14
CMS: Create record for the validation code of 2010 MinimumBias dataset


AnxhelaDani:
Create new record: "Validation code for 2010 MinimumBias dataset, based on track and
vertex spectra" according to Achim's instructions

Issue #1123, opened 2016-06-14
CMS: Create record for the validation code for 2010 Commissioning data sets, based on dimuon mass spectrum


AnxhelaDani:
Create record: "Validation code for 2010 Commissioning data sets, based on
dimuon mass spectrum" according to detailed instructions from Achim

PR #1126, opened 2016-06-14
CMS: Validation code for 2010 Commissioning dataset and MinimumBias d…

AnxhelaDani:
New records for the validation code of 2010 Commissioning dataset and MinimumBias dataset (addresses #1123 & #1125)


tiborsimko:
Thanks, fixed XML syntax, amended message, and merged into `production` branch as 172d300fe08ad3b8902785bdb5dfa2801b21dfe0.

Issue #1142, opened 2016-07-26
ATLAS: Create MC Samples records


AnxhelaDani:

tiborsimko:
I'll now populate EOS information...

PR #1143, opened 2016-07-26
  ATLAS: Create atlas-simulated-datasets.xml

AnxhelaDani:
* addresses #1142


tiborsimko:
There are some XML issues, I'll cherry pick and fix them locally...

tiborsimko:
I've added the record IDs from 3800 until 3840 for these records.

I see that the titles use HTML heavily, but sometimes there are opening/closing tag troubles.  Also, `<p>` are not really suitable for record titles... but I'll merge anyway and you could amend the values afterwards.

(Generally speaking, it would be better to use TeX syntax with MathJax rather than HTML syntax in for formulae in titles, but this can come later...)

Issue #1146, opened 2016-07-26
ATLAS:Create VM records


AnxhelaDani:

tiborsimko:
I'll now populate EOS information...

PR #1147, opened 2016-07-26
ATLAS:create atlas-tools-VM.xml

AnxhelaDani:
* addresses #1146


tiborsimko:
Adding record IDs 3851..3854 and moving the records into the usual `atlas-tools.xml` file for simplicity...

Issue #1179, opened 2016-08-05
Update of opendata.cfg


AnxhelaDani:
* Addition of `786__a` & `786__w`  (related record) to the data model. 

Then, as @tiborsimko  mentioned we should amend templates (see #1094)

PR #1180, opened 2016-08-05
records: data model field update (786)

AnxhelaDani:
* Addition of `786__a` & `786__w` to the data model (addresses #1179 )


tiborsimko:
Thanks, squashed and merged via #1195.

Issue #1181, opened 2016-08-05
ATLAS: amend ATLAS Derived Datasets records 


AnxhelaDani:
For better search and disambiguation purposes:
- Correction of text in `110__ $a` and `520__ $a`
- Correction of `6931_` field to `693__` according to https://github.com/cernopendata/opendata.cern.ch/issues/779#issuecomment-207483375
- Addition of `264_0 $c`, `960__ $c` , `964_0 $c` and `980__ $a`

PR #1182, opened 2016-08-05
ATLAS:update of ATLAS Derived Datasets records

AnxhelaDani:
- Correction of text in `110__ $a` and `520__ $a`
- Correction of `6931_` field to `693__`
- Addition of `264_0 $c`, `960__ $c` , `964_0 $c` and `980__ $a`

(addresses #1181 )


tiborsimko:
Thanks, merged via #1194.

BTW just a quick check whether the 2011->2012 changes were always OK? E.g. Felix mentioned that the W-path 2015 files were for 2011 recorded data. Let's double check IRL.

Issue #1183, opened 2016-08-09
Updates to the trigger descriptions of the 2010 datasets


katilp:
Following updates to the 2010 datasets (these texts should replace the current text under "How were these data selected?":
## ZeroBias dataset:  (http://opendata.cern.ch/record/13)

Since the actual ZeroBias triggers were moved to the MinimumBias dataset,
the remaining events do not contain useful information and this sample
should not be used.
## MinimumBias dataset: (http://opendata.cern.ch/record/8)

There are four categories of triggers in the MinimumBias dataset:

1) ~60% Minimum-bias-like triggers (some tagger requirements
         or random triggers)
             HLT_L1_BPTX_MinusOnly
             HLT_L1_BPTX_PlusOnly
             HLT_L1_BPTX
             HLT_L1Tech_BSC_HighMultiplicity
             HLT_ZeroBias
             HLT_Random

2) ~10% large track multiplicity triggers (three subcategories:
         >75, 85, 100 tracks) for "ridge" studies
             HLT_PixelTracks_Multiplicity100
             HLT_PixelTracks_Multiplicity85
             HLT_PixelTracks_Multiplicity70

3) ~28% triggers for exotic stopped High mass Stable
         Charged Particles (out of time, out of beam(?))
             HLT_StoppedHSCP35_v3
             HLT_StoppedHSCP20_v3
             HLT_StoppedHSCP

4) ~2%, a technical "muon" (RPC based) trigger
             HLT_L1Tech_RPC_TTU_RBst1_collisions
## Commissioning dataset: (http://opendata.cern.ch/record/2)

There are four categories of triggers in the Commissioning dataset:

1) ~30% "almost" Minimum-bias-like trigger (some tagger requirements)
             HLT_L1_BptxXOR_BscMinBiasOR

2) ~10% multi-vertex triggers with or without energy threshold (biased)
             HLT_MultiVertex6
             HLT_MultiVertex8_L1ETT60

3) ~30% isolated track triggers with some L1 jet (e.g. for taus)
             HLT_IsoTrackHB_v2
             HLT_IsoTrackHB
             HLT_IsoTrackHE_v3
             HLT_IsoTrackHE_v2
             HLT_IsoTrackHE

4) ~20% technical "muon" (DT or CSC based) triggers
             HLT_Activity_CSC
             HLT_Activity_DT
             HLT_Activity_DT_Tuned
## Mu dataset: (record http://opendata.cern.ch/record/14)

[...]

PR #1184, opened 2016-08-15
CMS:update of text in 567 for 2010 primary datasets

AnxhelaDani:
* addresses #1183


tiborsimko:
Thanks, simplified and merged via #1194.

Issue #1185, opened 2016-08-18
ALICE: Update of reconstructed data records


AnxhelaDani:
* Update reconstructed data records with more information

PR #1186, opened 2016-08-18
ALICE: Update of reconstructed data records

AnxhelaDani:
* addresses #1185


tiborsimko:
Thanks, merged via #1194.

Issue #1138, opened 2016-07-12
Luminosity calculation for CMS primary datasets 


katilp:
Need to add the instructions (somewhere) on how to compute the luminosity for primary datasets.
These instructions are available in CMS-internal pages
https://twiki.cern.ch/twiki/bin/view/CMS/TWikiLUM (general information about the luminosity group)
and https://twiki.cern.ch/twiki/bin/viewauth/CMS/LumiCalc (quite good instructions on how to use the tool for Run 1 data)
Achim will have a look and we can then extract the relevant part of the instructions.
We will need to decide how to add this on the portal (a page like getting started or an item in a collection).

tiborsimko:
Indeed, I get access denied.  @RaoOfPhysics can you please have a look and advise on creating either a template page or a documentation record?

katilp:
We can wait to see how the luminosity calculation goes with the 2011 examples. For the 2011-jets example, Matias is using a script pixelLumiCalc.py which is described in the twiki page linked above (pdf of that page attached here)

[CMS_LumiCalc.pdf](https://github.com/cernopendata/opendata.cern.ch/files/390243/CMS_LumiCalc.pdf)

katilp:
Working on the luminosity issue, two options
- use pixelLumiCalc.py (as above 1138#issuecomment-236139738) and find out what needs to be added for it to work on Open Data VM (used in Run1 analyses)
- use bril and check if it works in Open Data VM (new for Run2 analyses, should work for Run1 as well.

Prefer the brilsw option to avoid two different luminosity tools for Run1 and Run2.

Further details from the luminosity group (Zhen Xie, Jonas Daugalas):

_Bril toolkit (brilws) relies on brilconda python environment (no other CMS tools are needed).

Brilconda environment and bril toolkit are not just different options, both are needed.

External user should install bril toolkit locally. User could also install brilconda environment locally, but because brilconda is a big bundle, we suggest that users use centrally installed environment e.g. at /cvmfs/cms.cern.ch/brilconda . The purpose of step "a" is for users to set PATH to brilconda (installed somewhere) before installing/using bril toolkit.

Usage instructions can be found at http://cms-service-lumi.web.cern.ch/cms-service-lumi/brilwsdoc.html
Quickstart example to get online luminosity of run 286302 for every lumisection:
$ brilcalc lumi -r 286302 --byls_

Instructions:

[...]

PR #1241, opened 2017-05-24
CMS: addition of luminosity calculation instructions 

ArtemisLav:
- Addition of `556` (Issues & Limitations) and `567` fields based on #1138 

Signed-off-by: Artemis Lavasa artemis.lavasa@cern.ch


ArtemisLav:
The FFT files and the information about prescaled triggers still need to be added

tiborsimko:
Thanks, took only the non-lumi parts since we now have #1275, rebased against production, and merged as #1276.

PR #1256, opened 2017-05-30
Create opera-author-list.xml

### LINKED ISSUE #1254: OPERA: add author record
Add a record listing all OPERA collaboration authors.

Plug this record to the OPERA dataset records.

This is similar to how the CMS collaboration authors are done.

ArtemisLav:
- New record for OPERA collaboration author list (addresses #1254)


ArtemisLav:
- Checked all authors on HepNames to try and find as many first names as possible (some are missing there and therefore are missing from here as well)
- With regard to the spelling of last names (special characters etc.), they have been documented according to the list provided and not INSPIRE 
- Converted affiliations to match the naming convention used on INSPIRE for standardisation purposes
- For cases where it was noted that an author has now changed institutions, only the current affiliation has been included

- TBD:

  1. ask all authors to review their information to make sure it is accurate and/or provide information if it is missing

  2. adding INSPIRE IDs: a lot of the HepNames records for the authors are incomplete, i.e. do not have an ID, but we could add those that do

tiborsimko:
Thanks, added into installation procedures and merged via #1282.

Issue #1326, opened 2017-08-28
datacite: provide DOI minter utility


tiborsimko:
Plug https://github.com/inveniosoftware/datacite helper library that will assist in minting the DOIs for dataset, software and other objects.

Write a small CLI utility script that would permit to get information about already-registered objects with DataCite. Populate PID store with these already-minted DOIs from the COD2 times.

Write a small CLI utility script that would permit to register a DOI for any UUID object held by COD3.
Beware of populating PID store etc.

(For example, check Zenodo's use case, except that in our use case we don't want to mint DOI for every new uploaded record by the user, only for a set of selected records after the upload by the service manager.)

tiborsimko:
We'll need to create DataCiteXML serialisers for our data records. Note the `datacite` serialiser in `invenio-records-rest` and an example in Zenodo overlay.

ArtemisLav:
Another thing to add here WRT the XMLs is that the latest version (v4.0) of the DataCite metadata schema requires the following properties for DOI registration: `Identifier` (with mandatory `type` sub‐property), `Creator`, `Title`, `Publisher`, `PublicationYear` & `ResourceType` (with mandatory **general type description sub‐property**).

There is a controlled list of values that can be used for `resourceTypeGeneral`: 

- Audiovisual
- Collection
- Dataset
- Event
- Image
- InteractiveResource
- Model
- PhysicalObject
- Service
- Software
- Sound
- Text
- Workflow
- Other

Obviously, `Dataset` and `Software` will be used for most of the records in COD, but we have to decide which value to use for everything else as well (documentation, tools, etc.) and whether we want to store that information in the JSON. The definitions for each term can be found [here](https://schema.datacite.org/meta/kernel-4.0/doc/DataCite-MetadataKernel_v4.0.pdf) (pages 32-32).

annatrz:
Current solution implemented in #1358 supports the latest version of DataCite metadata. For resourceTypeGeneral, we use Dataset or Software, and for other types of records 'Other'.

Issue #1323, opened 2017-08-22
data model: refactor `data-policies-v1.0.0.json`


tiborsimko:
`data-policies-v1.0.0` document type was added recently as an example of how the several record data types could be added.  It should be refined based on discussions in #1313. For example, we can end up having an abstract data type `publication` that would encompass both data policy papers and analysis papers...

tiborsimko:
In the absence of alternative proposals, what about going for "documentation" following the storage policy?

ioannistsanaktsidis:
I think ```documentation``` should do the job. cc @daslerr @ArtemisLav

daslerr:
I agree with using `documentation` so as to match the storage policy.

tiborsimko:
Update `documentation` schema following up the new record model. E.g. type/subtype, etc. The goal is to use the same fields.

Or simply remove, since data policy records are part of the record fixture.

PR #1334, opened 2017-08-30
more fixtures

tiborsimko:
- 835989f14ac690f375d7c19e6516cea6cd7ef185 scripts: clean-instance.sh
- 5d05722a239d4616d86bfd654115ff70d9a83765 fixtures: news and data-policies

@ioannistsanaktsidis @annatrz Note that this PR adds a helper script to clean the instance and removes the custom `init-db.sh` that you introduced because it is not really necessary. (Better not to duplicate things in two places.) See commit notes.

@annatrz This PR adds the data policy records that @pamfilos created. The records contain true DOIs that you can use for the DataCite mining library (#1326).  Note that the record JSON will change soon, because we won't have "data policy" data type, but rather a "publication" data type (#1323), but that shouldn't really matter for the writing the minter wrapper.


scripts/clean-instance.sh
```diff
@@ -0,0 +1,109 @@
...
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+# check environment variables:
+if [ -v ${INVENIO_WEB_HOST} ]; then
+    echo "[ERROR] Please set environment variable INVENIO_WEB_HOST before runnning this script."
+    echo "[ERROR] Example: export INVENIO_WEB_HOST=192.168.50.10"
```
ioannistsanaktsidis: Should we add the default ones if one is not set ?

tiborsimko: Yes, we could.  Currently, there is `.inveniorc` that you can do `source .inveniorc` and then you have all the variables set.  In which case you don't even have to set default ones...

I'd say (1) let's have a look at the current `.inveniorc` content, to see whether the values suit your habits, or whether we amend them in some way; and then let's agree that (2a) either we shall be doing `source .inveniorc` everytime, or (3b) we'd put the values into scripts for safety.  This should also be OK to do, since the scripts were "cloned" from generic Invenio ones to custom COD ones; it would be simply a bit harder to port changes over.  Would you prefer 3a or 3b on the East End? We can again see IRL...

tiborsimko:
OK, good. Any changes needed for `.inveniorc` to match your bare-metal use case? or shall you make a PR later?

ioannistsanaktsidis:
@tiborsimko there are some small ones, but I think we can amend later, for now it is ok to go! :shipit:

PR #1336, opened 2017-08-31
glossary: adds tootip for terms

### LINKED ISSUE #1314: glossary: link terms from pages
We need to offer glossary term explanation to users when they are reading any kind of web page on the CERN Open Data portal. 

One use case scenario: a user is on the page such as `/about/CMS` where he/she sees a long content that contains also the word "muon" that happens to be a glossary term.  The word "muon" should be dot-underlined or otherwise graphically highlighted, and when the user hovers over it with the mouse, he/she should see the explanation of the term.  (and perhaps also see link to the detailed record page for the word muon (`/term/muon`) if we want).

In COD2 world, recognising glossary terms in pages, and linking to them, was done using the `zglossary`  library. In COD3, we are fully free to change the technology, now that we have glossary terms live as records. E.g. we can use the same library and prepare its JSON-needed data format from the term records.  E.g. we can create a fully new way the terms will be linked across pages. E.g. simple JS, e.g. Angular, looking at the future "live instance" use case with deposit and whatnot.

ioannistsanaktsidis:
* Closes #1314.

### DIFF
--- a/cernopendata/templates/cernopendata_theme/page.html
+++ b/cernopendata/templates/cernopendata_theme/page.html
@@ -5,6 +5,9 @@
     {%- assets "cernopendata_theme_css" %}
     <link href="{{ ASSET_URL }}" rel="stylesheet">
     {%- endassets %}
+    {%- assets "cernopendata_glossary_css" %}
+      <link href="{{ ASSET_URL }}" rel="stylesheet">
+    {%- endassets %}
     <!-- HTML5 shim and Respond.js for IE8 support of HTML5 elements and media queries -->
     <!--[if lt IE 9]>
       <script src="https://oss.maxcdn.com/html5shiv/3.7.2/html5shiv.min.js"></script>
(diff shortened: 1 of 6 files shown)


cernopendata/templates/cernopendata_theme/page.html
```diff
@@ -35,4 +38,13 @@
 
 {%- endblock body_inner %}
 
+{%- block javascript %}
+  {{super()}}
+  {% assets "cernopendata_glossary_js" %}<script src="{{ ASSET_URL }}"></script>{% endassets %}
+  <script>
+    $(document).ready(function () {
+      $('body').glossary('/static/test_data/glossary.json');
```
tiborsimko: Do you want to commit a static file for now, and create a new task on how to populate this file from glossary records? Or do you want to address it as part of this PR?

(E.g. we'd need to create this file for zglossary when the Flask app starts, or have a CLI that would compare this file to records and possibly regenerate it, etc.)

ioannistsanaktsidis: I agree this static file should be removed and taken from the one in fixtures as they are identical. For now I just put it out of simplicity so as to test the library. I think I will address the issue as part of this PR, thats why I will let WIP.

ioannistsanaktsidis:
Everything is updated. @tiborsimko please check and we can merge it.

cernopendata/templates/cernopendata_theme/page.html
tiborsimko: 👍 (And let's recall to create this file once we go into "live instance" work package that won't have full fixtures anymore.)

PR #1346, opened 2017-09-05
global: installs correct version of d3

ioannistsanaktsidis:

### DIFF
--- a/cernopendata/modules/theme/bundles.py
+++ b/cernopendata/modules/theme/bundles.py
@@ -29,6 +29,11 @@ from __future__ import absolute_import, print_function
 from flask_assets import Bundle
 from invenio_assets import NpmBundle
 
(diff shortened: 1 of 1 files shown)


cernopendata/modules/theme/bundles.py
```diff
@@ -29,6 +29,11 @@
 from flask_assets import Bundle
 from invenio_assets import NpmBundle
 
+from invenio_search_ui.bundles import js as _js
+
+# Override d3 version from search-ui
+_js.npm['d3'] = '~3.3.13'
```
tiborsimko: `npm install d3@3.3.13` should probably be added to `scripts/create-instance.sh` then.

ioannistsanaktsidis:
Closing as we pin the dependency in ```create-instance.sh```.

PR #1356, opened 2017-09-08
records: adds data policies detailed template

ioannistsanaktsidis:


cernopendata/templates/cernopendata_records_ui/records/detail.html
```diff
@@ -0,0 +1,94 @@
...
+              <a href="http://doi.org/{{id.id}}">{{id.id}}</a>
+            {% endfor %}
+          </span>
+        </div>
+      </div>
+      <p></p>
+      <div class="row">
+        <div class="col-md-12">
+          <label>Experiment:</label>
+          <span><a href="/search?p=CERN+LHC">{{record.experiment}}</a></span>
```
tiborsimko: Note that `search?p=...` is now `search?q=...` so these links wouldn't work... Please change. 

(Although we may perhaps want to implement the former to keep compatibility... let's muse in another issue.)

ioannistsanaktsidis: True didn't notice, will update!

PR #1357, opened 2017-09-08
files: introduces files in records

pamfilos:
* uses EOS as Storage

* ~~for this PR to work changes need to be done in https://github.com/inveniosoftware/invenio-files-rest/blob/aa90e5e995527262eb559c89c5676b324d9ea9ab/invenio_files_rest/storage/pyfs.py#L177 to accept kwargs~~

* ~~workaround is commenting out following lines https://github.com/inveniosoftware/invenio-xrootd/blob/master/invenio_xrootd/storage.py#L71-L72\~~

### DIFF
--- a/cernopendata/modules/fixtures/data/data-policies-v1.0.0.json
+++ b/cernopendata/modules/fixtures/data/data-policies-v1.0.0.json
@@ -24,7 +24,7 @@
       {
           "uri": "root://eospublic.cern.ch//eos/opendata/lhcb/documentation/LHCb-Data-Policy.pdf",
           "size": 254288,
-          "checksum": "5e48e8042c8edc2f828770b41e21562b5570d689"
+          "checksum": "sha1:5e48e8042c8edc2f828770b41e21562b5570d689"
       }
     ],
     "year": "2013",
--- a/scripts/create-instance.sh
+++ b/scripts/create-instance.sh
@@ -122,4 +122,5 @@ CI=true npm install
 # sphinxdoc-collect-and-build-assets-begin
 ${INVENIO_WEB_INSTANCE} collect -v
 ${INVENIO_WEB_INSTANCE} assets build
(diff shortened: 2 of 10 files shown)


cernopendata/modules/fixtures/data/data-policies-v1.0.0.json
```diff
@@ -24,7 +24,7 @@
       {
           "uri": "root://eospublic.cern.ch//eos/opendata/lhcb/documentation/LHCb-Data-Policy.pdf",
           "size": 254288,
-          "checksum": "5e48e8042c8edc2f828770b41e21562b5570d689"
+          "checksum": "md:5e48e8042c8edc2f828770b41e21562b5570d689"
```
pamfilos: ```md:``` prefix need to be set in the checksum for invenio not to through an error, as of https://github.com/inveniosoftware/invenio-files-rest/blob/b1a30ebae9021c153c1466f978726511edbafcaf/invenio_files_rest/storage/base.py#L117-L121

@tiborsimko FYI for future data

tiborsimko: Actually I used SHA1 checksum there, not MD5... this one is less safe.

tiborsimko: BTW `eos cp --checksum` command seems to use adler32 by default for checksumming...

scripts/create-instance.sh
```diff
@@ -122,4 +122,5 @@ CI=true npm install
 # sphinxdoc-collect-and-build-assets-begin
 ${INVENIO_WEB_INSTANCE} collect -v
 ${INVENIO_WEB_INSTANCE} assets build
+${INVENIO_WEB_INSTANCE} files location local var/data --default
```
tiborsimko: This is not related to asset building, and the command would fail at the build time if the DB host is not up.  We need to move this for later...

PR #1358, opened 2017-09-11
datacite: add basic cli for datacite DOIs

annatrz:

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -300,3 +300,38 @@ SQLALCHEMY_DATABASE_URI = os.environ.get(
     'APP_SQLALCHEMY_DATABASE_URI',
     'postgresql+psycopg2://localhost/cernopendata'
 )
+
+# DataCite
+# ========
+#: DataCite API - URL endpoint.
+PIDSTORE_DATACITE_URL = "https://mds.datacite.org"
+
+#: DataCite API - Disable test mode (we however use the test prefix).
+PIDSTORE_DATACITE_TESTMODE = os.environ.get(
+    "APP_PIDSTORE_DATACITE_TESTMODE",
+    False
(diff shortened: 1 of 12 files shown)


cernopendata/config.py
```diff
@@ -190,3 +190,16 @@
...
+#: DataCite API - URL endpoint.
+PIDSTORE_DATACITE_URL = "https://mds.datacite.org"
+#: DataCite API - Disable test mode (we however use the test prefix).
+PIDSTORE_DATACITE_TESTMODE = False
+#: DataCite API - Prefix for minting DOIs in (10.5072 is a test prefix).
+PIDSTORE_DATACITE_DOI_PREFIX = "10.5072"
+#: DataCite MDS username.
+PIDSTORE_DATACITE_USERNAME = "CERN.OPENDATA"
+#: DataCite MDS password.
+PIDSTORE_DATACITE_PASSWORD = "CHANGE_ME"
```
tiborsimko: I hope the password will be read from an environment variable or something? Would be good to add an example test value to `docker-compose.yml` perhaps?

cernopendata/modules/datacite/cli.py
```diff
@@ -0,0 +1,85 @@
...
+
+    print(dcp.doi_get_all())
+
+
+@datacite.command()
+@click.option(
+    '--recid',
+    help='Register record with given recid')
+@with_appcontext
+def register(recid):
```
tiborsimko: Hmm, are we going to mind record IDs for new objects in the future?  Provided we expose DOIs and stuff, we may not even have internal record IDs, only UUIDs...

annatrz: sure, will use UUIDs instead

PR #1399, opened 2017-09-12
docker: OpenShift compatible setup

diegodelemos:
Signed-off-by: Diego Rodriguez <diego.rodriguez@cern.ch>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,9 @@
 # Use CentOS7:
 FROM centos:7
 
+# Configure virtualenvwrapper root directory
+ENV WORKON_HOME=/usr/local/var
+ENV PROJECT_HOME=/usr/local/var
 # Prepare provisioning script:
 COPY scripts/provision-web.sh /tmp/
 
@@ -35,13 +38,8 @@ RUN /tmp/provision-web.sh
 WORKDIR /code
 ADD . /code
 
-# Run container as user `invenio` with UID `1000`, which should match
-# current host user in most situations:
-RUN adduser --uid 1000 invenio && \
-    chown -R invenio:invenio /code
-USER invenio
-
 # Configure CERN Open Data Portal instance:
+ENV INVENIO_INSTANCE_PATH=/usr/local/var/cernopendata/var/cernopendata-instance
 ENV INVENIO_WEB_HOST=web
 ENV INVENIO_WEB_INSTANCE=cernopendata
 ENV INVENIO_WEB_VENV=cernopendata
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -25,44 +25,42 @@
...
 ADD . /code
 
-# Run container as user `invenio` with UID `1000`, which should match
-# current host user in most situations:
-RUN adduser --uid 1000 --disabled-password --gecos '' invenio && \
-    chown -R invenio:invenio /code
-USER invenio
-
-# Configure CERN Open Data Portal instance:
-ENV INVENIO_WEB_HOST=web
```
tiborsimko: Please keep `INVENIO_...` so that kickstart scripts will work

Dockerfile
```diff
@@ -25,44 +25,42 @@
...
 
 # Add CERN Open Data Portal sources to `code` and work there:
 WORKDIR /code
 ADD . /code
 
-# Run container as user `invenio` with UID `1000`, which should match
-# current host user in most situations:
-RUN adduser --uid 1000 --disabled-password --gecos '' invenio && \
-    chown -R invenio:invenio /code
-USER invenio
```
tiborsimko: User 1000 can be removed, but then check volume mounting on Docker. E.g. files created within the container won't have the same user identity as the laptop user, which is usually a problem.  (but only in DEV setups, not in PROD setups, so for the sake of speeding up the OpenShift deployment, this could be done...)

Dockerfile
```diff
@@ -25,44 +25,42 @@
...
-RUN echo "source /usr/local/bin/virtualenvwrapper.sh" >> ~/.bashrc
-RUN echo "workon cernopendata" >> ~/.bashrc
+# Install CERN Open Data Portal
+RUN pip install .[all] \
+    && pip install jinja2-cli[yaml] \
+    && mkdir -p ${INVENIO_INSTANCE_PATH} \
+    && ${INVENIO_WEB_INSTANCE} npm \
+    && npm update && npm install --silent -g node-sass@3.8.0 clean-css@3.4.19 uglify-js@2.7.3 requirejs@2.2.0 \
+    && cd ${INVENIO_INSTANCE_PATH}/static \
+    && CI=true npm install \
```
tiborsimko: Please keep  create-instance scripts so that the instructions for bare-metal, vagrant, docker are all in the same place.

nginx/cernopendata.conf
```diff
@@ -5,7 +5,7 @@ server {
     charset utf-8;
 
     location /static {
-        root /home/invenio/.virtualenvs/cernopendata/var/cernopendata-instance;
```
tiborsimko: We can still keep the venv inside docker, for the sake of harmony with bare-metal and vagrant deployments, but simply force its location to `/usr/local/var/...`

diegodelemos:
@tiborsimko

tiborsimko:
I'm fixing the xrootd issue, so let's rebase this afterwards...

diegodelemos:
updated @tiborsimko

scripts/ensure_user_name.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
tiborsimko: Please rename to `ensure-user-name.sh`, we usually use dashes instead of underscores in shell script file names.

Please add some comments to the script that this is needed in case the container user does not have a user name such as in OpenStack etc etc.

Please `chmod u+rxw,og+rx` to clean the Travis CI.

Dockerfile
```diff
@@ -35,13 +38,8 @@ RUN /tmp/provision-web.sh
...
 ADD . /code
 
-# Run container as user `invenio` with UID `1000`, which should match
-# current host user in most situations:
-RUN adduser --uid 1000 invenio && \
-    chown -R invenio:invenio /code
-USER invenio
-
 # Configure CERN Open Data Portal instance:
+ENV INVENIO_INSTANCE_PATH=/usr/local/var/cernopendata/var/cernopendata-instance
```
tiborsimko: Since we use `INVENIO_<host>_variable` convention for the various `INVENIO_*_*` environment variables for installation, what about using `INVENIO_WEB_INSTANCE_PATH`.

PR #1335, opened 2017-08-30
templates: evaluating markdown support (WIP)

hjhsalo:
* Two implementations to support GFM-style markdown in Jinja-templates
  for opendata.cern.ch.
  One with Flask-Markdown and one with Flask-Mistune.

* For both extensions a wrapper class is used to support customization
  of the extension in invenio-style startup process of entrypoints
  and custom loaders.

* A custom pygments renderer and formatter is used for Flask-Mistune
  in order to support pygments style code highlighting.
  Out-of-the-box only HTML5 `<code>`-tag is used.

Signed-off-by: Harri Hirvonsalo <harri.hirvonsalo@cern.ch>

PR #1406, opened 2017-09-14
articles: fixtures and markdown introduction

### LINKED ISSUE #1311: content: investigate markdown display options
The source code of some static template pages (=guides) or of some dynamic record pages (=news articles) might be written in the Markdown format rather than HTML.

This ticket is about investigating available tools that could help us in displaying Markdown-authored pages in the COD3 ecosystem.

For example, check `Flask-Markdown` extensions that would permit to write Flask templates not in Jinja, but in Markdown.

For example, check `pandoc` tools that would read Article record body in Markdown from the DB and render them to users in HTML.  (Later, a user could deposit articles via UI, so having a Markdown previewer plugged in the deposit form will also be interesting... but that will come later.)

**Remember to investigate both scenarios, rendering static guides and rendering dynamic records.**

pamfilos:
* take over of PR #1335 
* moves created markdown page examples to fixture data
* removes ‘/md’ views
* updates article schema to accept md/html body content
* amends article ‘detail’ for changes
* facets: new "year" and "experiment" facet
* closes #1311


cernopendata/config.py
```diff
@@ -162,8 +162,10 @@
 RECORDS_REST_FACETS = {
```
tiborsimko: This would have been better in a PR of its own, since it is fully orthogonal to "markdown for articles"...

cernopendata/modules/fixtures/data/articles/news/news.json
```diff
@@ -6,7 +6,8 @@
       "format": "html"
     },
     "collections": [
-      {"experiment": "ATLAS"}
+      {"experiment": "ATLAS"},
```
tiborsimko: The "experiment" field will be outside of "collections" field.

cernopendata/modules/fixtures/data/articles/about/md/cms.md
```diff
@@ -0,0 +1,75 @@
+
+# About CMS
```
tiborsimko: I don't like very much the structure

- `data/articles/about/md/cms.md`

when `md` is exposed in the path as a separate subdirectory... 

Better to have `/articles/<id>.json`, `/articles/<id>-body.md`, `/articles/<id>-fig1.png` perhaps?

Can muse IRL...

cernopendata/modules/fixtures/data/articles/news/news.json
pamfilos: Left it the same way, so when all data model changes are done we can amend (not to change every time)

cernopendata/modules/fixtures/data/articles/about/md/cms.md
pamfilos: My thinking was that after a while it would become a mess, since a lot more records are coming and if with every new one there are files attached, they wouldn't be that manageable.

Lets talk IRL

PR #1415, opened 2017-09-19
records: datasets and software

tiborsimko:
- c83f8314d9382dcb3e7a374d660c07d765c03479 fixtures: datasets and software
- 1cce4883639866239170872e3860d2d2e5eb6fad records: JSON schemas for datasets and software
- 62ceb9a4ba7859503e563b4ca02c13aee273545c records: CMS datasets and software
- cf73f096834863de3acbebad2883c4a0269b2438 records: OPERA datasets and events

PR #1416, opened 2017-09-19
records: display, download files

pamfilos:


tiborsimko:
Please rebase on top of #1415 when it'll get merged... You'll have more use cases for file display there that you can check?

Issue #1447, opened 2017-09-20
content: investigate mathematic symbols support for markdown


hjhsalo:
**It should be investigated how current markdown parser/renderer can be made "mathematics-aware" utilizing at least TeX-syntax `($ ... $)` .**

Some of the records / articles in Open Data Portal utilize the `$ ... $` (inline) or `$$ ... $$` (block) TeX syntax. In the old Invenio2 -based Open Data Portal implementation this syntax was rendered usign [MathJax](https://www.mathjax.org/) Javascript library. See example [article](http://opendata.cern.ch/record/328) and it's corresponding [markup](https://github.com/cernopendata/opendata.cern.ch/commit/a38e461e094c25f44c0eafd7e3fe8ffed88145fb) for details.

Currently no MathJax support is implemented, but that shouldn't be a big problem as it is enough to add mathjax.js -file to COD3 Jinja- base template and configure it. Part of old implementation for inspiration [here](https://github.com/cernopendata/opendata.cern.ch/commit/8bd904133405ed169e4fa22a319cc16639514cde). Note that it might beneficial to make including of mathjax.js conditional so that it is only included when needed.

Bigger problem regarding for proper math symbol support is that current markdown parser/renderer implementation doesn't take TeX (or any other mathematical symbol syntax) into account when rendering markdown to HTML. 

Current situation is that if MathJax would be included it would parse everything that markdown parser outputs producing and often produce correct results. Problem arises when markdown parser gets rid of symbols or adding content inside `$ ... $` syntax as it will currently react on markdown control characters such as `* ... * ` or `_ ... _` regardless of surrounding TeX syntax.
In these cases MathJax would produce incorrect results if it is able to produce any results at all.

hjhsalo:
So far I have only investigated how this would be done for Python-Markdown based renderer.
At least these two possible extensions for Python-Markdown could be used:
* python-markdown-mathjax:
  https://github.com/mayoff/python-markdown-mathjax
* python-markdown-math:
  https://github.com/mitya57/python-markdown-math

There are also other Python-Markdown extensions that add mathematical symbol support even using different syntaxes than TeX-syntax:
* https://github.com/Python-Markdown/markdown/wiki/Third-Party-Extensions#math--latex

I will look how to implement TeX-syntax support for Mistune-based markdown renderer implementation tomorrow.

tiborsimko:
Example test case where the TeX math display can be checked:

- http://0.0.0.0/glossary/electron

[...]

PR #1457, opened 2017-09-27
Support for math content in TeX-syntax for HTML and Markdown (Mistune)

hjhsalo:
* Add MathJax.js to render content in TeX-syntax
* Make Mistune (markdown parser) and Python-Markdown compatible with TeX-syntax (`$ ... $`, `$$ ... $$`, `\begin ... \end`). Otherwise characters matching markdown control characters (e.g. `__ ... __` or `** ... **`) that reside inside TeX-content, might mess up the TeX-content.

### DIFF
--- a/cernopendata/templates/cernopendata_theme/page.html
+++ b/cernopendata/templates/cernopendata_theme/page.html
@@ -52,6 +52,7 @@
         });
     </script>
   {% endblock %}
+
   {%- block mathjax_support %}
     <script type="text/x-mathjax-config">
       MathJax.Hub.Config({
@@ -65,6 +66,8 @@
         "HTML-CSS": { availableFonts: ["TeX"] }
       });
     </script>
+
     <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS_HTML"></script>
   {% endblock %}
+
 {% endblock %}
--- a/cernopendata/templates/cernopendata_theme/page.html
+++ b/cernopendata/templates/cernopendata_theme/page.html
@@ -52,5 +52,19 @@
         });
     </script>
   {% endblock %}
+  {%- block mathjax_support %}
+    <script type="text/x-mathjax-config">
+      MathJax.Hub.Config({
+        extensions: ["tex2jax.js"],
+        jax: ["input/TeX", "output/HTML-CSS"],
+        tex2jax: {
+          inlineMath: [ ['$','$'], ["\\(","\\)"] ],
+          displayMath: [ ['$$','$$'], ["\\[","\\]"] ],
+          processEscapes: true
+        },
+        "HTML-CSS": { availableFonts: ["TeX"] }
+      });
+    </script>
+    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS_HTML"></script>
+  {% endblock %}
 {% endblock %}
-
--- a/cernopendata/modules/markdown/__init__.py
+++ b/cernopendata/modules/markdown/__init__.py
@@ -28,4 +28,4 @@ from __future__ import absolute_import, print_function
 
 from .ext import CernopendataMarkdown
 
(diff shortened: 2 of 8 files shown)


hjhsalo:
Note that this doesn't necessarily close #1447 since I haven't implemented matching functionality in Python-Markdown. We haven't yet decided which one of the two parsers to use.

ioannistsanaktsidis:
@hjhsalo is this one ready for merging ? If yes please rebase  and I will merge it.

hjhsalo:
Reviewer can use `http://127.0.0.1/md/debug` as a test page (even without running `populate-instance.sh`) or http://127.0.0.1/record/328, which contains quite a lot of "real" math content (Dataset from the ATLAS Higgs Boson Machine Learning Challenge 2014). For comparing the results see http://opendata.cern.ch/record/328

hjhsalo:
Comments and opinions about the MathJax config used currently are most welcome :)

hjhsalo:
@ioannistsanaktsidis 
Should be ready for merge now (I haven't rebased it today), but it might be good to check the code e.g. for style issues, etc.

ioannistsanaktsidis:
@hjhsalo  it works fine :) just one question, do we need mistune ? cc @tiborsimko  @pamfilos

cernopendata/templates/cernopendata_theme/page.html
```diff
@@ -52,5 +52,22 @@
...
+        tex2jax: {
+          inlineMath: [ ['$','$'], ["\\(","\\)"] ],
+          displayMath: [ ['$$','$$'], ["\\[","\\]"] ],
+          processEscapes: true
+        },
+        "HTML-CSS": { availableFonts: ["TeX"] }
+      });
+    </script>
+
+    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS_HTML"></script>
```
ioannistsanaktsidis: Keep in mind this https://www.mathjax.org/cdn-shutting-down/ .

hjhsalo: Hmm, I've seen that I thought that I had taken it into account. I will check it once again.

hjhsalo: Did I misunderstood something?
The PR uses the recommended _cdnjs.cloudflare.com_ URL

> The easiest way for most site owners will be to simply switch to another CDN provider.
> 
> We recommend cdnjs which also uses CloudFlare for delivery (and on the higher “enterprise” level!). We have been in touch with cdnjs’s maintainers and will help push future MathJax releases to cdnjs.
> 
> For example, if you have been using the latest MathJax version (v2.7.0) change
> 
> <script type="text/javascript" async
>   src="https://cdn.mathjax.org/mathjax/2.7-latest/MathJax.js?...">
> </script>
> 
> to
> 
> <script type="text/javascript" async
>   src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?...">
> </script>
>

ioannistsanaktsidis: The comment was mostly for considering adding cdn as a bundle via npm and completely get rid of the cdn. Check this https://www.npmjs.com/package/mathjax . But it is ok for now we can leave it as it is. Just to keep in mind for future.

hjhsalo: OK. Thanks.

hjhsalo:
> just one question, do we need mistune ?

Mistune provides better performance.

Python-Markdown provides better extensibility through extensions (e.g. proper and defined extension API / pattern, more extension available than for Mistune).

(I haven't made any benchmarking or comparing performance comparing between cod-mistune cod-markdown.)

I've written the included cod-mistune module to be feature-wise on par with the configuration (Python-Markdown extensions and their configuration) of cod-markdown-module.

Basically I wanted to keep Mistune around in case there are any performance issues with Python-Markdown. So far, none have been observed and probably won't be observed given that we rely on quite aggressive caching.

I would merge this one as is, and if cod-mistune needs to be removed let's remove it in a separate commit, just to keep cod-mistune in the code base in case someone find use for it in the future?

PR #1458, opened 2017-09-28
templates: Collections pages

annatrz:
Resolves collection/experiment urls to faceted page, just with short description on the top (descriptions in separate file)


annatrz:
if approved, we can add it for remaining collections, that don't require customized pages

cernopendata/modules/records/descriptions.py
```diff
@@ -0,0 +1,32 @@
...
+# You should have received a copy of the GNU General Public License
+# along with CERN Open Data Portal; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Descriptions for supporting collections pages for CERN Open Data Portal."""
```
tiborsimko: - Since these are not really describing records, but collections, perhaps they should live in `cernopendata/modules/collections/descriptions.py` rather than `cernopendata/modules/records/descriptions.py`?  I guess you did not want to create a new module, but it may be safer?

- A collection description may perhaps have a picture associated.  Shall we aim at more flexible description structure already?

tiborsimko: Something like the following perhaps:

```
collections = {
  'cms': { 'name': 'CMS',
               'description': 'Foo bar blah....',
               'logo': '/static/cms/.png'
             },
}
```

I think we don't need to keep the notion of ancestor tree, since all collections were directly attached under one of the four experiments.

Otherwise LGTM.
```

PR #1462, opened 2017-10-03
templates: use of get for attributes

ioannistsanaktsidis:
* Updates detailed template.

 * Updates data-policy schema.


ioannistsanaktsidis:
@tiborsimko  please check my schema change. There are some inconsistencies to the schema that we should muse. E.g In ```datasets-v1.0.0.json``` we have this
```
"collaboration": {
        "type": "object",
        "properties": {
          "name": {
            "type": "string"
          },
          "recid": {
            "type": "integer"
          }
        }
      },
```
and in ```data-policies-v1.0.0.json```
we have this 
```
"collaboration": {
      "type": "string",
      "title": "Collaboration"
    },
```
Which one is the correct to use ?

tiborsimko:
I'd say we should take the datasets as the example... `data-policies` was a previous attempt that will be transformed into `documentation`; we could take it as part of https://github.com/cernopendata/opendata.cern.ch/issues/1323.

PR #1581, opened 2017-11-02
 records: amendments to type, formats, collision information

### LINKED ISSUE #1580: records: amend data
As discussed with @katilp @ArtemisLav:

- ALICE reconstructed -> collision data
- OPERA Event Data -> Derived
- file type should be richer to have IG, AODSIM
- type Environment / Virtual Machine + contextualisation scripts + condition data
- Configuration -> Supplementaries
- CMS list of validated runs -> Supplementaries
- HLT trigger, L1 trigger -> Supplementaries
- collision_information: enrich type and use "pp" and "PbPb" for ALICE when appropriate
- check display of HTML fields

tiborsimko:
* Amends values of record type, distribution formats and collision
      information. (closes #1580)
    
### DIFF
--- a/cernopendata/modules/fixtures/data/records/alice-derived-datasets.json
+++ b/cernopendata/modules/fixtures/data/records/alice-derived-datasets.json
@@ -99,7 +99,7 @@
   "type": {
     "primary": "Dataset", 
     "secondary": [
-      "Derived Data"
+      "Derived"
     ]
   }
 }
--- a/cernopendata/modules/fixtures/data/records/alice-learning-resources.json
+++ b/cernopendata/modules/fixtures/data/records/alice-learning-resources.json
@@ -10,6 +10,11 @@
   "collections": [
     "ALICE-Learning-Resources"
   ], 
+  "distribution": {
+    "formats": [
+      "jpg"
+    ]
+  }, 
   "experiment": "ALICE", 
   "keywords": [
     "masterclasses", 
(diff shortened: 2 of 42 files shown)


tiborsimko:
@ArtemisLav @katilp Here is how the (expanded) facets look like with this change:

facets

ArtemisLav:
I still think `Supplementaries` is odd and would prefer `Supplementary`.
Also, I'm now noticing that `Condition` looks a bit weird when it is by itself like that.

tiborsimko:
@katilp `*.configfile` -> `*.py` is now done. Do you have any preference regarding `Supplementaries` vs `Supplementary` and regarding `Condition`, as highlighted above by @ArtemisLav?

katilp:
@tiborsimko No strong opinions wrt Supplementaries/Supplementary with a light preference for the plural, but I'm happy with both( @RaoOfPhysics suggestions? Requirements: one word, noun, meaning supplementary/additional stuff needed for research use of data).

For Condition, can you fit Condition data or db? Not one word though....

tiborsimko:
OK, let's go with plural for now, and display the state on DEV, so that people can test for themselves how things look like with faceting.  I'll open another issue with the background information about prettifying the values; e.g. if we have two words, then we might also want to change "trigger" and other such "derived" values that may read funny when standalone. Since this is only value update, without any structural changes, it'll be easy to update values afterwards.

tiborsimko:
> `Condition`

Regarding "Condition", I see many people refer to "Condition Data" as "Conditions Database", so we could use "Conditions" in place of "Condition", which would look nicer perhaps.

Issue #1384, opened 2017-09-11
templates: /collection/xxx


tiborsimko:
Once we have the demo data, check that all collection pages reply successfully with the same or similar content as the production site.

This will take a while...

- [ ] /collection/
- [x] /collection/ALICE
- [x] /collection/ALICE-Derived-Datasets
- [ ] /collection/ALICE-Learning-Resources
- [x] /collection/ALICE-Reconstructed-Data
- [x] /collection/ALICE-Tools
- [x] /collection/ATLAS
- [x] /collection/ATLAS-Derived-Datasets
- [ ] /collection/ATLAS-Higgs-Challenge-2014
- [x] /collection/ATLAS-Learning-Resources
- [x] /collection/ATLAS-Simulated-Datasets
- [x] /collection/ATLAS-Tools
- [x] /collection/Author-Lists
- [x] /collection/CMS
- [x] /collection/CMS-Condition-Data
- [x] /collection/CMS-Configuration-Files
- [x] /collection/CMS-Derived-Datasets
- [ ] /collection/CMS-Learning-Resources
- [x] /collection/CMS-Luminosity-Information
- [x] /collection/CMS-Open-Data-Instructions
- [x] /collection/CMS-Primary-Datasets
- [x] /collection/CMS-Simulated-Datasets
- [ ] /collection/CMS-Tools
- [ ] /collection/CMS-Trigger-Information
- [x] /collection/CMS-Validated-Runs
- [x] /collection/CMS-Validation-Utilities
- [x] /collection/Data-Policies
- [x] /collection/LHCb
- [x] /collection/LHCb-Derived-Datasets
- [x] /collection/LHCb-Learning-Resources
- [x] /collection/LHCb-Tools
- [ ] /collection/OPERA
- [ ] /collection/OPERA-Detector-Events
- [ ] /collection/OPERA-Electronic-Detector-Datasets
- [ ] /collection/OPERA-Emulsion-Detector-Datasets

tiborsimko:
@annatrz What is the meaning of the checkboxes above?  The collections that are checked do not seem to work yet...

annatrz:
All of them works with lowercase letters urls e.g. /collection/alice, by redirecting to faceted search page. Pages like LHCb-Derived-Datasets, where assuming that this will be a subtype of record, but now with changes in data model, it's probably not valid anymore..

tiborsimko:
Even with lowercasing, some collections work, like:

[...]

Issue #865, opened 2015-12-02
facets: new "year" facet and "category" facet


tiborsimko:
Introduce new facets:
-  "year" e.g. 2010, 2011
-  "category" e.g. "SM Inclusive".

to ease the browsing.

The category facet is used for CMS Simulated Datasets for now, but we can populate category field for other collections later too.

tiborsimko:
Possible new facets: 
- "keywords" (e.g. vector boson, drell-yan) 
- "formats" (e.g. ig, csv, root)

pherterich:
for format, I suggest MARC 347$b. @tiborsimko should we add this info manually or is there a smart way how you can fill it automatically?

tiborsimko:
> for format, I suggest MARC 347$b

What about using `256` that we already "misuse" for other file related information?

> is there a smart way how you can fill it automatically

Can be scripted for most use cases...

pherterich:
Well, if there's a proper MARC field, I prefer to use that. Also because I wouldn't know which subfield to use as f is taken for files already.

tiborsimko:
The search UI and facets will be addressed in Invenio 3 based overlay.

suenjedt:
Can I close this @tiborsimko - because we get facets with Invenio 3 anyway? The search/facets UI need deeper research (and seperate issues anyway).

tiborsimko:
We get many faceting options Invenio 3, this issue is about deciding which facets to offer to users.  Amended milestone as we are just about to attack this...

sefeg:
Did you already prepare a list of different facets that we have to decide on?

I'm thinking that there is a facet box that refers to the **record type** and should be used for the search result page and for all the resource pages in the mockup.

Another box will be about the **collections** (ALICE, ATLAS, ..) and will probably be used on most pages where facets apply.

Also the **run** (2010, 2011, 2012) represents probably a box that will be present on most of those pages with facets.

The facets for the dataset pages likely entail further information like the **data type** (Primary, Derived, Simulated)

So, those are some of my initial thoughts, also looking at the mockup.

sefeg:
Also, based on the HST 2017 feedback ( #1292 ): on the dataset pages, file types could be of interest: 

> add a file type as a search key (i.e csv, ig, root, AOD, AODSIM)

tiborsimko:
See also:
- fdbcf39a41a6f1669921b73db768d0f9603781ef
[...]

PR #1591, opened 2017-11-06
records: add more facets, fix collections pages

### LINKED ISSUE #1593: search: rich facets
Introduce new facets:

- data-taking year, e.g. 2010, 2011
- topic category, e.g. "SM Inclusive".
- collision type, e.g. pp, PbPb
- collision energy, e.g. 3.5TeV, 7TeV
- keyword, e.g. masterclasses

This will permit to user-test various options and retain the useful ones.

(stemmed from #865)

annatrz:
* Fix collection pages to work with collections field #1384 
* Fix year facet, to work with date_created field #865 
* Add collision type, collision energy, keywords and topic_category facets closes #1593 


tiborsimko:
I see the following problem when trying locally. Try empty search, you get 3005 hits, then select "csv" from "file type" facet, you get 834 hits.  So far so good; this is the expected behaviour. Now try the same with new facets e.g. collision energy and 8TeV: the number of hits won't change and remains on 3005 hits, while it should go down to 50... Can you please have a look?

annatrz:
fixed now

PR #1597, opened 2017-11-07
search: display searched parameter

### LINKED ISSUE #1583: search: display searched parameter
On COD3, if you search in the homepage search box for something, say `configuration file`, then on the search results page the original searched query string is not displayed at all:

- http://opendata-dev.web.cern.ch/search?page=1&size=20&q=configuration%20file

This is not good because people cannot see to which query the facets correspond, they cannot amend the originally searched text to make refinements, etc.

The query string should be ideally kept displayed in the search box, so that it is amendable.

annatrz:
* Closes #1583


ioannistsanaktsidis:
Rebase and then I merge ;)

PR #1602, opened 2017-11-07
cli: add option to load specific files

### LINKED ISSUE #1594: fixtures: load certain files only
In COD3, we currently have an option to load all fixtures of given type, say: 

```
$ cernopendata fixtures records --skip-files
```

It would be useful to add a CLI option to load only certain files specified in the `-f` argument:

```console
    -f cernopendata/modules/fixtures/data/records/cms-hamburg-files.json
    -f cernopendata/modules/fixtures/data/records/cms-l1-trigger-information-Run2011A.json
    -f /tmp/something
    -f /tmp/somethingelse
``` 

as we used to have in COD2.

That is, if `-f` is present, load these files only; if `-f` is missing, load everything.

diegodelemos:
* (closes #1594).

Signed-off-by: Diego Rodriguez <diego.rodriguez@cern.ch>

### DIFF
--- a/cernopendata/modules/fixtures/cli.py
+++ b/cernopendata/modules/fixtures/cli.py
@@ -51,8 +51,12 @@ def fixtures():
 @fixtures.command()
 @click.option('--skip-files', is_flag=True, default=False,
               help='Skip loading of files')
+@click.option('files', '--file', '-f', multiple=True,
+              type=click.Path(exists=True),
+              help='Path to the file(s) to be loaded. If not provided, all'
+                   'files will be loaded')
 @with_appcontext
-def records(skip_files):
+def records(skip_files, files):
     """Load all records."""
     from invenio_db import db
     from invenio_records_files.api import Record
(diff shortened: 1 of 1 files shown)


cernopendata/modules/fixtures/cli.py
```diff
@@ -51,9 +51,22 @@ def fixtures():
 @fixtures.command()
 @click.option('--skip-files', is_flag=True, default=False,
               help='Skip loading of files')
+@click.option('files', '--file', '-f', multiple=True,
+              type=click.Path(exists=True),
+              help='Path to the file(s) to be loaded. If not provided, all'
+                   'files will be loaded')
+@click.option('--profile', is_flag=True,
+              help='Outputs cProfile stats to stdout.')
```
tiborsimko: Output instead of outputs; we use declarative language in help.  Could be simplified to "Output profiling information".

tiborsimko:
(Also, it is better to use `fixtures: ...` instead of `cli: ..` in the commit log header message, in order to signal the COD3 module; see `git log --format=oneline -- cernopendata/modules/fixtures/cli.py` for past practices.)

PR #1665, opened 2017-11-08
detail: updates record detail pages

ioannistsanaktsidis:


ioannistsanaktsidis:
This one adds all the basic components of detailed pages. So as we can address all the remaining tickets on top of that.

PR #1669, opened 2017-11-08
detail: fixes more detail pages

### LINKED ISSUE #1650: records: rich metadata display for atlas-higgs-challenge-2014.json
Ensure that the records from atlas-higgs-challenge-2014.json are displayed in a complete manner in
both brief and detailed formats.

Compare the COD2 and COD3 display for several representative records from the
file, for example:

- COD2 http://opendata.cern.ch/record/328
- COD3 http://0.0.0.0/record/328

All metadata fields displayed in COD2 should be displayed also in COD3 in
roughly the same order and the same style.

Don't forget to check also the brief formats as the records are displayed in
search.

Update both brief and detailed record templates as necessary to display all the
fields:

- `cernopendata/static/templates/search/briefs/record.html`
- `cernopendata/templates/cernopendata_records_ui/records/record_detail.html`

Assume that fields may be optional; use `record.get(somefield, 'somedefault')`
technique.

When the record field contains HTML, it is safe to assume that the contents is
safe to display in Jinja.

ioannistsanaktsidis:
* Closes #1650.

### DIFF
--- a/cernopendata/modules/theme/views.py
+++ b/cernopendata/modules/theme/views.py
@@ -41,6 +41,10 @@ def get_record_title(recid):
     """Fetches record title by id."""
     from invenio_records.api import Record
     from invenio_pidstore.models import PersistentIdentifier
-    pid = PersistentIdentifier.get('recid', recid)
+    from invenio_pidstore.errors import PIDDoesNotExistError
+    try:
+        pid = PersistentIdentifier.get('recid', recid)
+    except PIDDoesNotExistError:
+        return None
     record = Record.get_record(pid.object_uuid)
     return record.get('title', '')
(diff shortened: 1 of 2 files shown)


ioannistsanaktsidis:
@tiborsimko  LGTM

tiborsimko:
N.B. The MathJax display part will have to be addressed separately; I'll create an issue.  See also https://github.com/cernopendata/opendata.cern.ch/issues/1447.

tiborsimko:
One more thing: we may need to change the order of boxes, for example http://opendata.cern.ch/record/352, see "Use with" and "How can you use this?" positions on COD2 and COD3. CC @ioannistsanaktsidis

ioannistsanaktsidis:
This could easily be fixed if we have the records splitted like /dataset, /software etc. Now with one general template it will be a bit ugly codewise. Lets check if it is really an issue for the users and if yes we can work something around.

tiborsimko:
@ioannistsanaktsidis The `use_with` property is used essentially in software-like records only, so if you place it in the correct position in the common Jinja template,  then the display of non-software-like records, such as datasets, won't be affected. 

```
   if record.get('title'):
      # title box
   if record.get('use_with'):
      # use_with box
   if record.get('usage')
      # usage box
```

One template may be enough, provided we use the same field to hold the same information for all the various data types, which should be the case.

Schematically speaking :)

PR #1683, opened 2017-11-09
detail: fixes more detailed pages

### LINKED ISSUE #1627: records: rich metadata display for lhcb-learning-resources.json
Ensure that the records from lhcb-learning-resources.json are displayed in a complete manner in
both brief and detailed formats.

Compare the COD2 and COD3 display for several representative records from the
file, for example:

- COD2 http://opendata.cern.ch/record/41
- COD3 http://0.0.0.0/record/41

All metadata fields displayed in COD2 should be displayed also in COD3 in
roughly the same order and the same style.

Don't forget to check also the brief formats as the records are displayed in
search.

Update both brief and detailed record templates as necessary to display all the
fields:

- `cernopendata/static/templates/search/briefs/record.html`
- `cernopendata/templates/cernopendata_records_ui/records/record_detail.html`

Assume that fields may be optional; use `record.get(somefield, 'somedefault')`
technique.

When the record field contains HTML, it is safe to assume that the contents is
safe to display in Jinja.

### LINKED ISSUE #1628: records: rich metadata display for alice-learning-resources.json
Ensure that the records from alice-learning-resources.json are displayed in a complete manner in
both brief and detailed formats.

Compare the COD2 and COD3 display for several representative records from the
file, for example:

- COD2 http://opendata.cern.ch/record/40
- COD3 http://0.0.0.0/record/40

All metadata fields displayed in COD2 should be displayed also in COD3 in
roughly the same order and the same style.

Don't forget to check also the brief formats as the records are displayed in
search.

Update both brief and detailed record templates as necessary to display all the
fields:

- `cernopendata/static/templates/search/briefs/record.html`
- `cernopendata/templates/cernopendata_records_ui/records/record_detail.html`

Assume that fields may be optional; use `record.get(somefield, 'somedefault')`
technique.

When the record field contains HTML, it is safe to assume that the contents is
safe to display in Jinja.

ioannistsanaktsidis:
* Closes #1627.

 * Closes #1628.

### DIFF
--- a/cernopendata/templates/cernopendata_records_ui/terms/detail.html
+++ b/cernopendata/templates/cernopendata_records_ui/terms/detail.html
@@ -14,7 +14,7 @@
             <div class="row">
               <div class="col-md-12">
                 {% for exp in record.experiment %}
-                <span class="badge badge-success">{{exp.name}}</span>
+                <a href="/search?experiment={{exp.name}}"><span class="badge badge-secondary">{{exp.name}}</span></a>
                 {% endfor %}
               </div>
             </div>
--- a/cernopendata/templates/cernopendata_records_ui/records/detail.html
+++ b/cernopendata/templates/cernopendata_records_ui/records/detail.html
@@ -87,7 +87,17 @@
                 <div class="col-md-12">
                   <h3>Disclaimer</h3>
                   {% if record.license %}
+                  {% if record.license.attribution != 'CC0'%}
                   <p>{{record.license.attribution}}</p>
+                  {% else %}
+                  <p>The open data are released under the <a href="https://creativecommons.org/publicdomain/zero/1.0/">Creative
+                    Commons CC0 waiver</a>. Neither {{record.experiment}} nor CERN endorse any
+                    works, scientific or otherwise, produced using these data.
+                    All releases will have a unique DOI that you are requested to cite in any applications or
+                    publications.
+                  </p>
+                  <img src="{{url_for('static', filename='img/cc-zero.svg')}}">
+                  {% endif %}
                   {% else %}
                   <p>The open data are released under the <a href="https://creativecommons.org/publicdomain/zero/1.0/">Creative
                     Commons CC0 waiver</a>. Neither {{record.experiment}} nor CERN endorse any
--- a/cernopendata/templates/cernopendata_records_ui/records/record_detail.html
+++ b/cernopendata/templates/cernopendata_records_ui/records/record_detail.html
@@ -79,6 +79,14 @@
                 <label>Accelerator:</label>
                 <span><a href="/search?q={{record.accelerator}}">{{record.accelerator}}</a></span>
                 {% endif %}
+                {% if record.relations %}
+                {% for relation in record.relations %}
+                {% if relation.type == 'isChildOf' %}
+                <label>Parent Dataset:</label>
+                <span><a href="/record/{{relation.recid}}">{{relation.title}}</a></span>
+                {% endif %}
+                {% endfor %}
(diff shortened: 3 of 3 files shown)


tiborsimko:
The order of boxes is not right as the last time... But the most visible thing is "How were these data selected?" which is displayed also for non-data. Example: http://0.0.0.0/record/250.  We should alter the alter heading to something more neutral...  (yeah, if dataset then ... otherwise <neutral term>)

tiborsimko:
... the rest of the detailed record page changes in this PR are otherwise OK.

But in the commit log you mention also `closes #3500`, but there is no such issue.  Which one did you meant?

cernopendata/templates/cernopendata_records_ui/terms/detail.html
```diff
@@ -14,7 +14,7 @@ <h2 class="d-inline">
             <div class="row">
               <div class="col-md-12">
                 {% for exp in record.experiment %}
-                <span class="badge badge-success">{{exp.name}}</span>
+                <a href="/search?experiment={{exp.name}}"><span class="badge badge-secondary">{{exp.name}}</span></a>
```
tiborsimko: Hmm, do we want to remain in the glossary context for these searchable links? Wondering because the other glossary labels work that way... If so, we would have to select also the type glossary for the links. CC @sefeg @daslerr

sefeg: I don't fully understand the question

tiborsimko: Imagine you are viewing http://opendata-dev.web.cern.ch/glossary/primary and you click on the "trigger" button (which leads to another fellow glossary term). What would you expect when you click on "cms"?  See only CMS-related glossary terms, or see everything we have that is CMS related?  

The current PR  would bring you to: 
- http://opendata-dev.web.cern.ch/search?page=1&size=20&experiment=CMS
representing the latter option.

It could also bring you to:
- http://opendata-dev.web.cern.ch/search?page=1&size=20&experiment=CMS&type=Glossary
representing the former option. (after we fix the query...)

Basically, this is about expected label click preferences... Have you seen any preferences in your user tests?

ioannistsanaktsidis:
@tiborsimko +1 for making the text general for all the headings. I think that this can be easily done so as to avoid  all the checks in the templates. The 3500 is indeed wrong. This was the number of the record, the issue that is closing is #1622. I updated the commit message. For the glossary I believe since when you open a detailed record you are still in the search page that the labels should lead in the general one e.g experiment=CMS.

tiborsimko:
`#1622` checked as well and looks good 😄

So it's only the order of boxes, the headings, and the possible glossary link preference to be seen... (where people are not seeing being "in the search page" exactly, because they are in `http://0.0.0.0/glossary/primary`...)

ioannistsanaktsidis:
@tiborsimko  perfect! I guess the remaining parts can be checked in a separate issue when we have finalised all the detail pages content. LGTM this one and after we finish all the record detail pages we create a PR that addresses your comments :)

cernopendata/templates/cernopendata_records_ui/terms/detail.html
sefeg: Ah alright, I was confused with the term "link", as I rather see them as "tags". And since they are designed as tags, the expectation should be to see all resources related to it. So, I am inclined to agree to the way the current PR deals with the issue.

tiborsimko:
Yes, the "order of boxes" can be finalised later... Just don't forget to keep track of examples that were problematic.

PR #1731, opened 2017-11-10
records: remove experiment keyword values

### LINKED ISSUE #1599: records: filter away some keywords values
The keyword facet looks like this:

keywords

Some values here are not really belonging to the keywords field, such as experiments.

We could filter them away during COD2 -> COD3 record migration.

We could also profit from the migration to enrich this field programmatically, if we discuss how.

CC @ArtemisLav @katilp

tiborsimko:
* Removes keyword values that are identical with experiment names.
      (closes #1599)
    

tiborsimko:
Note: built on top of #1730.

Issue #1392, opened 2017-09-11
templates: /VM/LHCb


tiborsimko:
Implement http://opendata.cern.ch/VM/LHCb

- case 3 - Collection. Collection of Articles about individual VM releases from the specified experiment
- case 1 - Record (Article)
- Body in Markdown

Issue #1391, opened 2017-09-11
templates: /VM/ALICE


tiborsimko:
Implement http://opendata.cern.ch/VM/ALICE

- case 3 - Collection. Collection of Articles about individual VM releases from the specified experiment
- case 1 - Record (Article)
- Body in Markdown

Issue #1366, opened 2017-09-11
templates: /getting-started/lhcb


tiborsimko:
Implement http://opendatadev.cern.ch/getting-started/LHCb

- case 4 - Alias to Faceted Search
- Each getting started page will become an Article and will be tagged with experiment and year.
- Port body to Markdown.

annatrz:
The link is hitting the faceted page now, implemented in #1442, only markdown needs to be done

Issue #1664, opened 2017-11-08
nginx: lowercase URLs


tiborsimko:
It will be useful to ignore the case in the URLs, see https://github.com/cernopendata/opendata.cern.ch/issues/1292.  So the following URLs should all lead to the same page:

- http://opendata.cern.ch/collection/CMS-Learning-Resources
- http://opendata.cern.ch/collection/cms-learning-resources
- http://opendata.cern.ch/collection/CMS-LeARniNG-ResOURCes

We can achieve it on the `nginx` side using `mod_rewrite` like technique to lowercase all incoming URLs.

(Beware of attached files though...)

tiborsimko:
Beware No.1: If we are going to make a convention that we use only lowercase-named files, then we have to rename some of our existing files (such as experiment logos) and also change the references to them across the source code everywhere:

```console
$ git ls-files | grep static | grep '[A-Z]' 
cernopendata/modules/pages/static/img/resources/VMs.jpg
cernopendata/modules/theme/static/img/ALICE.gif
cernopendata/modules/theme/static/img/ATLAS.gif
cernopendata/modules/theme/static/img/CMS.gif
cernopendata/modules/theme/static/img/LHCb.gif
cernopendata/modules/theme/static/img/OPERA.gif
cernopendata/static/js/components/resultsBrief.js
cernopendata/static/test_data/MuRun2010B_0.csv
cernopendata/static/test_data/dimuon-Jpsi.csv
```

In this case the nginx-lowering can remain very simple..

tiborsimko:
Beware No.2: For user-contributed files, we have URLs like `/record/NNN/files/Tau.ig`.  Here, EOS distinguishes between uppercase and lowercase names.  So we have to be careful not to lowercase any URL that starts with `/record/...`.  In this way, we'll have a nice setup where "our own" files are all lowercase, and "user-contributed files" that may be uppercase are only in certain URLs.  So the nginx-lowercasing rules remain reasonably simple.

PR #1743, opened 2017-11-13
templates: VM and getting_started pages

atrisovic:
(closes: #1391)
(closes: #1392)
(closes: #1365)
(closes: #1366)


tiborsimko:
Have you checked whether the URLs that are closed by this PR are returning something?

Example: if an issue was about porting `/VM/ALICE`, and the new URL is `/articles/virtual-machines-alice`, then we should make sure that the Flask application answers *something* on `/vm/alice`.  (Lowercase, since the capitalisation will be handled elsewhere, see #1664.) 

I know that @annatrz did some redirections in the past to the search results... We just need to make sure that for closed pages, not only the Markdown is done, but also those redirections... otherwise we would loose track.  Should be quick... 

(BTW it's better not to redirect to `/search`, but to the targeted `/articles/...` itself...)

ioannistsanaktsidis:
```/vm/alice``` is not returning anything. Just checked the new urls themselfs. I guess we need to introduce the ```/vm/...```

tiborsimko:
Yes, that was part of the purpose of this issue 😉 besides markdown porting.

@atrisovic PLease see for example `cms_pileup_simulation()` defined in `cernopendata/modules/pages/views.py`.  You can do similar for the other URLs that were closed by this PR...

PR #1747, opened 2017-11-13
nginx: rewrite non-lowercase URLs to lowercase

### LINKED ISSUE #1664: nginx: lowercase URLs

hjhsalo:
* Rewrite non-lowercase URLs to lowercase.
  Supports excluding of specific path(s) from lowercasing.
  Doesn't update URL displayed in browser.
* Currently URLs starting with `/files/`, `/record/`, `/static/`
  and `/eos/` are not lowercased.
  (closes #1664)


hjhsalo:
Includes two implementations.

Currently active one which rewrites URLs to lowercase but doesn't update URL displayed in browser.
```
server {
  location ~ [A-Z] {
    perl 'sub { my $r = shift; $r->internal_redirect(lc($r->uri)); }';
  }
```
Alternative one which rewrites URLs to lowercase and updates URL displayed in browser (currently commented out). 
This one is (much?) less efficient as HTTP 301 Redirect + an additional HTTP GET request is required to update the URL displayed by browser, but usable in case we need to make the URL update.
```
perl_set $uri_lowercase 'sub {
    my $r = shift;
    my $uri = $r->uri;
    $uri = lc($uri);
    return $uri;
  }';

server {
  location ~ [A-Z] {
    if ( $uri != $uri_lowercase ) {
      rewrite . http://$host$uri_lowercase permanent;
    }
  }
  ...
```

nginx/Dockerfile
```diff
@@ -22,6 +22,9 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
-FROM nginx
+FROM nginx:perl
 RUN rm /etc/nginx/conf.d/default.conf
 ADD cernopendata.conf /etc/nginx/conf.d/
+
+# Need to include perl-module because of uri lowercasing implementation
+CMD ["nginx", "-g", "daemon off; load_module /etc/nginx/modules/ngx_http_perl_module-debug.so;"]
```
tiborsimko: Do we need `debug` here?

nginx/cernopendata.conf
```diff
@@ -1,9 +1,40 @@
-server {
 
+# Rewrites uri to lowercase
```
tiborsimko: This comment is perhaps not necessary, since the same implementation is down below?

(You could have both comments in the PR, and keep only the final code in the config file, to make the final committed code very short...)

nginx/Dockerfile
hjhsalo: Fixed

nginx/cernopendata.conf
hjhsalo: Removed the alternative implementation from the commit.

hjhsalo:
Updated PR according to new comments in #1664.
Tested just with `curl` and checking the URI transformation from nginx logs:
```
$ curl 127.0.0.1/files/TEST  --> /files/TEST
$ curl 127.0.0.1/files/TesT  --> /files/TesT
$ curl 127.0.0.1/file/TEST   --> /file/test
$ curl 127.0.0.1/filesa/TesT --> /filesa/test
$ curl 127.0.0.1/files./TEST --> /files./test
```

ioannistsanaktsidis:
@tiborsimko is this one ok ? Should I merge?  @hjhsalo please rebase to latest master

PR #1748, opened 2017-11-13
collections: adds more collection pages

ioannistsanaktsidis:
* References #1384.


cernopendata/modules/collections/descriptions.py
```diff
@@ -105,6 +117,14 @@
         'name': 'CMS Simulated Datasets',
         'text': 'This collection contains CMS Simulated Datasets.'
     },
+    'cms-tools': {
+        'name': 'CMS Tools',
+        'text': ''  # TOFIX Add description
+    },
+    'cms-trigger-information': {
+        'name': 'CMS Trigger Information',
+        'text': ''  # TOFIX Add description
```
tiborsimko: The collection descriptions could be taken from the COD2 `qa` branch, see file
`invenio_opendata/base/fixtures/websearch.py`.  If we need them at all, actually... since we want to hide the notion of "collections" in COD3?!

ioannistsanaktsidis: Actually I don't really see the point of having descriptions as we always redirect to the search  page. WDYT ?

ioannistsanaktsidis:
@tiborsimko  LGTM

PR #1601, opened 2017-11-07
templates: VM and getting_started pages

atrisovic:
(closes: #1391)
(closes: #1392)
(closes: #1365)
(closes: #1366)

PR #1749, opened 2017-11-13
templates: new cms content

atrisovic:

### DIFF
--- a/cernopendata/modules/fixtures/data/articles/cms-guide-troubleshooting/cms.json
+++ b/cernopendata/modules/fixtures/data/articles/cms-guide-troubleshooting/cms.json
@@ -1,0 +1,22 @@
+[
+  {
+    "title": "CMS Response to errors",
+    "type": {"primary": "Documentation", "secondary": ["Guide"]},
+    "experiment": "CMS",
+    "collections": [
+      {
+        "experiment": "CMS"
+      },
+      {
+        "primary": "education"
+      }
+    ],
+    "tags":[
+        "Getting Started"
+    ],
+    "body": {
+      "content": "errors.md",
+      "format": "md"
+    }
+  }
+]
--- a/cernopendata/modules/fixtures/data/articles/cms-guide-for-education/cms.json
+++ b/cernopendata/modules/fixtures/data/articles/cms-guide-for-education/cms.json
@@ -1,0 +1,25 @@
+[
+  {
+    "title": "CMS Guide to education use of CMS Open Data",
+    "type": {"primary": "Documentation", "secondary": ["Guide"]},
+    "experiment": "CMS",
+    "collections": [
+      {
+        "experiment": "CMS"
+      },
+      {
+        "primary": "education"
+      },
+      {
+        "year": "2011"
+      }
+    ],
+    "tags":[
+        "Getting Started"
+    ],
+    "body": {
+      "content": "education.md",
+      "format": "md"
+    }
+  }
+]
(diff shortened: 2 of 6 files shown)


sefeg:
As discussed, tags (CMS, Documentation, Guide, Getting Started) need to be placed correctly and links need to be added. For the moment, some links will point to the current version of the portal and links that refer to tag combinations (e.g. CMS -> Guides) will point to -dev. We have to update those links before the final release, but can use the old links and the -dev links for testing of the guide right now.

atrisovic:
I asked @ioannistsanaktsidis about the tags, and he suggested we ask @tiborsimko  😄 

I will fix the links now

cernopendata/modules/fixtures/data/articles/getting-started-with-cms/cms.json
```diff
@@ -21,5 +21,68 @@
       "content": "cms.md",
       "format": "md"
     }
+  },
+  {
```
tiborsimko: This would add the new articles in the same JSON under the same `getting-started-with-cms` directory.

If we go for the full slug-oriented  approach discussed in https://github.com/cernopendata/opendata.cern.ch/pull/1601#discussion_r149937078 then it would be better to split them up to have:

```
- /articles/geting-started-with-cms-open-data/...
- /articles/cms-guide-for-education/...
- /articles/cms-guide-for-research/...
```

where the directory name would represent the slug of each given article.
```

Comment (atrisovic) [reply cernopendata/modules/fixtures/data/articles/getting-started-with-cms/cms.json]: ok! 👍🏻

**Comment by tiborsimko** [comment]
> the -dev links for testing of the guide right now.

Regarding DEV links in this PR, instead of:

- `http://opendata-dev.web.cern.ch/search?page=1&size=20&type=Documentation&experiment=CMS&subtype=Guide`

you can use the relative link: 

- `/search?page=1&size=20&type=Documentation&experiment=CMS&subtype=Guide`

which will work on both DEV, QA and PROD later, without having to change the host name...

File: cernopendata/modules/fixtures/data/articles/cms-guide-for-errors/errors.md [hunk 2]
```diff
@@ -0,0 +1,23 @@
+**I have trouble installing Virtualbox**
```
tiborsimko: If we go for the filename-represents-slug technique discussed elsewhere, then let's not call this "cms-guide-for-errors" 😄 but rather something like "cms-guide-troubleshooting".  @sefeg @daslerr can advise on the slugs.

(IMO it would be hepful to construct the full table like:

content | old URL | new URL
----------|--------|--------------
VMs for ALICE | `/VM/ALICE` | `/articles/virtual-machines-alice`
CMS guide written by Achim | *new content* | `/articles/cms-guide-for-research`

which can help to design the full URL slug map...)

sefeg: cms-guide-troubleshooting sounds really great!

atrisovic: 👍

cernopendata/modules/fixtures/data/articles/cms-guide-troubleshooting/cms.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
+    "title": "CMS Response to errors",
```
tiborsimko: All looks good, except for this title, which generates the URL:

- http://0.0.0.0/articles/cms-response-to-errors

with the visible title:

- CMS Response to errors

It may read a bit weird... Do we want to rename the title to something like "CMS Guide Troubleshooting" or something?

Can merge this anyway for quick showcasing to @katilp tomorrow...

atrisovic: Ok! I'll add that with the redirection view. Sorry I didn't see it earlier

PR #1756, opened 2017-11-14
templates:  redirection /vm/experiment

atrisovic:
In this commit:

- Changed guide title to "CMS Guide to troubleshooting"
- Created a view:
```
@blueprint.route('/vm/<exp>', defaults={'year': None})
@blueprint.route('/vm/<exp>/<year>')
def vm_redirect(exp, year):
```
- Changed year from 2010 to 2011 in CMS VM guide


cernopendata/modules/fixtures/data/articles/virtual-machines-cms-2011/cms.json
```diff
@@ -1,6 +1,6 @@
 [
   {
-    "title": "CMS 2010 Virtual Machines: How to install",
+    "title": "CMS 2011 Virtual Machines: How to install",
```
tiborsimko: Please rename also the files `*-cms-2011` so that there won't be any confusion when we add the 2010 ones 😉

tiborsimko: Oh, I see it's renamed everywhere, sorry 😉

PR #1759, opened 2017-11-14
files: addition of file metadata in record '_files'

pamfilos:
* refactors records_files to accept and import metadata to ```_files``` inside record

* @tiborsimko we should rethink just pushing the files metadata inside record's metadata, since we will get the same info as default records_files gives, plus everything else we need ( like 'type', 'name', etc), without all the trouble

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -28,13 +28,13 @@ from __future__ import absolute_import, print_function
 
 import os
 
-from invenio_records_files.api import _Record
 from invenio_records_rest.config import RECORDS_REST_ENDPOINTS
 from invenio_records_rest.facets import terms_filter
 from invenio_records_rest.utils import allow_all
 
 from cernopendata.modules.pages.config import *
 from cernopendata.modules.theme.config import *
+from cernopendata.modules.records.api import Record as _Record
 
 # Debug
 DEBUG = True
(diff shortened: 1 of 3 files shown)


tiborsimko:
> we should rethink just pushing the files metadata inside record's metadata

Do you mean for the `record.files.type` property?  If it simplifies things, we can go for two different  `record.files` and `record.file-indexes` properties, where one would list all the dataset files, and the other would list all the dataset file indexes. We discussed such an option last week instead of using the `record.files.type` keys... If you would prefer that way over this PR, we can surely explore it.  (But you might have to "join" the two parts in the fixture loader...)

pamfilos:
The issue is not how we structure them ( will not make any difference ). 

We have two options ( both will work for checking ```record.files.type``` or other file metadata needed):

(1) use this PR which refactors a bit the way ```invenio-records-files``` connects records and files to create ```_files```. This way ```_files``` exists with file metadata inside, but might take more time to load the fixtures on DB. 

(2) use default ```invenio-records-files``` to connect files with records (but without file metadata) and keep the ```files``` property from fixtures ( instead of popping them as we do now) to have it stored within the record metadata in the DB

tiborsimko:
@pamfilos Thanks, I see...  Related to your considerations is also the point about how do we expose files in our REST API.  This is something that people will surely want for fine analyses.  Currently we do this: 

- http://opendata.web.cern.ch/api/records/14

which exposes our buckets:

- http://opendata.web.cern.ch/api/files/45d8c283-c2e3-4b7c-9e0f-e7dac5b24a57

which is not very XRootD friendly, and which goes via our web site as a bottleneck.  

People basically want to see EOSPUBLIC's XRootD paths, which will permit them to use EOSPUBLIC directly, which will also be more scalable thanks to a multitude of EOSPUBLIC front end nodes.

If we consider this issue at the same time, this leads us perhaps towards the solutino #2 that you proposed...

tiborsimko:
@pamfilos Let's close this one now that option no.2 was implemented in #1779.

PR #1810, opened 2017-11-16
templates: fixes parent dataset

### LINKED ISSUE #1763: templates: fix display of `relations` property
**Example:**

- http://opendata-dev.web.cern.ch/record/14

**Problem:**

`Parent Dataset: /Mu/Run2010B-v1/RAW`  is linked even though we don't host the given RAW dataset.

**Solution:**

Don't link the `title` property in `relations` in the `isChildOf` and other cases. Display the title only.  Link only when we have `recid` or `doi` or other such knowable linkable property there.

ioannistsanaktsidis:
* Closes #1763.

### DIFF
--- a/cernopendata/templates/cernopendata_records_ui/records/record_detail.html
+++ b/cernopendata/templates/cernopendata_records_ui/records/record_detail.html
@@ -86,7 +86,13 @@
                 {% for relation in record.relations %}
                 {% if relation.type == 'isChildOf' %}
                 <label>Parent Dataset:</label>
-                <span><a href="/record/{{relation.recid}}">{{relation.title}}</a></span>
+                <span>
+                  {% if relation.recid %}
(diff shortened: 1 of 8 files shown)


cernopendata/templates/cernopendata_records_ui/records/record_detail.html
```diff
@@ -86,7 +86,13 @@ <h6 class="card-title">
                 {% for relation in record.relations %}
                 {% if relation.type == 'isChildOf' %}
                 <label>Parent Dataset:</label>
-                <span><a href="/record/{{relation.recid}}">{{relation.title}}</a></span>
+                <span>
+                  {% if relation.recid %}
+                  <a href="/record/{{relation.recid}}">{{relation.title}}</a>
```
tiborsimko: Beware, `relation.title` is usually not present when `relation.recid` is... but we can take it up later as things will appear.

PR #1820, opened 2017-11-16
templates: redirect for getting-started

atrisovic:
redirects:

`getting-started/alice` ->`/articles/getting-started-with-alice-open-data`
`getting-started/cms` -> `/articles/getting-started-with-cms-open-data`
`getting-started/lhcb` -> `/articles/getting-started-with-lhcb-open-data`

### DIFF
--- a/cernopendata/modules/pages/views.py
+++ b/cernopendata/modules/pages/views.py
@@ -253,6 +253,13 @@ def about_opera():
     return render_template('cernopendata_pages/about/about_opera.html')
 
 
+@blueprint.route('/getting-started/<exp>')
+def getting_started_redirect(exp):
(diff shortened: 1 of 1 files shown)


ioannistsanaktsidis:
Not sure we need this one. @tiborsimko right now ```getting-started/alice``` redirects to a faceted search E.g ```http://127.0.0.1:5000/search?page=1&size=20&experiment=ALICE&tags=Getting%20Started```. Do we need to change that ?

tiborsimko:
@ioannistsanaktsidis Yes, we do.  Redirecting to the final web page, rather than to an intermediate search page, is better when we know where to redirect 😄

PR #1825, opened 2017-11-17
installation: dev friendly docker deployment

### LINKED ISSUE #1769: installation: DEV friendly setup
We currently have:

``` console
$ grep DEBUG cernopendata/config.py
DEBUG = True
# ASSETS_DEBUG = True
```

It is not good to hardcode DEBUG in the config file like that.

Let's create `docker-compose.yml` (=suitable for production) and `docker-compose-dev.yml` (=suitable for development) where we can set up all the various asset debugging related environment variables properly (off for production, on for development).

This will be especially important once we have nginx caching in place (#1728) that should only be used in production-like conditions.

(See COD2 files for inspiration.)

hjhsalo:
* Add a docker-compose-dev.yml that handles development
  friendly docker deployment of cernopendata.
  (Turns off nginx-caching, sets DEBUG and FLASK_DEBUG environment
  variables)
* Use `docker-compose -f docker-compose-dev.yml up` to run.
  (closes #1769)


cernopendata/config.py
```diff
@@ -37,7 +37,10 @@
 from cernopendata.modules.theme.config import *
 
 # Debug
-DEBUG = True
+DEBUG = os.environ.get(
+    'DEBUG',
+    'False'
+)
 TEMPLATES_AUTO_RELOAD = True
```
tiborsimko: I'd also treat `TEMPLATES_AUTO_RELOAD` as a debugging-related variable, i.e. when the `DEBUG` environment variable is True, then both `config.DEBUG` and `config.TEMPLATES_AUTO_RELOAD` are True; otherwise both are False by default.

nginx/Dockerfile
```diff
@@ -23,8 +23,12 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
 FROM nginx:perl
+
 RUN rm /etc/nginx/conf.d/default.conf
 ADD cernopendata.conf /etc/nginx/conf.d/
 
+# Turn caching in development deployment
+RUN if [ "$DEBUG" = "1" ]; then sed -i 's/proxy_cache cache_cod;/proxy_cache off;/' /etc/nginx/conf.d/cernopendata.conf; fi
```
tiborsimko: `if [ -z "$DEBUG" ]` might be safer regarding value selection.

nginx/Dockerfile
```diff
@@ -23,8 +23,15 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
 FROM nginx:perl
+
 RUN rm /etc/nginx/conf.d/default.conf
 ADD cernopendata.conf /etc/nginx/conf.d/
 
+# Turn off caching in development deployment
+RUN if [ -z "$DEBUG" ]; then \
```
tiborsimko: Maybe remove the whole `sed` block, since people will use port 5000 for direct developments..

docker-compose.yml
```diff
@@ -104,35 +114,33 @@ services:
...
+      - webdata:/usr/local/var/cernopendata/var/cernopendata-instance/static
 
-  static:
-    restart: "no"
-    build: .
-    volumes:
-      - /usr/local/var/cernopendata/var/cernopendata-instance/static
-    user: invenio
+volumes:
+  webdata:
```
tiborsimko: To keep naming harmony with containers: 

- webdata 
- postgresqldata
- elasticsearchdata

PR #1831, opened 2017-11-20
records: no xrootd in files.type

### LINKED ISSUE #1811: fixtures: wrong filetype in authors-list
filetype is assigned as ```xrootd```. Should be something else or empty

https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/modules/fixtures/data/records/cms-author-list.json#L19036

https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/modules/fixtures/data/records/opera-author-list.json#L538

https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/modules/fixtures/data/records/cms-author-list-Run2011A.json#L19861

tiborsimko:
* Removes `xrootd` values in `files.type`. (closes #1811)

### DIFF
--- a/cernopendata/modules/fixtures/data/records/alice-analysis-modules.json
+++ b/cernopendata/modules/fixtures/data/records/alice-analysis-modules.json
@@ -20,7 +20,6 @@
     {
       "checksum": "sha1:0000000000000000000000000000000000000000", 
       "size": 2086046, 
-      "type": "xrootd", 
       "uri": "root://eospublic.cern.ch//eos/opendata/alice/modules/PtAnalysis.tgz"
     }
   ], 
@@ -63,7 +62,6 @@
     {
       "checksum": "sha1:0000000000000000000000000000000000000000", 
       "size": 31139203, 
-      "type": "xrootd", 
       "uri": "root://eospublic.cern.ch//eos/opendata/alice/modules/alice-mc.tgz"
     }
   ], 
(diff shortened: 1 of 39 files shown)


ArtemisLav:
Should we also remove the `type` property from the "master" schema (i.e. `record-v1.0.0.json`)? Actually, for `files`, I think we only use `checksum`, `size` and `uri` (and `type`), but the schema includes other properties as well, which I'm not sure we need:
```
       "_files": {
            "type": "array",
            "uniqueItems": true,
            "items": {
                "type": "object",
                "properties": {
                    "uri": {
                        "type": "string"
                    },
                    "key": {
                        "type": "string"
                    },
                    "bucket": {
                        "type": "string"
                    },
                    "version_id": {
                        "type": "string"
                    },
                    "size": {
                        "type": "integer"
                    },
                    "type": {
                        "type": "string"
                    },
                    "description": {
                        "type": "string"
                    },
                    "checksum": {
                        "type": "string"
                    }
                }
            }
        }
```

tiborsimko:
@ArtemisLav Note that this is `_files`, where we need bucket ID etc for internal file storage. This is an Invenio thing, so to say...

ArtemisLav:
Right, I see, so if they are all needed, that's fine with me. 
However, I _was_ expecting to find a `files` property in the schema, which is why I assumed it was this one. That's not relevant here though, so I'm approving the PR now.

Issue #1785, opened 2017-11-15
CMS: Update the pileup info page to include information for 2012


katilp:
The original twiki page has been updated by Mike in:
https://twiki.cern.ch/twiki/bin/view/CMSPublic/PublicPileupInfo
The pile-up record can now be updated with the new information (i,e the text and the plot which is under 2012 in the page above).

In the old portal we had two records

http://opendata.cern.ch/record/70
http://opendata.cern.ch/about/CMS-Pileup-Simulation
(NB, one is pileup, other is pile-up....)

Is there a good reason for this separation? Can we have only one? This would be rather documentation than a record. @tiborsimko What do you think? Maybe we can have a look together if needed.

katilp:
The 2012 text mentions reweighing and we may want to explain/add an example of what it means:

Mike says there is a DAS module on this, which is here:

https://twiki.cern.ch/twiki/bin/view/CMS/SWGuideCMSDataAnalysisSchool2012PileupReweighting

Definitely examples are available there.  The main Twiki page is:

https://twiki.cern.ch/twiki/bin/viewauth/CMS/PileupInformation#Reweighting_Tools

the tools haven’t changed, but the page definitely looks old.

katilp:
@atrisovic Could you have a look at this issue? Having two pages (as currently: http://opendata.cern.ch/record/70  and http://opendata.cern.ch/about/CMS-Pileup-Simulation), is not necessary, they can be combined in one article (Category Documentation, subcategory Guide).
Then the information from 2012 can be added to that page.
The information source is https://twiki.cern.ch/twiki/bin/view/CMSPublic/PublicPileupInfo

PR #1863, opened 2017-11-22
templates: CMS: pileup simulation page

atrisovic:
closes: #1785


katilp:
These are corrections to the already existing text, I guess:
1. There's an "education" keyword (if I understand correctly), it should be removed. This is certainly more research oriented information.
2. We may want to add  "[CMSSW](http://opendata-dev.web.cern.ch/search?page=1&size=20&q=CMSSW&type=Software)" before words "package"  (with the correct path...)
3. I'm checking if we can have the information of the pwd protected twiki page to the public, and meanwhile it will need a FIXME as the page is not accessible

atrisovic:
I think I implemented all three points in the new commit: +1:

PR #1887, opened 2017-11-24
cli: adds replace option

### LINKED ISSUE #1836: fixtures: insert/replace record loading mode
In order to work with records on development and production machines, it is useful to have an easy possibility to load the same record more than once, for example to propagate record changes following data updates. This basically means doing sometimes a REPLACE operation rather than an INSERT operation.

The usage scenario:

```console
$ cernopendata fixtures records -f cernopendata/modules/fixtures/data/records/myrec123.json
$ vim cernopendata/modules/fixtures/data/records/myrec123.json
$ cernopendata fixtures records -f cernopendata/modules/fixtures/data/records/myrec123.json
[ERROR] Cannot create record, PID recid 123 already exists.
$ cernopendata fixtures records -f cernopendata/modules/fixtures/data/records/myrec123.json --replace
[INFO] Replacing record 123... Done.
```

The wanted behaviour, "insert" or "replace", could be governed by a CLI switch:

- e.g. we can always require the presence of `--mode` and the user should explicitly specify either `--mode insert` or `--mode replace` always;
- e.g. we can do "insert" by default and refuse to update existing records, as is the case now, unless the `--replace` CLI switch is present (as illustrated above). 

P.S. Note that in COD2, we had two separate options (`-i` for insert and `-r` for replace), which was useful when the input file had some records changed, and some records added. A user could specify either `-i`, or `-r`, or both `-i -r` meaning to either insert the record (if not found) or to replace the record (if found).

ioannistsanaktsidis:
* Closes #1836.

### DIFF
--- a/cernopendata/modules/fixtures/cli.py
+++ b/cernopendata/modules/fixtures/cli.py
@@ -32,6 +32,18 @@ from flask import current_app
 from flask.cli import with_appcontext
 from sqlalchemy.orm.attributes import flag_modified
 
+from invenio_db import db
+from invenio_records_files.api import Record
+from invenio_indexer.api import RecordIndexer
+from cernopendata.modules.records.minters.recid import \
+    cernopendata_recid_minter
+
+from invenio_files_rest.models import \
+    Bucket, FileInstance, ObjectVersion
+from invenio_records_files.models import RecordsBuckets
+from invenio_pidstore.models import PersistentIdentifier
+from invenio_pidstore.errors import PIDDoesNotExistError
+
 
 def get_jsons_from_dir(dir):
     """Get JSON files inside a dir."""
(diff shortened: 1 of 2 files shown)


tiborsimko:
@ioannistsanaktsidis @pamfilos Tested the PR locally, works nicely, here are some observations for improvements:

**1. Repetitive uploads.** *(important)* if you load a file that has 100 records, say, and it fails in the middle, and you fix a problem for record number 50, then you cannot reload the file again, because:
  - using "insert" would lead to problems that PIDs already exist for existing records, and early exit;
  - using "replace" would work for only already-uploaded files, so no new records would be added.   

You can try with:

```
    -f ./cernopendata/modules/fixtures/data/records/cms-primary-datasets.json \
    --skip-files --mode insert --verbose
```

that you would interrupt in the middle after say only 6 records are loaded.

=> Can you add an insert-or-replace option, for example allow multiple values `--mode insert --mode replace` that would do the job? Or for example make a new value `--mode insert-or-replace`?

The wanted behaviour is:

**mode** | **behaviour when PID exists** | **behaviour when PID does not exist** | 
--- | --- | --- |
insert | fail | upload as new | 
replace | update as existing | fail | 
insert or replace | update existing | upload as new | 

We basically need a new CLI option to express the 3rd scenario.  Allowing multiple values `--mode insert --mode replace` could be one way; creating a new value `--mode insert-or-replace` could be another.

**2. Output messages.** *(cosmetics)* You could catch the exception `invenio_pidstore.errors.PIDAlreadyExists` and reply somehing more user friendly, like:

```
[ERROR] Record recid 15 exists already; cannot insert it.  
[ERROR] Record recid 15 does not exist; cannot replace it.
```

for the two "fail" situations in the above table.

You could also introduce informative messages:

```
[INFO] Record recid 15 inserted.
[INFO] Record recid 15 updated.
```

for the other situations in the above table, when things are OK and when the user used `--verbose` argument.

**3. Non-record content.** *(future outlook)* For now, doing this for records is enough.  For future, it would be good to think about doing this for other fixtures, such as "docs", so that one could log into the PROD machine and upload only the latest given news document, without disturbing the others. Noting this down just for completeness; not necessary to address this now as part of this PR!

ioannistsanaktsidis:
@tiborsimko updated ! You can check now if it fits our case :) I think it covers all your comments.

tiborsimko:
@ioannistsanaktsidis Great, everything looks OK!

Issue #1937, opened 2017-11-29
Fix and update links in about CMS page


katilp:
Links to search rather than to collection, mention RAW in http://opendata-dev.web.cern.ch/articles/about-cms

katilp:
Closed by #1941

Issue #1787, opened 2017-11-15
Add guides to "About CMS Open Data" page


sefeg:
As discussed with @katilp and Achim, we should list the most important guides on the [About CMS](http://opendata-dev.web.cern.ch/about/cms) collaboration page.

We have to decide which resources we want to add to such a box.

So far we discussed:
- [Guide to education use of CMS Open Data](http://opendata-dev.web.cern.ch/articles/cms-guide-to-education-use-of-cms-open-data)
- [Guide to research use of CMS Open Data](http://opendata-dev.web.cern.ch/articles/cms-guide-to-research-use-of-cms-open-data)

I imagine that we should also add [Getting Started with CMS Open Data](http://opendata-dev.web.cern.ch/articles/getting-started-with-cms-open-data)

Do you think that something else should be there? @katilp

katilp:
@sefeg 

These two guides are quite comprehensive, so in my view the two of them will be enough in About CMS.

I'm a bit hesitant for Getting started, I would not like to distract the readers while they are going through it. Maybe a link to [Research guide](http://opendata-dev.web.cern.ch/articles/cms-guide-to-research-use-of-cms-open-data) at the end of the instructions....

katilp:
@atrisovic @RaoOfPhysics Could you take care of this?

RaoOfPhysics:
@atrisovic: Should I?

atrisovic:
you should ;)

RaoOfPhysics:
- [ ] All links to primary, derived etc. are broken and need fixing.

katilp:
Closed by #1941

PR #1938, opened 2017-11-29
CMS: templates: add guide links and fixes other link in about cms

katilp:
Addresses #1787 and #1937

### DIFF
--- a/cernopendata/modules/fixtures/data/articles/about-cms/cms.md
+++ b/cernopendata/modules/fixtures/data/articles/about-cms/cms.md
@@ -1,8 +1,10 @@
 The Compact Muon Solenoid (CMS) Experiment is one of the large particle detectors at CERN's Large Hadron Collider. The CMS Collaboration consists of more than 3000 scientists, engineers, technicians and students from 180+ institutes and universities from 40+ countries. You can find more information about the CMS [detector design](https://cms.cern/news/cms-detector-design) and [overview](https://cms.cern/news/detector-overview) on the official [CMS website](https://cms.cern).
 
-This page gives a brief overview of CMS Open Data contents, you can find usage instructions and suggestions in two detailed guides:
+You can find usage instructions and suggestions in two detailed guides:
 - [Guide to education use of CMS Open Data](/articles/cms-guide-to-education-use-of-cms-open-data)
 - [Guide to research use of CMS Open Data](/articles/cms-guide-to-education-use-of-cms-open-data).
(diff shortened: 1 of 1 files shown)


katilp:
Adding the "Guide" links is important, but as they are there now with this PR, it reads very strange (the page starts with introduction to CMS, then deviates to the guide links and continues again with the core subject of the page: the open data...). It would be best to have some graphical buttons for the guides so that they would be more "separate" from the normal text.

cernopendata/modules/fixtures/data/articles/about-cms/cms.md
```diff
@@ -36,7 +42,8 @@ The following are provided through this portal:
     * Definition of same objects is different in each analysis.
 
 * The files can be read in [ROOT](http://root.cern.ch/), but they cannot be opened (and understood) as simple data tables.
-* Only the runs that are validated by data quality monitoring should be used in any analysis. The [list of the validated runs](/collection/CMS-Validation-Utilities) is provided.
+* Only the runs that are validated by data quality monitoring should be used in any analysis. The [list of the validated runs](/search?page=1&size=20&q="CMS list of validated runs") (FIXME markdown does not accpet hyphen in links?) is provided.
```
RaoOfPhysics: Try this: `/search?page=1&size=20&q="CMS%20list%20of%20validated%20runs"`

cernopendata/modules/fixtures/data/articles/about-cms/cms.md
```diff
@@ -36,7 +42,8 @@ The following are provided through this portal:
...
+* A small sample of [RAW data](/search?page=1&size=20&q=&experiment=CMS&file_type=raw) is also provided.
```
RaoOfPhysics: Should this be "raw" rather than "RAW"?

katilp: We are sort of used to RAW in CMS (i.e. RAW, RECO, AOD), but indeed here probably "raw data" is better

katilp:
Thanks, updates done!

katilp:
For merging, let me know if I should so this differently 😃

RaoOfPhysics:
@katilp: In your original post I think one of the issue numbers is wrong. And maybe you can remove the bit about "FIXME"? Thanks!

ioannistsanaktsidis:
Closing as superseded by #1941

PR #2004, opened 2017-12-04
docker: add elasticsearch 5 support

### LINKED ISSUE #2001: docker: add elasticsearch 5 support
Currently Elasticsearch version used in Docker / Docker-Compose based installation is Elasticsearch 5.

-  [ ] Make changes to Docker / Docker-Compose files to enable use of Elasticsearch 5.

Note that _mapper-attachments_ -plugin has been deprecated (but still installable) in Elasticsearch 5 and replaced by  _ingest-attachment_ -plugin. (https://www.elastic.co/guide/en/elasticsearch/plugins/2.4/mapper-attachments.html)
However, since opendata.cern.ch is not using full text search / indexing this probably doesn't matter.

hjhsalo:
* Add necessary build argument (ES_VERSION) to specify which
  major version of Elasticsearch should be used.
  Tested with ES_VERSION=2 and ES_VERSION=5
  (closes #2001)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,12 @@ sudo: false
 
 language: python
 
+env:
+  # Latest available version of Docker-Compose at the time of commit.
+  # To download latest available version, one needs to do some (JSON) parsing
+  # as Github doesn't provide an easy way to download the latest release.
+  - DOCKER_COMPOSE_VERSION=1.17.1
+
 matrix:
   fast_finish: true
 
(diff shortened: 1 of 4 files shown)


hjhsalo:
Failing on Travis. Seems to fail on Dockerfile syntax, which works fine on my machine with Docker 17.09.0-ce. Probably Travis uses an older version of Docker.

diegodelemos:
What about setting an environment variable (see [here](https://docs.docker.com/compose/compose-file/#variable-substitution)) in the shell that runs the `docker-compose` and then in the `docker-compose.yaml`:
```yaml
elasticsearch:
  restart: "always"
  build:
    context: .
    image: "elasticsearch:${ES_VERSION}"
    dockerfile: ./elasticsearch/Dockerfile
```
If this doesn't work, we could also try with two Dockerfiles, let say `./elasticsearch/Dockerfile.2` and `./elasticsearch/Dockerfile.5` each one with the corresponding `FROM` clause. Then inside the `docker-compose.yml` doing something like:
```yaml
elasticsearch:
  restart: "always"
  build:
    context: .
    dockerfile: ./elasticsearch/Dockerfile.${ES_VERSION}
```

hjhsalo:
@diegodelemos 
In the end I updated Docker-CE and Docker-Compose on Travis-CI.

It adds a little bit of build time, but it seems to be kind of promoted by Travis-CI team as they have update instructions for Docker and Docker-compose in their "Using Docker in Builds" -document:
https://docs.travis-ci.com/user/docker/#Using-Docker-Compose
https://docs.travis-ci.com/user/docker/#Installing-a-newer-Docker-version

diegodelemos:
could we get advantage of Travis caching using `apt-get` for `docker` install and `pip` for `docker-compose`? it is also listed among the best practices for `docker` installation

diegodelemos:
taken over by #2080

PR #2014, opened 2017-12-04
Minor fixes

pamfilos:
* changes example search to 'keywords:education'

* updates icon for scrolling

* removes 'supplementaries' from 'Explore'

* fix glossary behaviour issues


cernopendata/templates/cernopendata_pages/front/main.html
```diff
@@ -22,7 +22,7 @@ <h2>
                 <div class="search-examples row justify-content-center">
                     <span class="example-title">search examples:</span>
                     <a href="{{url_for('invenio_search_ui.search', q = 'collision data') }}">collision datasets</a>,
-                    <a href="{{url_for('invenio_search_ui.search', q = 'configuration') }}">configurations</a>,
+                    <a href="{{url_for('invenio_search_ui.search', q = 'keywords:education') }}">keywords:education</a>,
```
tiborsimko: BTW `keywords:education` looks a bit weird if we use plural... We should remember to use singular later on at some point...

tiborsimko: ... when time will permit to address search issues. CC @annatrz

PR #2056, opened 2017-12-08
docs: changes to cms-simulated-dataset-names

atrisovic:


atrisovic:
No, I don't think so

katilp:
OK, I will open a issue for it

cernopendata/modules/fixtures/data/docs/cms-simulated-dataset-names/cms-simulated-dataset-names.json
```diff
@@ -1,7 +1,7 @@
 [
   {
     "title": "CMS Simulated Dataset Names",
-    "slug": "cms-dataset-names",
+    "slug": "cms-simulated-dataset-names",
```
tiborsimko: This was necessary to change in `body.content` too, otherwise the MD file is wrong and the installation breaks... I'll quick-merge a fix. PR #2065

PR #2086, opened 2017-12-11
installation: remove provisioning scripts

### LINKED ISSUE #1471: docker: remove custom entrypoint to name user at runtime
Since https://github.com/xrootd/xrootd/issues/584#event-1277883871 has been merged, remove https://github.com/cernopendata/opendata.cern.ch/blob/master/scripts/docker-entrypoint.sh and https://github.com/cernopendata/opendata.cern.ch/blob/master/Dockerfile#L72 hopefully everything should work 😇

diegodelemos:
* Removes provisioning scripts.

* Removes hack to name random user at runtime (closes #1471).

* Remove usage of `virtualenv`.


Dockerfile
```diff
@@ -25,63 +25,70 @@
...
-# Configure virtualenvwrapper root directory
-ENV WORKON_HOME=/usr/local/var
-ENV PROJECT_HOME=/usr/local/var
-# Prepare provisioning script:
-COPY scripts/provision-web.sh /tmp/
-
-# Install CERN Open Data Portal web node pre-requisites:
-RUN /tmp/provision-web.sh
+# Configuration for CERN Open Data Portal instance:
+ENV APP_INSTANCE_PATH=/usr/local/var/cernopendata/var/cernopendata-instance
```
diegodelemos: this might not be the best path ever, but changing it now would involve more changes on the running instances+testing just in case

PR #2099, opened 2017-12-12
scripts: bring create-instance.sh back

diegodelemos:


Dockerfile
```diff
@@ -67,22 +51,25 @@ RUN yum update -y && \
         xrootd-python && \
     yum clean -y all
 
+# Configuration for CERN Open Data Portal instance:
+ENV APP_INSTANCE_PATH=/usr/local/var/cernopendata/var/cernopendata-instance
+ARG UWSGI_WSGI_MODULE=cernopendata.wsgi:application
```
hjhsalo: Can L56-63 be moved to happen after L74 (or even after L78) so in the case they change the Dockerfile might not need to be rebuild.

Dockerfile
```diff
@@ -67,22 +51,25 @@ RUN yum update -y && \
...
-    mkdir -p ${APP_INSTANCE_PATH} && \
-    ${INVENIO_WEB_INSTANCE} npm && \
-    cd ${APP_INSTANCE_PATH}/static && \
-    CI=true npm install && \
-    npm install d3@3.3.13 && \
-    ${INVENIO_WEB_INSTANCE} collect -v && \
-    ${INVENIO_WEB_INSTANCE} assets build && \
+# Add CERN Open Data Portal sources to `code` and work there:
+WORKDIR /code
+ADD . /code
```
tiborsimko: This would cause the requirements to be reinstalled for any change in the code base e.g. even if you touch fixtures.
Since we removed the central provisioning scripts, we can take better advantage of Docker build cache.

E.g. you try to introduce something like: *(pseudo language)*

```
ADD requirements-production-local-forks.txt /tmp # containing our forks
ADD requirements-production.txt /tmp 
RUN pip install -r /tmp/requirements-production-local-forks.txt
RUN pip install -r /tmp/requirements-production.txt

WORKDIR /code
ADD . /code
# ... and continue as before
```

This would optimise rebuild times already nicely, since we rarely touch python requisites.

In the future, we could even look into caching the output of `npm` and friends, but we can look at that later...

Dockerfile
diegodelemos: not all, `APP_INSTANCE_PATH` needs to remain before `create-instance.sh` but is true that this we wont change, whereas tuning uWSGI parameters is likely to happen

Dockerfile
diegodelemos: thats a very good point, but then we loose again `create-instance.sh` or, if we keep it it's highly likely that they will get outdated (we will potentially end up with different ways of installing)

tiborsimko: I meant to repeat it... i.e. to introduce the new section I wrote above, but also to keep `create-instance.sh` later as well, as you had it. Basically, in `Dockerfile` we simply pro-actively pre-install everthing that we know that will be needed later (we take advantage of our knowledge!) so that when `create-instance.sh` will come and do its pip commands, the requirements will be already fulfilled.... hence this "second step" will finish ultra quickly.  In this way we both profit from the Docker cache, and we have container-agnostic `create-instance.sh` helper script.

PR #2104, opened 2017-12-12
global: adds threshold for file download

### LINKED ISSUE #2066: templates: file download warning box
The file download warning box seems to have disappeared.

Let us bring it back, as it is important do warn the users that they don't have to download big files if they use VM for analysing.

See #1437 and #1451.

ioannistsanaktsidis:
* Closes #2066.

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -385,6 +385,11 @@ FILES_REST_FILE_URI_MAX_LEN = os.environ.get(
     "FILES_REST_FILE_URI_MAX_LEN",
     512
 )
+#: Files max size threshold(bytes)
+CERNOPENDATA_MAX_DOWNLOAD_SIZE = os.environ.get(
+    "CERNOPENDATA_MAX_DOWNLOAD_SIZE",
+    200000000
+)
 
 # Search
 # ======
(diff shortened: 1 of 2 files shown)


cernopendata/config.py
```diff
@@ -379,6 +379,8 @@
 FILES_REST_PERMISSION_FACTORY = allow_all
 #: Allow URI's longer than 255 chars.
 FILES_REST_FILE_URI_MAX_LEN = 512
+#: Files max size threshold(bytes)
+FILES_REST_FILE_MAX_SIZE = 200000000
```
tiborsimko: > `FILES_REST_FILE_URI_MAX_LEN`
> `FILES_REST_FILE_MAX_SIZE`

We should probably name the new config variable differently, so that it wouldn't seem that it *also* comes  from upstream `invenio-files-rest`... What about `CERNOPENDATA_MAX_DOWNLOAD_SIZE`?  

(Thinking of having `CERNOPENDATA_...` prefix for all variables related to our instance...)

> FILES_REST_FILE_MAX_SIZE = 200000000

Please also try to get it from environment variables; this is useful in production deployments. Similarly to:

```python
FILES_REST_FILE_URI_MAX_LEN = os.environ.get(
    "FILES_REST_FILE_URI_MAX_LEN",
    512
)
```

Issue #2093, opened 2017-12-12
files: create rich index files


tiborsimko:
For every index file in TXT, create rich index file in JSON. Example content:

```json
[
    {
        "filename": "foo.zip",
        "size": 1234,
        "checkum": "abcde",
        "uri": "root://eospublic.cern.ch//eos/opendata/.../foo.zip"
    },
    {
        "filename": "bar.C",
        "size": 567,
        "checkum": "fghijk",
        "uri": "root://eospublic.cern.ch//eos/opendata/.../bar.C"
    }
]
```

@tiborsimko to create and attach JSON files on EOS for each concerned record. (PR #2075)
@pamfilos to take these files from `/record/NNN/files/XXX.json` and use them in the file pagination box on the detailed record pages.

Stemmed from  #1826.

tiborsimko:
Progress:
- [ ] CMS 2010 collision data
- [ ] CMS 2011 collision data
- [ ] CMS 2011 raw data
- [ ] CMS 2011 simulated data
- [ ] CMS 2012 collision data 
- [ ] CMS 2021 simulated data

ioannistsanaktsidis:
Closing as already implemented.

PR #2108, opened 2017-12-13
files: refactor file handling

pamfilos:
* addresses #2093

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -118,9 +118,15 @@ RECORDS_UI_ENDPOINTS = dict(
         view_imp='cernopendata.modules.records.utils:file_download_ui',
         record_class='invenio_records_files.api:Record',
     ),
+    recid_files_assets=dict(
+        pid_type='recid',
+        route='/record/<pid_value>/files/assets/<path:filepath>',
+        view_imp='cernopendata.modules.records.utils:eos_file_download_ui',
+        record_class='invenio_records_files.api:Record',
+    ),
     recid_files_page=dict(
         pid_type='recid',
-        route='/record/<pid_value>/files/p/<int:page>',
+        route='/record/<pid_value>/filepage/<int:page>',
         view_imp='cernopendata.modules.records.utils:record_file_page',
         record_class='invenio_records_files.api:Record',
     ),
@@ -334,7 +340,7 @@ RECORDS_REST_FACETS = {
             type=dict(terms=dict(
                 field='type.primary.keyword',
                 order=dict(_term='asc')),
-                      aggs=dict(subtype=dict(terms=dict(
+                aggs=dict(subtype=dict(terms=dict(
                           field="type.secondary.keyword",
                           order=dict(_term='asc'))))),
             file_type=dict(terms=dict(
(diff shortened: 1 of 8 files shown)


cernopendata/templates/cernopendata_records_ui/records/detail.html
```diff
@@ -252,6 +294,65 @@ <h4> Export </h4>
...
+            windowClass: 'index-file-list-modal',
+            resolve: {
+              $http: function() {
+                return $http;
+              },
+              pid: function() {
+                return pid;
+              },
+              uri: function() {
+                if (file.endsWith('index.txt')) {
```
tiborsimko: You can update this now as per new `type` values of "index.txt" and "index.json"

cernopendata/config.py
```diff
@@ -385,6 +391,8 @@
     "FILES_REST_FILE_URI_MAX_LEN",
     512
 )
+#: Make download availalbe for Root files
+FILES_ROOT_DOWNLOAD_DISABLED = False
```
tiborsimko: `CERNOPENDATA_DISABLE_DOWNLOADS`?

Please wrap in `os.environ.get` see my earlier comment regarding naming convention https://github.com/cernopendata/opendata.cern.ch/pull/2104#discussion_r156620651

cernopendata/config.py
```diff
@@ -118,6 +118,12 @@
         view_imp='cernopendata.modules.records.utils:file_download_ui',
         record_class='invenio_records_files.api:Record',
     ),
+    recid_files_root=dict(
+        pid_type='recid',
+        route='/record/<pid_value>/files/root',
```
tiborsimko: This is not only for `root` files, so we may invent a different name to express the notion of "asset files behind index files".

cernopendata/config.py
```diff
@@ -390,6 +396,8 @@
     "CERNOPENDATA_MAX_DOWNLOAD_SIZE",
     200000000
 )
+#: Make download availalbe for Root files
+CERNOPENDATA_DISABLE_DOWNLOADS = False
```
tiborsimko: Please wrap in `os.environ.get()` so that one can influence it rapidly in a deployed containerised environment.

Example:

```
CERNOPENDATA_DISABLE_DOWNLOAD = os.environ.get(
     "CERNOPENDATA_DISABLE_DOWNLOADS",
     0
 )
```

tiborsimko:
Works nicely 👍 Here are four quick observations:

**1.** Please wrap `CERNOPENDATA_DISABLE_DOWNLOADS` in `os.environ.get()` call as mentioned in https://github.com/cernopendata/opendata.cern.ch/pull/2108/files#r157023381

**2.** The name of the "Browse" button may be misleading, since people will think they will go into "browsing text file" mode, which means basically "viewing text file live", as opposed to "Download" button which is right next to it.  What about inventing some other name for "Browse", such as "List files"? This is basically what is printed later... @sefeg @daslerr could chime in here once we deploy.  

**3.** The Files box is currently showing the JSON index files.  Please hide them. The Files box should only show files that don't have `type` property starting by "index".  (Hence both TXT and JSON index files should not appear there...

**4.** The modal button works well in the "List of files" box as well, but if I start to download the file, the proposed filename to store is `filename` and not `34BE0595-3982-E211-95D3-002618943845.root`.

filebox

PR #2112, opened 2017-12-13
records: rich index files for CMS 2012 collision datasets

tiborsimko:
* Adds rich index files (in TXT and JSON formats) for CMS 2012 collision
  datasets. (addresses #2093)


tiborsimko:
@pamfilos This is the rich format we have been discussing.  Please note that until I'm done with converting everything, we'll have three `type` values:

- index
- index.txt (=equal to index)
- index.json (=new files) 

You can display both "index" and "index.type" in the "File indexes" box until, so that things are working fine until we are done converting everything...

PR #2148, opened 2017-12-14
CMS: records: fixed and updates to 5500

### LINKED ISSUE #2142: CMS: final updates to 5500
In Description:
- last sentence: ",which can also be obtained from corresponding
(separate) tutorials" might be obsolete since you have kindly inserted the
generic ROOT link. Or add the link to the ROOT tutorials from level 2 also
here?

- ROOT/Root/root in different places.

In Notes:

- 2nd paragraph): change "roughly equivalent to the public set"
to "roughly equivalent to the public sets" 

How can you use this?

- level 1: add links to the corresponding files on that record

- level2: how to download (all files....)
- add link to M4Lnormdatall.cc

- level 3: Again, is it possible to link the file names directly to the
files (if not, it is fine as it is).

- level 4: add links to files

- on the 2nd last bullet (merge all ...) , add:
"If everything ran correctly, at this stage the content of your rootfiles
should be identical to the one in the predefined rootfiles. If you don't
want to run everything, you can also run only some of the datasets (but
all events of all index files of a dataset must be treated and merged),
and use the predefined ones for the others."

- on the last bullet, add
", i.e. edit the file M4Lnormdatall.cc accordingly. If everything worked
correctly you should get an output identical to mass4lcombine.pdf, but
now using your own files."

katilp:
Closes #2142


cernopendata/modules/fixtures/data/records/cms-tools-higgsexample20112012.json
```diff
@@ -112,7 +112,7 @@
...
...provide support for the single virtual machine mode), you may also run the analysis in parallel on different CPUs, correspondingly speeding up the result.</li>\n <li> merge all the files from different index files of a dataset by using root tools</li>\n>ul><li>If everything ran correctly, at this stage the content of your rootfiles should be identical to the one in the predefined rootfiles. If you don't want to run everything, you can also run only some of the datasets (but all events of all index files of a dataset must be treated and merged), and use the predefined ones for the others. </li> </ul>\n <li> go to 2., using your own root output files instead of the predefined ones, i.e. edit the file <a href=\"/record/5500/files/M4Lnormdatall.cc\">M4Lnormdatall.cc</a> accordingly. If everything worked correctly you should get an output identical to mass4lcombine.pdf, but now using your own files.</li>\n </ul></ol>\n In addition to these instructions, which guide you through the example in detail, <a href=\"https://github.com/cms-opendata-analyses/HiggsExample20112012\">a github repository</a> based on this original code with minor modifications for direct download is available.</p> "
```
ArtemisLav: There is an unnecessary `ul>` in the end of the second to last bullet in 4. and perhaps for 5, you could change it to `<strong>Go to 2.</strong>` to match the rest

katilp: Whoops, thanks, I should have seen that. No fixed to <ul> as intended

Issue #2151, opened 2017-12-14
records: CMS MC 2012 datasets


tiborsimko:
Add CMS MC 2012 datasets based on #1188.

Add corresponding RECO and HLT configuration files.

PR #2152, opened 2017-12-14
records: CMS 2012 simulated datasets and configuration files

tiborsimko:
* Adds CMS 2012 simulated datasets and the corresponding configuration files.
  (addresses #2151)


tiborsimko:
@katilp @ArtemisLav Needs work to complete some fields, but you can have the first look.
We could even deploy on DEV if it heps with finalising the analysis examples and the documentation...

PR #2167, opened 2017-12-15
records: fix for HTML tag in 5501

ArtemisLav:


ArtemisLav:
@katilp Done.

PR #2193, opened 2017-12-18
records: CMS 2012 simulated dataset generation information

### LINKED ISSUE #2151: records: CMS MC 2012 datasets

tiborsimko:
* Adds generation information to CMS 2012 simulated datasets. (closes #2151)

* Adds new configuration files accordingly. (Beware, record IDs may have
  changed.)

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-configuration-files-2012.json
+++ b/cernopendata/modules/fixtures/data/records/cms-configuration-files-2012.json
@@ -49,6 +49,54 @@
   },
   {
     "abstract": {
+      "description": "The configuration file used in SIM data processing step."
+    },
+    "accelerator": "CERN-LHC",
+    "cms_confdb_id": "0d0714743f0204ed3c0144941eb3ca01",
+    "collaboration": {
+      "name": "CMS collaboration",
+      "recid": "451"
+    },
+    "collections": [
+      "CMS-Configuration-Files"
+    ],
+    "collision_information": {
+      "energy": "8TeV",
+      "type": "pp"
+    },
+    "date_created": "2012",
+    "date_published": "2017",
+    "distribution": {
+      "formats": [
+        "py"
+      ],
+      "number_files": 1,
+      "size": 5015
+    },
+    "experiment": "CMS",
+    "files": [
+      {
+        "checksum": "sha1:77558fd0580cb6f896cfe804cc17f1bea5c1e982",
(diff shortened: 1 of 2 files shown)


tiborsimko:
@katilp Regarding MC generation:

(1) I have found quite a few cases when the parent dataset generation configuration is missing:

- example: simulated dataset `/QCD_Pt_30to50_CTEQ6L1_8TeV_herwig6/Summer12_DR53X-NoPileUp_START53_V19-v1/AODSIM`
- ... for which I can find RECO and HLT configurations well
- its parent is `/QCD_Pt_30to50_CTEQ6L1_8TeV_herwig6/Summer12-START50_V13-v2/GEN-SIM`
- ... which has no `byoutputdataset` configuration from where I am taking the configuration (but has several `byinputdataset` configuration files)
- ... which has no parent itself (no `GEN`)

Can you please check this concrete use case on the CMS DAS web interface?

In case like this, I'm currently leaving the parent step out, which gives only the following:

mcg

(2) Note also that the information about release and global tag for each step is also missing due to empty `mcm dataset` output.

katilp:
I can confirm that DAS web interface does not show the grand-parent either.
Does any of /QCD_Pt_*to*_CTEQ6L1_8TeV_herwig6/Summer12_DR53X-NoPileUp_START53_V19-v1/AODSIM
have it?
If yes, an educated user could look up the generator parameters from the one which has it because they only differ by the pt cut. But we have to come up with a proper way of showing it (maybe not for the release)

katilp:
It seems also that this a very particular dataset (looking at the configs). Are there also other types of datasets (other than` /QCD_Pt_*to*_CTEQ6L1_8TeV_herwig6/Summer12_DR53X-NoPileUp_START53_V19-v1/AODSIM`) which do not have parents properly?

tiborsimko:
> Does any of /QCD_Pt_*to*_CTEQ6L1_8TeV_herwig6/Summer12_DR53X-NoPileUp_START53_V19-v1/AODSIM have it?

None of them.

> Are there also other types of datasets (other than/QCD_Pt_*to*_CTEQ6L1_8TeV_herwig6/Summer12_DR53X-NoPileUp_START53_V19-v1/AODSIM) which do not have parents properly?

Yes, I'll send you a concise list.

PR #2243, opened 2018-01-08
docs: GitHub issue and PR templates

tiborsimko:

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,11 @@
+Please start by describing concisely the nature of the bug or the feature request.
+
+In case of a bug report, please describe the steps how the problem may be reproduced. Does the problem occur on the main [CERN Open Data](http://opendata.cern.ch/) portal? Does it work on the [development](http://opendatadev.cern.ch/) server? If you are a developer, can you reproduce it locally?
+
+In case of a feature request, please provide the motivation and the concrete use case which your request is seeking to address.
+
+Finally, if you are a member of the ``cernopendata`` team, please select appropriate labels to indicate the topic and the complexity of the task, as well as select the concrete milestone if applicable.
+
+This message should be replaced with the description of the issue at hand.
+
+Thank you.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,22 @@
+Please start by choosing an appropriate base branch for your pull request.
+
+- ``production`` branch: quick fixes for problems observed on production service, fixing minor bugs, improving existing documentation or record metadata;
+
+- ``master`` branch: introducing new features.
+
+Please reference all issues that this PR addresses or closes. Do not forget to include ``(addresses #123)`` or ``(closes #123)`` in your commit log message.
+
+Please run the tests locally in order to speed up developments and save Travis CI cycles.  Look for ``./run-tests.sh``.
+
+Finally, if you are a member of the ``cernopendata`` team, please select an appropriate issue labels and the concrete milestone to ease our housekeeping.
+
+Checklist:
+- [ ] Verified that the system builds and works locally as described?
+- [ ] Added tests to ensure success and verified failures?
+- [ ] Updated the documentation?
+- [ ] Described the nature of the change in a self-understandable commit log
+      message?
+
+This message should be replaced with the description of the pull request at hand.
+
+Thank you.


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
+Please start by describing concisely the nature of the bug or the feature
+request.
+
+In case of a bug report, please describe the steps how the problem may be
+reproduced. Does the problem occur on production servers? Does it work on the
```
RaoOfPhysics: If we expect non-technical people to submit issues (and I hope they do!), it would be good to limit the use of developer-speak. :) Can we say, "Does the problem occur on (1) the CERN Open Data portal or (2) if you are a developer, a local build?" Or something to that effect?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+development server? Can you reproduce it locally on the `master` branch?
```
RaoOfPhysics: I would limit mention of branches here, again. Most of the issues come from us on the team, and we may not all be building the instance to test it.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+
+In case of a feature request, please provide the motivation and the concrete use
+case which your request is seeking to address.
```
RaoOfPhysics: 👍

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+
+Finally, before opening the issue, please select appropriate labels to indicate
+the topic and the complexity of the task. Please select the concrete milestone
+if applicable.
```
RaoOfPhysics: Can non-members select labels? I don't think so. Maybe we shouldn't confuse people, so I would say, "If you are a member of the `cernopendata` team, please make sure you select appropriate labels…"

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+
+This message should be replaced with the description of the issue at hand.
```
RaoOfPhysics: Use comments in the template (more info in my main Review):

```markdown
[//]: # "==================
Please provide a description of the issue at hand below.
Thank you.
=========================="
```

This will appear in the issue creation, but won't render even if the user leaves it in.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,29 @@
+Please start by choosing an appropriate base branch for your pull request.
```
RaoOfPhysics: I would make the next two points here bullets.

- ``production`` branch: Quick fixes to problems or documentation enhancement
- ``master`` branch: Introduction of new features

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,29 @@
...
+bug, improving existing documentation or record metadata? Make your pull request
+against the ``production`` branch.
+
+Introducing a new feature? Make your pull request against the ``master`` branch.
+
+Please reference all issues that this PR addresses or closes. Do not forget to
+include ``(addresses #123)`` or ``(closes #123)`` in your commit log message.
+
+Please run the tests locally in order to speed up developments and save Travis
+CI cycles.  Look for ``./run-tests.sh``.
```
RaoOfPhysics: Again, not all of us will run these tests, so how do we balance this?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,29 @@
...
+
+Finally, before submitting the pull request, please select an appropriate issue
+labels and the concrete milestone to ease our housekeeping.
```
RaoOfPhysics: Labels and milestones can only be assigned by team members. Remove mention?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,29 @@
...
+
+Checklist:
+- [ ] Verified that the system builds and works locally as described?
+- [ ] Added tests to ensure success and failure?
+- [ ] Updated the documentation?
+- [ ] Described the nature of the change in a self-understandable commit log
+      message?
+- [ ] Provided all information to facilitate review and ensure fast issue flow?
```
RaoOfPhysics: This last point is a bit vague.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,29 @@
...
+
+This message should be replaced with the description of the pull request at
+hand.
```
RaoOfPhysics: Use comments (see main Review comment).

.github/PULL_REQUEST_TEMPLATE.md
RaoOfPhysics: Also, what happens if I submit a PR for some minor changes? I won't need to check these, right? Do we expect many non-CERN developers contributing? Maybe we can remove this checklist or explicitly state it's for developers? I don't expect documentation or metadata changes to need all these items.

.github/ISSUE_TEMPLATE.md
tiborsimko: Hmm, I'm not sure about using comments... E.g. some people may think of updating text blob within comment, not paying attention to the arcane syntax around, which could lead to not seeing *any* visible impact, as it were.  My whole template was basically meant as a helper text, just to guide people which things to check first.... and the whole text was meant to be replaced by the submitter.

.github/PULL_REQUEST_TEMPLATE.md
tiborsimko: It wouldn't apply in that case... but somebody (i.e. the integrator) will have to run the tests in those cases.

.github/PULL_REQUEST_TEMPLATE.md
tiborsimko: Used again the introductory "if you are a member..." that should address this too. Although we very rarely get pull requests from outside the organisation... I recall only a couple of examples from the past.

.github/PULL_REQUEST_TEMPLATE.md
tiborsimko: The checklist is mostly for developers.  It does not hurt to home in the message which things should be usually done before we make PRs 😄  We'll see whether we retain the checklist "as is" with each PR, so that the author will tick items as appropriate for the reviewer to see about the status, or whether this is only for the initial information of the authorr who would be free to delete the whole blob.

PR #2253, opened 2018-01-11
records: fix pagination issues

### LINKED ISSUE #2242: Only first page of file index files shows in records
E.g. in
http://opendata.cern.ch/record/6004
only first 5 file index list files is shown, next button does not bring to anywhere (no matter if I click on the box or on the text/number - see also #2238)

pamfilos:
* closes #2242

### DIFF
--- a/cernopendata/modules/records/utils.py
+++ b/cernopendata/modules/records/utils.py
@@ -153,9 +153,12 @@ def record_file_page(pid, record, page=1, **kwargs):
         return jsonify(grouped_files)
 
     file_type_filter = request.args.get('type')
-    if file_type_filter is 'index':
-        filtered_files = [d for d in rf if (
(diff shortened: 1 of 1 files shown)


cernopendata/modules/records/utils.py
```diff
@@ -153,9 +153,12 @@ def record_file_page(pid, record, page=1, **kwargs):
...
     file_type_filter = request.args.get('type')
-    if file_type_filter is 'index':
-        filtered_files = [d for d in rf if (
-                          d.get('type', "") in ['index', 'index.txt'])]
+
+    if file_type_filter == 'index_files':
+        filtered_files = [d
+                          for d in rf
+                          if (d.get('type', "")
+                              in ['index', 'index.txt', 'index.json'])]
```
tiborsimko: This would show also the `index.json` files that are to be hidden...

Issue #2259, opened 2018-01-16
CMS: 2011-2012 collision data records, links wrong


katilp:
The link redirection does not work (see also #2250. #2257). Please:

In all 2011 collision data records

http://opendata.cern.ch/search?page=1&size=20&type=Dataset&subtype=Collision&experiment=CMS&year=2011

- [x] replace opendata.cern.ch/VM/CMS#2011 with http://opendata.cern.ch/docs/cms-virtual-machine-2011

- [x] replace opendata.cern.ch/getting-started/CMS#2011 with http://opendata.cern.ch/docs/cms-getting-started-2011

In all 2012 collision data records

http://opendata.cern.ch/search?page=1&size=20&type=Dataset&subtype=Collision&experiment=CMS&year=2012

- [x] replace http://opendata.cern.ch/getting-started/cms/2011 with http://opendata.cern.ch/docs/cms-getting-started-2011

PR #2285, opened 2018-02-05
records: update links

atrisovic:
* Closes: #2259


cernopendata/modules/fixtures/data/records/cms-primary-datasets-Run2012B.json
```diff
@@ -223,7 +223,7 @@
       },
       {
         "description": "Getting started with CMS open data",
-        "url": "/getting-started/cms/2011"
+        "url": "/docs/cms-getting-started-2011"
```
tiborsimko: It seems better to me to upgrade the `/getting-started/...` redirection logic rather than the records metadata, ecause there may be many users out there using the old published URLs... Can you have a look at that?

atrisovic: ok :)

PR #2286, opened 2018-02-05
records: ALICE reconstructed dataset index files

tiborsimko:
- dcfa3e964bea8e46ee39f4744e9e0b34e9f6e036 fixtures: removal of trailing whitespace
- 2c13f6f6a4ac1567aed4108cf63ba3e7e39f497e records: ALICE reconstructed dataset index files
- 62b63a0fc7530d435c999cd8cc09650543e05ffe records: ALICE derived dataset files


tiborsimko:
Since all ALICE files are named the same, the manual file download box does not look that great, for example for record 1102:

alice-files

We have two solutions:

1. Keep the files named in this way and amend the UI file display box in order display the least recognisable path, i.e. the filename, or if not unique the filename plus parent directory, or if not unique add also grandparent, etc.

2. Amend the displayed file name to something unique, i.e. using the dataset name, which would give `LHC10h_PbPb_ESD_139038_0001_ALIESDs.root`, `LHC10h_PbPb_ESD_139038_0002_ALIESDs.root`, etc.  (This would be only for the UI file box display; the files would remain the same on EOS.)  

Let's check with ALICE the preferences.

tiborsimko:
> Let's check with ALICE the preferences.

OK to go for the standard and rich index list files. OK for solution 1 that will come in another PR.

PR #2295, opened 2018-02-06
records: add schema.org json-ld serializer

### LINKED ISSUE #2273: records: add schema.org serialiser
Introduce schema.org serialiser that outputs in JSON-LD.  (Stemmed from #1324.)

We can start with simple mappings (Datasets, Software) and leave harder mappings (configuration files? Other?) for later. 

See also:

- cernopendata's `DataCiteSerializer` 
- `zenodo/modules/records/serializers/schemaorg.py`

### LINKED ISSUE #2275: records: inject JSON-LD
Once we have schema.org serialiser (#2273), enrich detailed record pages by injecting JSON-LD with it there.

Consider (1) injecting full content directly to `/record/NNN` page body vs (2) injecting only reference to other URL exposing JSON-LD such as `/record/NNN/export/FFF`.

Start with simple metadata author/title/type etc for the first prototype, but think of the later use case about how to deal with numerous file assets.  (Some datasets can have more than 10,000 files to download, so some sort of pagination would be nice to have perhaps.)

hjhsalo:
* Add schema.org JSON-LD serializer
  (closes #2273)
* Add schema.org JSON-LD description of Record to detail page of
  Dataset and Software Records.
  (closes #2275)

### DIFF
--- a/cernopendata/modules/records/serializers/schemaorg.py
+++ b/cernopendata/modules/records/serializers/schemaorg.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of CERN Open Data Portal.
+# Copyright (C) 2018 CERN.
+#
+# CERN Open Data Portal is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# CERN Open Data Portal is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with CERN Open Data Portal; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+"""CERN Opendata schema.org JSON-LD serializer."""
+
+
+from invenio_records_rest.serializers.json import JSONSerializer
+
+from cernopendata.modules.records.serializers.schemas \
+    import schemaorg_schemas as schemas
+
+
+class CODSchemaorgSerializer(JSONSerializer):
+    """CERN Open Data schema.org serializer.
+
+    Serializes a Record based on it's type (Dataset, Software, etc.) to
+    schema.org compatible JSON-LD syntax.
+    """
+
+    SCHEMAS = {
+        'Dataset': schemas.DatasetSchemaorgSchema,
+        'Software': schemas.SoftwareSchemaorgSchema,
+        'Environment': schemas.RecordSchemaorgSchema,
+        'Supplementaries': schemas.RecordSchemaorgSchema,
+        'Documentation': schemas.DocumentationSchemaorgSchema,
+    }
+    """Marsmallow Schemas for each Record type in CERN Open Data."""
+
+    def dump(self, obj, context=None):
+        """Serialize object with schema."""
+        schema_cls = None
+
+        obj = obj.get('metadata', obj)
+        type = obj['type']['primary']
+
+        schema_cls = self.SCHEMAS[type]
+
+        return schema_cls(context=context).dump(obj).data
--- a/cernopendata/modules/records/serializers/schemas/schemaorg_schemas.py
+++ b/cernopendata/modules/records/serializers/schemas/schemaorg_schemas.py
@@ -1,0 +1,151 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of CERN Open Data Portal.
+# Copyright (C) 2018 CERN.
+#
+# CERN Open Data Portal is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# CERN Open Data Portal is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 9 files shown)


hjhsalo:
Test for example with `curl` or check sources of rendered detail page of a record.
Note that for Documentation, schema.org JSON-LD is intentionally not included in the rendered page.

Dataset Record
```
$ curl -H "Accept: application/ld+json" http://127.0.0.1:5000/api/records/3806
```

Software Record
```
$ curl -H "Accept: application/ld+json" http://127.0.0.1:5000/api/records/1000
```

Environment Record
```
$ curl -H "Accept: application/ld+json" http://127.0.0.1:5000/api/records/560
```

Supplementaries Record
```
$ curl -H "Accept: application/ld+json" http://127.0.0.1:5000/api/records/3132
```

Documentation Record
```
http://127.0.0.1:5000/docs/cms-virtual-machine-2010/export/jsonld
```

cernopendata/modules/fixtures/cli.py
```diff
@@ -190,10 +190,10 @@ def records(skip_files, files, profile, mode):
         record_json = glob.glob(os.path.join(data, '*.json'))
 
     for filename in record_json:
-        name = filename.split('/')[-1]
-        if name.startswith('opera'):
-            click.echo('Skipping opera records ...')
-            continue
+        # name = filename.split('/')[-1]
```
tiborsimko: Since we are merging to `qa`, please leave the OPERA skipping in place...

hjhsalo: Sorry, I picked up unnecessary commit(s) from master. Fixed.

cernopendata/modules/records/serializers/schemas/schemaorg_schemas.py
```diff
@@ -0,0 +1,149 @@
...
+    def get_creator(self, obj):
+        """Get creator(s) based on authors or collaboration field."""
+        authors = obj.get('authors', None)
+
+        if authors:
+            return [{"@type": "Person", 'name': x['name']} for x in authors]
+        else:
+            # Note: Can there be many Organizations in a Record?
+            #       Now only one is expected.
+            collaboration = obj.get('collaboration', None).get('name', None)
```
tiborsimko: `None` does not have `get`, so this could fail if the `collaboration` key is not in the object.  It should be safer to use:

```python
collaboration = obj.get('collaboration', {}.get('name', None)
```

Also, instead of the second `None`, maybe empty string might be clearer there... as if indicating the usual "output type", as it were...

cernopendata/modules/records/serializers/schemaorg.py
```diff
@@ -0,0 +1,58 @@
...
+"""CERN Opendata schema.org JSON-LD serializer."""
+
+
+from invenio_records_rest.serializers.json import JSONSerializer
+
+from cernopendata.modules.records.serializers.schemas \
+    import schemaorg_schemas as schemas
+
+
+class OpendataSchemaorgSerializer(JSONSerializer):
```
tiborsimko: Elsewhere we use `class COD...` e.g. in mistune.  Would be nice to standardise it.

cernopendata/modules/records/serializers/schemaorg.py
```diff
@@ -0,0 +1,58 @@
...
+    """CERN Open Data schema.org serializer.
+
+    Serializes a Record based on it's type (Dataset, Software, etc.) to
+    schema.org compatible JSON-LD syntax.
+    """
+
+    SCHEMAS = {
+        'Dataset': schemas.DatasetSchemaorgSchema,
```
tiborsimko: If we are worried that we intermix "JSON schemas" and "Marshmallow schemas" here and there in the codebase, so that `git grep` will become hard to read, we can think of inventing e.g. "MSchema" for the latter concept, but I'm not sure it would be worth the editing trouble...

cernopendata/modules/records/serializers/schemas/schemaorg_schemas.py
```diff
@@ -0,0 +1,149 @@
...
+    id_ = fields.Method('get_identifier', dump_to='@id')
+
+    # Minimum functional
+    type_ = fields.Method('get_type', dump_to='@type')
+    name = fields.Str(attribute='title', dump_to='name')
+    description = fields.Method('get_description', dump_to='description')
+    identifier = fields.Method('get_identifier', dump_to='identifier')
+    url = fields.Method('get_url', dump_to='url')
+    creator = fields.Method('get_creator', dump_to='creator')
+    date_created = fields.Str(attribute='date_published',
```
tiborsimko: Both `date_created` and `date_published` are taken from the same field.  We should probably use:

```python
date_created = fields.Str(attribute='date_created',
                         dump_to='dateCreated')
```

which will give the data-taking year there...

cernopendata/modules/records/serializers/schemas/schemaorg_schemas.py
```diff
@@ -0,0 +1,149 @@
...
+        return None
+
+    def get_type(self, obj):
+        """Get resource type based on Record type."""
+        return "CreativeWork"
+
+    def get_url(self, obj):
+        """Get url based on recid field."""
+        recid = obj.get('recid', None)
+        return "http://opendata.cern.ch/record/{}".format(recid)
```
tiborsimko: We should probably not hard-code `opendata.cern.ch`, but use the base URL of the deployed instance here, so that if somebody is using the QA service, the link would stay pointing to QA?

cernopendata/modules/records/serializers/schemas/schemaorg_schemas.py
tiborsimko: ... otherwise things work well locally 👍

hjhsalo: Actually, this was intentional, but I cannot anymore remember why :)
I changed to `date_created`.

cernopendata/modules/records/serializers/schemas/schemaorg_schemas.py
hjhsalo: Introduced a `SITE_URL` config variable which is used. Didn't want to use `request.url_root` as it might give funny/incorrect values and it wouldn't be available if we do serializing outside request context.
Probably an application wide config variable wouldn't be a bad idea.

PR #2296, opened 2018-02-06
sitemap: add sitemap package

### LINKED ISSUE #2274: sitemap: add sitemap
Introduce sitemap to ease record discovery.  (Stemmed from #1324.)  

See also:

- `Flask-Sitemap`
- `zenodo/modules/sitemap`

Since we don't have much dynamic uploads, rather planned releases, we could even pre-generate the sitemap files during image build time, without having to have the worker node do this "live" regularly.

diegodelemos:
* (closes #2274).

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -46,6 +46,8 @@ TEMPLATES_AUTO_RELOAD = DEBUG
 # Piwik tracking code: set None to disabled it
 THEME_PIWIK_ID = os.environ.get('PIWIK_ID', None)
 
+SITE_URL = os.environ.get('CERNOPENDATA_SITE_URL', 'opendata.cern.ch')
+
 # Assets
 # ======
 #: Switch of assets debug.
--- a/cernopendata/modules/sitemap/__init__.py
+++ b/cernopendata/modules/sitemap/__init__.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of CERN Open Data Portal.
+# Copyright (C) 2018 CERN.
+#
+# CERN Open Data Portal is free software; you can redistribute it and/or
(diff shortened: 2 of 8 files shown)


diegodelemos:
To test:
```console
$ ./scripts/populate-instance.sh
$ mkdir cernopendata/modules/sitemap/static
$ cernopendata sitemap generate -o cernopendata/modules/sitemap/static
...
Writing sitemap 0 file: cernopendata/modules/sitemap/static/sitemap0.xml ...
Writing sitemap index file: cernopendata/modules/sitemap/static/sitemap.xml ...
```

jirikuncar:
Hi @diegodelemos, just out of curiosity, is there any problem with [Flask-Sitemap](https://github.com/inveniosoftware/flask-sitemap)?

tiborsimko:
@diegodelemos Yes, let's remove the CLI.  The number of records on COD3 remains relatively small (~4,000), so that the sitemap generation remains quick.  (I guess we won't need to include direct links to files there one day, because that would be ~400,000!)

@jirikuncar The guys tried on Zenodo some time ago, but get into some issues (IIRC doing sitemap in memory etc) so they went for "bibsched task" like solution with a nightly celery task generating the sitemaps. Otherwise I fully agree that we should revive the `Flask-Sitemap` package or make a general `invenio-sitemaps` one...

ioannistsanaktsidis:
@diegodelemos please rebase so as to check and merge it directly :).

tiborsimko:
@ioannistsanaktsidis I wouldn't approve it if it did not work for me locally 😄 Taking care of merging...

ioannistsanaktsidis:
@tiborsimko yes I was going to test it, but I saw that you approved already before I comment. Anyway will test it locally either way :)

PR #2314, opened 2018-02-08
records: CMS 2011 collision datasets rich index files

tiborsimko:
* Adds rich index files (TXT, JSON) to `cms-primary-datasets-Run2011A`
  records. (addresses #2093)

* Removes empty record (coming from a conversion of a deleted dataset that was
  never officially released).

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-primary-datasets-Run2011A.json
+++ b/cernopendata/modules/fixtures/data/records/cms-primary-datasets-Run2011A.json
@@ -31,17 +31,31 @@
   "experiment": "CMS",
   "files": [
     {
-      "checksum": "sha1:a0c48ed090c7e5d88d8a4c73b5e99f2ab188f957",
+      "checksum": "adler32:70cf4527",
+      "description": "BTag AOD dataset file index (1 of 2) for access to data via CMS virtual machine",
+      "size": 270,
+      "type": "index.json",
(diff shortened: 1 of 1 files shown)


tiborsimko:
@katilp While working on rich file indexes, I have checked the full file list for 2011 collision datasets and everything is OK except one file missing:

- `/eos/opendata/cms/Run2011A/HT/AOD/12Oct2013-v1/20000/76CFA6F2-0A40-E311-8EF4-0026189438AA.root`

Based on my notes it was definitely there in 2015...

Otherwise we could also gain some disk space by removing unreleased (but transferred) 2011 collision datasets:

- ForwardTriggers
- MinimumBias0
- MinimumBias1
- MinimumBias2

Please let me know.

katilp:
@tiborsimko btw, for the disk space, was the full 2PB which was agreed for 2012 release already allocated, I should I ask for that?

katilp:
For the missing file, thanks for checking! We had also the other one from RAW of 2012, maybe add an issue and I will check if I get them transferred (or copied). Phedex transfers only by blocks or datasets, not by files.

tiborsimko:
@katilp An issue for the missing file added (#2324). The CMS EOS space should be OK I guess, because it is occupied at the 75% level only -- but please note that one CMS-only quota volume is not visible to me.  (You can try `eos quota /eos/opendata/cms` on LXPLUS.)  If you give me green light, I could delete those ForwardTriggers/MinimumBias0,1,2 files, but it is not really critical at the moment.

PR #2319, opened 2018-02-08
docs: suggestion to developing.rst

atrisovic:
Added instructions to clean the instance.

### DIFF
--- a/DEVELOPING.rst
+++ b/DEVELOPING.rst
@@ -22,6 +22,13 @@ integration purposes. A usage example:
    $ firefox http://0.0.0.0:5000/
    $ docker-compose -f docker-compose-dev.yml down
 
+However, when you are rerunning the server it is necessary to first clean the instance:
+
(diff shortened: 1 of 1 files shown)


DEVELOPING.rst
```diff
@@ -22,6 +22,13 @@ integration purposes. A usage example:
    $ firefox http://0.0.0.0:5000/
    $ docker-compose -f docker-compose-dev.yml down
 
+However, when you are rerunning the server it is necessary to first clean the instance:
```
tiborsimko: What do you mean by "rerunning the server"?  Perhaps better formulated as "Note that if you want to reupload some files, you can use..."  but we should move this out of the "Installation" section (that should remain short) into its own section.  I have also some tips I could add regarding cache etc....

atrisovic: ah ok!

atrisovic:
Superseded by #2323

PR #2321, opened 2018-02-08
records: CMS 2010 collision datasets rich index files

tiborsimko:
* Adds rich index files (TXT, JSON) to `cms-primary-datasets` records.
  (addresses #2093)

* Adds previously forgotten 10000 and 10001 volumes to the MuOnia dataset.

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-primary-datasets.json
+++ b/cernopendata/modules/fixtures/data/records/cms-primary-datasets.json
@@ -31,45 +31,87 @@
   "experiment": "CMS",
   "files": [
     {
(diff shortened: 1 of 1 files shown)


tiborsimko:
CC @katilp I run a file check here as well, and I discovered that I had apparently forgotten to include two MuOnia volumes (10000 and 10001) in the COD2 times... so I have added them here.  

This means that there are two new index files for MuOnia:

https://eospublichttp.cern.ch/eos/opendata/cms/Run2010B/MuOnia/AOD/Apr21ReReco-v1

PR #2323, opened 2018-02-08
docs: suggestion to developing.rst

atrisovic:


DEVELOPING.rst
```diff
@@ -35,6 +35,16 @@ mounted in the container in this case. A usage example:
    $ firefox http://0.0.0.0/
    $ docker-compose -f docker-compose-dev.yml down -v
 
+Development notes
+======================
+
+Please note that when restarting the server (eg. in order to upload something), it is necessary to first clean the instance:
```
tiborsimko: The wording "restarting the server" is not fully apropriate for the example that follows, because one can restart the COD3 instance like this:

```console
$ docker-compose -f docker-compose-dev.yml restar
```

(It helps with getting the XRootD connectivity back in case of wifi network changes for example.)

DEVELOPING.rst
```diff
@@ -35,6 +35,16 @@ mounted in the container in this case. A usage example:
    $ firefox http://0.0.0.0/
    $ docker-compose -f docker-compose-dev.yml down -v
 
+Development notes
+======================
```
tiborsimko: `=================`

atrisovic: 👍

PR #2329, opened 2018-02-09
views: sanitise arguments

atrisovic:

Please start by choosing an appropriate base branch for your pull request.

- ``production`` branch: should not be used for pull requests; it is fast-forwarded to `qa` branch at chosen points in time for production deployments;

- ``qa`` branch: shorter-term rollout nature (think a few days); used for fixes and new features observed on the production service; should be a good default branch for most pull requests;

- ``master`` branch: longer-term rollout nature (think a few weeks); bigger developments that may need longer stabilisation periods and iterations with the experiments.

Please reference all issues that this PR addresses or closes. Do not forget to include ``(addresses #123)`` or ``(closes #123)`` in your commit log message.

Please run the tests locally in order to speed up developments and save Travis CI cycles.  Look for ``./run-tests.sh``.

Finally, if you are a member of the ``cernopendata`` team, please select an appropriate issue labels and the concrete milestone to ease our housekeeping.

Checklist:
- [x] Verified that the system builds and works locally as described?
- [ ] Added tests to ensure success and verified failures?
- [ ] Updated the documentation?
- [x] Described the nature of the change in a self-understandable commit log
      message?

This message should be replaced with the description of the pull request at hand.

Thank you.

### DIFF
--- a/cernopendata/modules/pages/views.py
+++ b/cernopendata/modules/pages/views.py
@@ -33,7 +33,6 @@ from flask import Blueprint, abort, current_app, escape, jsonify, redirect, \
     render_template, request, url_for
 from flask_babelex import lazy_gettext as _
 from flask_breadcrumbs import default_breadcrumb_root, register_breadcrumb
-from flask_menu import register_menu
 from jinja2.exceptions import TemplateNotFound
 from speaklater import make_lazy_string
 
@@ -163,7 +162,10 @@ def about():
 @blueprint.route('/about/<exp>')
 def about_exp(exp):
     """Render about <experiment> pages."""
-    return redirect('/docs/about-{}'.format(exp))
+    if exp in ['lhcb', 'atlas', 'cms', 'alice', 'opera']:
+        return redirect('/docs/about-{}'.format(exp))
+    else:
+        abort(404)
 
 
 @blueprint.route('/about/cms-pileup-simulation')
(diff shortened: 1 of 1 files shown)


cernopendata/modules/pages/views.py
```diff
@@ -190,21 +192,37 @@ def getting_started_redirect(exp):
...
 
 
 @blueprint.route('/vm/<exp>/<year>')
 def vm_redirect_year(exp, year):
     """Redirects to associated experiment."""
-    return redirect('/docs/%s-virtual-machine-%s' % (exp, year), code=302)
+    if check_year(year) and exp in ['lhcb', 'cms', 'alice']:  # FIXME add exp
+        return redirect('/docs/%s-virtual-machine-%s' % (exp, year), code=302)
+    else:
+        abort(422)
```
tiborsimko: Works well for me, but instead of 422, what about simply returning 404?  Accessing `/vm/xyzzy` is not different from accessing `/foo`; both don't exist.

tiborsimko: Works well for me, but instead of 422, what about simply returning 404?  Accessing `/vm/xyzzy` is not different from accessing `/foo`; both don't exist.

tiborsimko: (sorry for the double click 😉 due to slower connection; the comment appplies obviously to all 422 in the patch)

atrisovic: Ah ok, 422 returns `Unprocessable Entity` I thought that's equivalent to 'wrong parameters' 😄

PR #2356, opened 2018-03-21
api: secure api programmaticaly

ioannistsanaktsidis:
WARNING: To be checked first , especially the part with invenio-accounts.

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -31,7 +31,7 @@ import os
 from invenio_records_files.api import _Record
 from invenio_records_rest.config import RECORDS_REST_ENDPOINTS
 from invenio_records_rest.facets import terms_filter
-from invenio_records_rest.utils import allow_all
+from invenio_records_rest.utils import allow_all, deny_all
 
 from cernopendata.modules.pages.config import *
 from cernopendata.modules.theme.config import *
@@ -55,6 +55,7 @@ SITE_URL = os.environ.get('CERNOPENDATA_SITE_URL', 'opendata.cern.ch')
 #: Switch of automatic building.
 # ASSETS_AUTO_BUILD = True
 
+ACCOUNTS_REGISTER_BLUEPRINT = False
 
 # Static file
 COLLECT_STORAGE = 'flask_collect.storage.link' \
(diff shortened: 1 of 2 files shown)


psaiz:
Wait, this is to merge directly in qa. @tiborsimko : could you please confirm that this is correct?

tiborsimko:
> Wait, this is to merge directly in qa. @tiborsimko : could you please confirm that this is correct?

I think we can just inspect how the new Invenio situation handles this sort of security in "master" branch, and adapt it accordingly. It should not be necessary to integrate this concrete patch into "qa" because (i) Invenio had been quite updated in the meantime and (ii) we'll soon have "master" generally available again. Hence I thought checking only new "master" in this respect.

psaiz:
I'm still missing what this thing is meant to improve. If I use the endpoint to `PUT`, `POST`, I get `405 Not Allowed`. The GET works, as expected:
```
 % curl -X POST https://opendata-dev.cern.ch/api/records/21570 -d '{"hello":"world"}'
<html>
<head><title>405 Not Allowed</title></head>
<body>
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx</center>
</body>
</html>
 % curl -X POST https://opendata-dev.cern.ch/api/records/1000 -d '{"hello":"world"}'
<html>
<head><title>405 Not Allowed</title></head>
<body>
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx</center>
</body>
</html>
 % curl -X GET https://opendata-dev.cern.ch/api/records/1000 -d '{"hello":"world"}' 
{"created":"2023-11-29T16:15:59.912622+00:00","id":"1000","links":{"bucket":"https://opendata-dev.cern.ch/api/files/34d0c8b5-607b-4921-9a04-12184530d03b","self":"https://opendata-dev.cern.ch/api/records/1000"}
```
Am I doing the right kind of queries?

setup.py
```diff
@@ -34,7 +34,7 @@
 # Get the version string. Cannot be done with import!
 g = {}
 with open(os.path.join('cernopendata', 'version.py'), 'rt') as fp:
-    exec(fp.read(), g)
+    exec (fp.read(), g)
```
zzacharo, line 37: is this space correct?

cernopendata/config.py
```diff
@@ -55,6 +55,7 @@
 #: Switch of automatic building.
 # ASSETS_AUTO_BUILD = True
 
+ACCOUNTS_REGISTER_BLUEPRINT = False
```
zzacharo, line 58: this is needed eventually as the default is None and it is checked here https://github.com/inveniosoftware/invenio-accounts/blob/04d5171d9e508835f4c54998e4f327443929b75d/invenio_accounts/ext.py#L147

psaiz: Thanks @zzacharo. Let me create a new PR only with this change then.

psaiz:
Addressed  in https://github.com/cernopendata/opendata.cern.ch/commit/cb6f05ba375adee53d428c1afebf332cf61e7757

Issue #2247, opened 2018-01-09
docs: add instructions on programmatic file download


tiborsimko:
Add instructions presenting REST API and telling users about persistent URLs and how to best download files via (1) canonical record URLs, (2) via EOSPUBLIC HTTP gateway, and (3) via XRootD protocol.

PR #2364, opened 2018-03-23
robots: introduces robots.txt

### LINKED ISSUE #700: system: introduce `robots.txt`
A sensible `robots.txt` should be introduced.

ioannistsanaktsidis:
* Closes #700.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -26,6 +26,7 @@ recursive-include cernopendata *.png
 recursive-include cernopendata *.scss
 recursive-include cernopendata *.svg
 recursive-include cernopendata *.tpl
+recursive-include cernopendata *.txt
 recursive-include cernopendata *.md
 recursive-include cernopendata *.xml
 recursive-include nginx *.conf
--- a/cernopendata/modules/pages/views.py
+++ b/cernopendata/modules/pages/views.py
@@ -30,7 +30,7 @@ import json
 
 import pkg_resources
 from flask import Blueprint, abort, current_app, escape, jsonify, redirect, \
-    render_template, request, url_for
+    render_template, request, url_for, Response
 from flask_babelex import lazy_gettext as _
 from flask_breadcrumbs import default_breadcrumb_root
 from jinja2.exceptions import TemplateNotFound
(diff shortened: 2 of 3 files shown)


robots.txt
```diff
@@ -0,0 +1,2 @@
+User-agent: *
+Allow: /
```
tiborsimko: There is no `Allow` field in the robots specs as per http://www.robotstxt.org/robotstxt.html ... 

Let's see whether we have some set of pages to restrict; that would be the cleanest solution to the no-allow problem.  

Maybe `/api/...` URLs, provided that we expose things via `/record/NNN/export/...` and `/record/NNN/files/FFF` kind of URLs?

tiborsimko: Depending on the #2247, yes, I'd say let's disallow `/api` for now.

tiborsimko:
We also need to plug this to the app route map, so that `opendata.cern.ch/robots.txt` would reply :wink: Please move the file somewhere under `static` maybe and please define a new view endpoint that would serve it.

tiborsimko:
(1) Looks good, but the file is exported as if it were HTML, not as plain text:

```console
$ wget -q -S --spider http://0.0.0.0:5000/robots.txt
  HTTP/1.0 200 OK
  Content-Type: text/html; charset=utf-8
  Content-Length: 29
  Server: Werkzeug/0.12.2 Python/2.7.5
  Date: Mon, 09 Apr 2018 08:26:43 GMT
```

Can you please change the content type?

(2) Also, I've seen that one can expose sitemap via `robots.txt`: https://www.sitemaps.org/protocol.html#submit_robots

```
Sitemap: http://www.example.com/sitemap.xml
```

Let's perhaps add that?  Although beware of QA, DEV, and PROD base URLs... we'd have to do that from within Pythonic app then.

ioannistsanaktsidis:
@tiborsimko so should I remove the old implementation and add the sitemap one ?

tiborsimko:
The old implementation is OK, we just need one more line with the `Sitemap` line (and fix the content type of the output). You are right, we can hard-code the PROD URL `http://opendata.cern.ch/`, no need to go into Pythonic app.

Regarding for QA and DEV, we should forbid everything... I can amend while merging up, We don't want these two development sites to be crawled and indexed.

ioannistsanaktsidis:
@tiborsimko everything is updated :)

PR #2368, opened 2018-03-29
fixtures: split missing config files

annatrz:


tiborsimko:
```diff
-        "files": [
+        "files":[
```

It would be good to keep the whitespace around operands; only the trailing whitespace is supressed.

Can you please amend the touched files with something like the following?

```python
import json

recordfile = 'cernopendata/modules/fixtures/data/records/cms-configuration-files-2012.json'

records = json.loads(open(recordfile, 'r').read())

new_content = json.dumps(records, indent=2, sort_keys=records)
for line in new_content.split('\n'):
    line = line.rstrip()
    print(line)
```

(untested, but that's basically what I'm using in some JSON file manipulations...)

PR #2400, opened 2018-04-19
records: abstracts for OPERA tau candidate events

### LINKED ISSUE #2382: records: create 9 tau candidate event records
Create 9 tau candidate event records, following up #2351.

ArtemisLav:
* Adds abstracts to the OPERA tau neutrino candidate event records (closes #2382).


tiborsimko:
Thanks, I had a quick look at the produced abstracts and found two issues:

- "The tau..." in the beginning of third paragraph in 12254000036 is not fully escaped.

- The description of 12123032048 has the initial paragraph twice and the 2nd paragraph is missing.  Also please check the ending that have some paragraphs missing.

tiborsimko:
> In 12123032048 at the end, I only found one "T" missing. Is there more?

Nope, that's all I think, I see that "A fit was performed..." section is well inside the paragraph, so we sholud be good.

Issue #2419, opened 2018-07-11
ORCID in COD


ArtemisLav:
- Add ORCID as a metadata field in the schema (record-v1.0.0.json)

For NEW records, when the author also submits an ORCID iD:
- a new element must be added to the DOI metadata when minting the DOI via the DataCite Metadata Store as part of `creator`, namely `<nameIdentifier nameIdentifierScheme="ORCID" schemeURI="http://orcid.org/">0000-XXXX-XXXX-XXXX</nameIdentifier>`

This is related to the CAP "PID Integrations" Milestone.

PR #2430, opened 2018-08-15
records: enrichment of data model (record schema)

### LINKED ISSUE #1529: data model: enrich JSON schema information
Once we have full examples for all collections, we should take time to enrich JSON schemas to do things like:

- [ ] mark fields as mandatory or optional
- [ ] describe the meaning of all fields

ArtemisLav:
* Adds descriptive text to give context to all fields
* Sets required properties (closes #1529)
* Adds `orcid` as a field (addresses #2419)


ArtemisLav:
The descriptions might require some tweaking. The most important part here is to agree on which fields need to be mandatory.

diegodelemos:
From https://pypi.org/project/locust/. 
> Use [locustio](https://pypi.org/project/locustio/) instead.

@ArtemisLav looks like we should have been using `locustio` package instead of `locust`, can you change in a different commit the `setup.py` to use it?
```diff
tests_require = [
    'check-manifest>=0.25',
    'coverage>=4.0',
    'isort>=4.2.2',
-   'locust>=0.8',
+   'locustio>=0.8',
    'mock>=1.3.0',
``` 
This should solve the problem.

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -61,57 +66,71 @@
...
     },
     "authors": {
       "items": {
         "properties": {
           "affiliation": {
+            "description": "The affiliation of the author",
             "type": "string"
           },
           "ccid": {
+            "description": "The author's ccid, if applicable",
```
tiborsimko: The "if applicable" bit could apply to virtually all the fields... I think it is better to leave these out everywhere.

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -132,47 +153,58 @@
       "type": "array"
     },
     "date_created": {
+      "description": "The data-taking year",
```
tiborsimko: Note that some fields such as `date_created` etc are applicable also to non-dataset records, so that seeing "data-taking year" for a software record might look bizarre.  It may be good to explain this.

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -132,47 +153,58 @@
...
     },
     "generator": {
       "properties": {
         "global_tag": {
+          "description": "The global tag for the generator",
           "type": "string"
         },
         "names": {
           "items": {
+            "description": "The name/s of the generator",
```
tiborsimko: `names` is a list, so you can use plural

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -182,16 +214,19 @@
     },
     "keywords": {
       "items": {
+        "description": "Relevant keywords for the record",
         "type": "string"
       },
       "type": "array"
     },
     "language": {
+      "description": "The language/s of the resource, based on ISO 639-2",
```
tiborsimko: Do we allow more than one language? I don't think we do, it is a string in the schema, hence singular is enough.

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -201,15 +236,19 @@
       "items": {
         "properties": {
           "description": {
+            "description": "A brief description of the link",
             "type": "string"
           },
           "recid": {
+            "description": "If the link is to another Open Data record, the record id of that record",
```
tiborsimko: Can you update the case everywhere, e.g. `ID` instead of `id`, `URL` instead of `url`

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -220,21 +259,26 @@
...
         },
         "links": {
           "items": {
             "properties": {
               "description": {
+                "description": "A brief description of the link associated with the methodology",
                 "type": "string"
               },
               "recid": {
+                "description": "If the link is to another Open Data record, the record id of that record",
```
tiborsimko: "another CERN Open Data portal record"  would be more precise regarding internal vs external open data links

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -248,21 +292,26 @@
     "note": {
       "properties": {
         "description": {
+          "description": "Any other notes that need to be displayed prominently in the record",
```
tiborsimko: "displayed" is a secondary motivation here, better say something like content-related, e.g. just "Notes of interest for the given resource" or something of the kind

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -304,42 +362,53 @@
     "prepublication": {
       "properties": {
         "date": {
+          "description": "The prepublication date, based on ISO 8601 (YYYY-MM-DD)",
```
tiborsimko: I like the notion of format `YYYY-MM-DDD` it may be useful to repeat this elsewhere...  as you do with `8TeV`... then the documentation is more readable standalone even without seeing record examples

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -486,5 +576,6 @@
       "type": "object"
     }
   },
-  "type": "object"
+  "type": "object",
+  "required": ["accelerator", "collections", "date_created", "date_published", "experiment", "license", "publisher", "recid", "title", "type"]
```
tiborsimko: Are you sure that all the records have these fields? E.g. the author list records don't have `accelator`, so we cannot require it.  

Here is a quick simple way to check:

```console
$ cd cernopendata/modules/fixtures/data/records
$ grep -c '"accelerator"' * | grep :0$
atlas-learning-resources.json:0
cms-author-list-2012.json:0
cms-author-list.json:0
cms-author-list-Run2011A.json:0
opera-author-list-multiplicity.json:0
opera-author-list-tau.json:0
```

... meaning that we cannot make it mandatory.   Or else we would have to amend the records.

(I have some more detailed JSON Schema checks, but that would require to install some Python libraries...) detailed check o...

ArtemisLav: So, I checked everything and I had to take out `abstract.description`, `accelerator`, `date_created` and `license`. There are also 3 files that don't have the `date_published` property, but I'll add it there, because I think it should definitely be required. 

For `authors.name` and `collaboration.name`, perhaps we could keep them by using something like `Oneof`? 

Also, it would be better if we could add `license` everywhere, since it's missing in a lot of records. Would this be easy to do programmatically? It should be just adding CC0 in most, if not all, these records.

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -64,54 +69,67 @@
...
           "inspireid": {
+            "description": "The author's inspire ID",
             "type": "string"
           },
           "name": {
+            "description": "The name of the author",
             "type": "string"
           }
         },
+        "required": ["name"],
```
tiborsimko: Getting validation errors on `required` parts:

```console
$ python ../data-curation/cod2-to-cod3/code/validate_records.py                                                                2 ↵
cms-validated-runs.json
Traceback (most recent call last):
  File "../data-curation/cod2-to-cod3/code/validate_records.py", line 15, in <module>
    jsonschema.validate(record, schema)
  File "/usr/lib/python3.7/site-packages/jsonschema/validators.py", line 540, in validate
    cls.check_schema(schema)
  File "/usr/lib/python3.7/site-packages/jsonschema/validators.py", line 83, in check_schema
    raise SchemaError.create_from(error)
jsonschema.exceptions.SchemaError: ['name'] is not of type 'boolean'

Failed validating 'type' in schema['properties']['properties']['additionalProperties']['properties']['required']:
    {'default': False, 'type': 'boolean'}

On instance['properties']['collaboration']['required']:
    ['name']
```

We are using JSON Schema version 3 in the file header, where the `required` property was only Boolean it seems https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.7

Perhaps we can upgrade to v4 everywhere:

```console
$ git grep draft
cernopendata/jsonschemas/records/data-policies-v1.0.0.json:  "$schema": "http://json-schema.org/draft-04/schema#",
cernopendata/jsonschemas/records/data-policies-v1.0.0.json:        "$schema": "http://json-schema.org/draft-04/schema#",
cernopendata/jsonschemas/records/datasets-v1.0.0.json:  "$schema": "http://json-schema.org/draft-03/schema#",
cernopendata/jsonschemas/records/docs-v1.0.0.json:  "$schema": "http://json-schema.org/draft-04/schema#",
cernopendata/jsonschemas/records/docs-v1.0.0.json:        "$schema": "http://json-schema.org/draft-04/schema#",
cernopendata/jsonschemas/records/glossary-term-v1.0.0.json:  "$schema": "http://json-schema.org/draft-04/schema#",
cernopendata/jsonschemas/records/record-v1.0.0.json:  "$schema": "http://json-schema.org/draft-03/schema#",
cernopendata/jsonschemas/records/software-v1.0.0.json:  "$schema": "http://json-schema.org/draft-03/schema#"
```

ArtemisLav: Done.

cernopendata/jsonschemas/records/record-v1.0.0.json
ArtemisLav: As we discussed IRL, we can revisit the `authors` and `collaboration` properties in the future; for now, we will only keep `experiment` as required.

tiborsimko:
Thanks, all the files seem OK expect for:

- `data-policies.json` e.g. missing the required field `publisher` (needs to be also outside `prepublication`)
- `cms-learning-resources.json` e.g. missing  the required field `date_published` (there are 7 records but only 3 have them)

PR #2432, opened 2018-08-27
CMS MC production overview

heitorPB:
Adds a new page to the portal with details on how to extract the generator's parameters of the MC datasets.


katilp:
Could you also add a screenshot on changes on a single MC record (it is hard to see in the json file)? Thanks!

heitorPB:
the dataset [VBFHiggs0MToZZTo4L_M-125p6_7TeV-JHUGenV4](http://opendata.cern.ch/record/1352) has the `sim config` with 

```
# Input source
process.source = cms.Source("PoolSource",
    secondaryFileNames = cms.untracked.vstring(),
    fileNames = cms.untracked.vstring('/store/generator/Summer11LegpLHE/GluGluTo2e2mu_SMH_M-125p6_7TeV-MCFM67-pythia6/GEN/START53_LV4-v1/00000/EEA88C0D-4E63-E411-A160-02163E0100B6.root')
)
```

That is indeed weird.

---

For this same dataset, the resulting record page now looks like this:

katilp:
That's strange indeed. Note however that this input may be a leftover and irrelevant as some of the input parameters in CMSSW config  get overwritten by the CRAB config setting, see https://twiki.cern.ch/twiki/bin/view/CMSPublic/WorkBookCRAB3Tutorial#CMSSW_parameter_set_configuratio

The link to the LHE note in the MC record is now fine, but it may be useful to add one for the SIM config parameters, especially in case when there's no LHE step (i.e. when the MC generator parameters are in SIM config only)

heitorPB:
It's missing gridpack information in the page.

tiborsimko:
@heitorPB Works well for me.  Can you please (1) rebase against latest master and (2) update the title of the new MD page to `CMS Monte Carlo production overview` so that the Overview would be in lowercase?

PR #2461, opened 2018-12-05
Create cms-trigger-information-2010.json

katilp:
This is the list of data-taking config files (HLT) for 2010 data, addresses #1201 #2450
The list includes RunA (to be released) and RunB (released) run ranges

PR #2481, opened 2018-12-12
add explanation about the HLT path definition

### LINKED ISSUE #2460: CMS: Add instructions to find how HLT trigger paths are defined
Lacking more comprehensive information, add instructions (as a separate guide, add in the trouble shooting page, else?) on how to find the HLT trigger path definitions.

A recent tentative in opendata-support (as an answer to a question about triggers for 
http://opendata.cern.ch/record/6017 (/MuOniaParked/Run2012B-22Jan2013-v1/AOD):

> For each trigger path, you can find basic information of their use range under respective records, i.e.
> for one of the trigger paths for this sample:
> 
> http://opendata.cern.ch/record/6296
> 
> which is the first record if you follow the link HLT_Dimuon10_Jpsi
> in HLT trigger path of the listing (I take this as an example)
> 
> This shows in which runs this particular trigger path has been used.
> 
> Furthermore, you can find the software definition of the trigger paths in the data-taking configuration file which is linked to the trigger path records above. You can have a look in
> a configuration file, for example in http://opendata.cern.ch/record/6123 and look for process.HLT_Dimuon10_Jpsi. 
> You'll find the "ingredients" of the trigger in question, and  you can further dig in to the file to find where the different constituents of the trigger have been defined...
> When you come down to the class instantiation such as (this is just one of the components that went to the definition of the trigger process, I take this as an example):
> 
> process.hltDimuon10JpsiL3Filtered = cms.EDFilter( "HLTMuonDimuonL3Filter",
> 
> etc, you will see the parameters used (in this case note for example MaxInvMass = cms.vdouble(3.35) and MinInvMass - cms.vdouble(2.8) to restrict the trigger around J/Psi mass), and you can then find the class e.g. HLTDMuonDiMuonL3Filter in this case in the open source CMSSW code in https://github.com/cms-sw/cmssw/tree/CMSSW_5_3_X/  and see how the algorithm is defined in
> https://github.com/cms-sw/cmssw/blob/CMSSW_5_3_X/HLTrigger/Muon/interface/HLTMuonDimuonL3Filter.h
> 
> Note also that we have a guide for using the trigger information in http://opendata.cern.ch/docs/cms-guide-trigger-system but it does not cover the definition of different triggers, but the instructions  on how to get access to the trigger decisions. 
>

katilp:
``(closes #2460)`` 
 needs (FIXME) the record id for HLT listing for 2010 from PR  #2461

### DIFF
--- a/cernopendata/modules/fixtures/data/docs/cms-guide-trigger-system/cms-guide-trigger-system.md
+++ b/cernopendata/modules/fixtures/data/docs/cms-guide-trigger-system/cms-guide-trigger-system.md
@@ -61,3 +61,29 @@ Find the software and usage instructions in
 
 - [Analysis code for extracting the trigger information from the CMS 2010 data](/record/5003)
 - [Analysis code for extracting the trigger information from the CMS 2011 data](/record/5004)
+
+### HLT Trigger Path definitions
+
+For each dataset, the possible HLT trigger paths are listed in the dataset record. Each HLT path is composed of different algorithms and they are defined in the data-taking configuration files, which are given in
+
+- [High-Level Trigger information for 2010 CMS open data](/record/FIXME)
+- [High-Level Trigger information for 2011 CMS open data](/record/1700)
+- [High-Level Trigger information for 2012 CMS open data](/record/1701)
+
+The exact parameters (such as pT or eta cuts, or else) for each component of the HLT path can be found in the data-taking configuration file.
+
+As an example, if you are looking for information about `HLT_Dimuon10_Jpsi` trigger, look for a string `process.HLT_Dimuon10_Jpsi`
+in the data-taking configuration file corresponding to the event range of your interest.
+You will find the corresponding CMSSW path, which consists of all the modules (i.e., specifically python-configured CMSSW code)
+and/or sequences (groups of modules) that make up that trigger path.
+You can then look for each of these modules or sequences in the same configuration file until your find
+the parameters with which they are configured.
+As an example, you will find one of the modules of this path, `hltDimuon10JpsiL3Filtered`, by searching for
+`process.hltDimuon10JpsiL3Filtered`and you will see that it is an instantiation of class `HLTMuonDimuonL3Filter`:
+
+`process.hltDimuon10JpsiL3Filtered = cms.EDFilter( "HLTMuonDimuonL3Filter",`
+
+followed by the values of different parameters for this algorithm.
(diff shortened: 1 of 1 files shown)


cernopendata/modules/fixtures/data/docs/cms-guide-trigger-system/cms-guide-trigger-system.md
```diff
@@ -61,3 +61,26 @@ Find the software and usage instructions in
...
+### HLT Trigger Path definitions
+
+For each dataset, the possible HLT trigger paths are listed in the dataset record. Each HLT path is composed of different algorithms and they are defined in the data-taking configuration files, which are given in 
+
+- [High-Level Trigger information for 2010 CMS open data](/record/FIXME)
+- [High-Level Trigger information for 2011 CMS open data](/record/1700)
+- [High-Level Trigger information for 2012 CMS open data](/record/1701)
+
+The exact parameters (such as pT or eta cuts, or else) for each component of the HLT path can be found in the data-taking configuration file.
+As an example, if you looking for information about `HLT_Dimuon10_Jpsi` trigger, look for a string `process.HLT_Dimuon10_Jpsi`
```
caredg: change "you looking" --> "you are looking"

cernopendata/modules/fixtures/data/docs/cms-guide-trigger-system/cms-guide-trigger-system.md
```diff
@@ -61,3 +61,26 @@ Find the software and usage instructions in
...
+in the data-taking configuration file corresponding to the event range of your interest. 
+You will find all components of this HLT path.  You can then find the parameters for each of the components
```
caredg: In the TSG world, the "components" are called modules (actually some of them are sequences).  So I would replace "You will find all components of this HLT path" and what follows with something like "You will find the corresponding CMSSW path, which consists of all the modules (i.e., specifically python-configured CMSSW code)  and/or sequences (groups of modules)  that make up that trigger path.  You can then look for each of these modules or sequences in the same configuration file until your find the parameters with which they are configured."

katilp: Better, indeed, done. In consequence, I've changed "components" to "modules" in the following phrase  as well(for the example case od hltDimuon10JpsiFiltered).

tiborsimko:
I'll fix the CI issues and repush...

PR #2486, opened 2018-12-12
record-template: changes to improve readability

### LINKED ISSUE #2470: RFC: Change "How were these data selected" to "How were these data generated" for MC?
In data records, as a  title for "methodology", we have now "How were these data selected". This is very good for collision data, but sounds somewhat odd for simulated data, as the simulated data are not selected from anywhere (if we intend with "selected" = select a subset of data from an existing bigger set of data), they are indeed generated from the start.
Is it possible to differentiate the titles between collision data and simulated data?

For simulated data, it would be better to have:
"How were these data generated"

heitorPB:
- Change "How were these data selected" to "How were these data generated" for MC. Fix #2470
- add "Recommended release for analysis:" to System details   for MC
- change display of "Characteristics" as in figure below (old to left, new to right):
  1681-diff

### DIFF
--- a/cernopendata/templates/cernopendata_records_ui/records/record_detail.html
+++ b/cernopendata/templates/cernopendata_records_ui/records/record_detail.html
@@ -101,16 +101,19 @@
     {% if record.distribution and (record.distribution.number_events or record.distribution.number_files or record.distribution.size) %}
         <div class="row">
             <div class="col-md-12">
-                <h2> Characteristics </h2>
-                <label> Dataset: </label>
+                {% if record.type.primary == 'Dataset' %}
+                <h2>Dataset Characteristics</h2>
+                {% else %}
+                <h2>Characteristics</h2>
+                {% endif %}
                 {% if record.distribution.number_events %}
-                <span><strong>{{record.distribution.number_events}}</strong> events</span>
+                <span><strong>{{record.distribution.number_events}}</strong> events. </span>
                 {% endif %}
                 {% if record.distribution.number_files %}
-                <span><strong>{{record.distribution.number_files}}</strong> files</span>
+                <span><strong>{{record.distribution.number_files}}</strong> files. </span>
                 {% endif %}
                 {% if record.distribution.size %}
-                <span><strong>{{record.distribution.size | filesizeformat }}</strong> in total</span>
+                <span><strong>{{record.distribution.size | filesizeformat }}</strong> in total.</span>
                 {% endif %}
             </div>
         </div>
@@ -131,7 +134,7 @@
             {% if record.system_details.release %}
                 {% if record.type.primary == 'Software' %}
                 <label>Software release:</label>
-                {% elif 'aod' in record.distribution.formats %}
+                {% elif 'aod' in record.distribution.formats or 'aodsim' in record.distribution.formats %}
                 <label>Recommended release for analysis:</label>
                 {% elif 'raw' in record.distribution.formats %}
                 <label>Recommended release for reconstruction:</label>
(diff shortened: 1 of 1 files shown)


cernopendata/templates/cernopendata_records_ui/records/record_detail.html
```diff
@@ -101,13 +101,12 @@ <h2>Related Datasets</h2>
     {% if record.distribution and (record.distribution.number_events or record.distribution.number_files or record.distribution.size) %}
         <div class="row">
             <div class="col-md-12">
-                <h2> Characteristics </h2>
-                <label> Dataset: </label>
+                <h2>Dataset Characteristics </h2>
```
tiborsimko: We cannot use `Dataset` here, since the template is used for all types of records. E.g. see http://opendata.cern.ch/record/250

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
```diff
@@ -101,13 +101,12 @@ <h2>Related Datasets</h2>
...
                 {% if record.distribution.number_events %}
-                <span><strong>{{record.distribution.number_events}}</strong> events</span>
+                <span><strong>{{record.distribution.number_events}}</strong> events</span><br>
                 {% endif %}
                 {% if record.distribution.number_files %}
-                <span><strong>{{record.distribution.number_files}}</strong> files</span>
+                <span><strong>{{record.distribution.number_files}}</strong> files</span><br>
```
tiborsimko: The line break here makes the page longer, and since we have enough horizontal space, let's rather use it...

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
```diff
@@ -131,7 +130,7 @@ <h2>System Details</h2>
             {% if record.system_details.release %}
                 {% if record.type.primary == 'Software' %}
                 <label>Software release:</label>
-                {% elif 'aod' in record.distribution.formats %}
+                {% elif 'aod' in record.distribution.formats or 'aodsim' in record.distribution.formats %}
```
tiborsimko: :+1:

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
```diff
@@ -176,8 +175,12 @@ <h2>URLs</h2>
...
             {% elif record.type.primary == 'Environment' %}
             <h2>How was this built?</h2>
-            {% else %}
-            <h2>How were these data selected?</h2>
+            {% elif record.type.primary == 'Dataset' %}
+                {% if 'Simulated' in record.type.secondary %}
+                <h2>How were these data generated?</h2>
+                {% else %}
+                <h2>How were these data selected?</h2>
+                {% endif %}
```
tiborsimko: Beware, there is no final `else` clause that would display something for non-dataset types we might have.

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
heitorPB: But `Dataset` always appears in the template, even for the VM Image: `Dataset: 1 files 15.2 MB in total`

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
heitorPB: Good point. Also, for `Software` we have `Data Selection` as `<h2>` see [this record](http://opendata-dev.cern.ch/record/234). Should we change this?

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
heitorPB: +1

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
tiborsimko: We could, although I am not sure what the good general heading title would be... At least in your concrete example, the record at hand says very early "Event selection" so it reads still nicely as it is.

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
tiborsimko: Yes, it is better to say "Data:" there, which is more generic... or perhaps even nothing at all and start simply with "25465895 events 2814 files 2.6 TB in total".

Issue #2529, opened 2019-03-25
models: add steps under methodology


mantasavas:
Amend JSON schema to facilitate new changes for placing steps in methodology. For more information see: #2465 and #2506.

tiborsimko:
Closed by #2530

PR #2530, opened 2019-03-25
models: add generation steps within methodology

mantasavas:
Adding dataset generation steps under methodology in data model (schema), closing #2529

### DIFF
--- a/cernopendata/jsonschemas/records/record-v1.0.0.json
+++ b/cernopendata/jsonschemas/records/record-v1.0.0.json
@@ -286,6 +286,75 @@
           "description": "A description of the methodology used for the production of this data/software",
           "type": "string"
         },
+        "steps": {
+          "items": [
+            {
+              "properties": {
+                "configuration_files": {
+                  "items": [
+                    {
+                      "properties": {
+                        "script": {
+                          "description": "Production script",
+                          "type": "string"
+                        },
+                        "title": {
+                          "description": "Title of the script",
+                          "type": "string"
+                        },
+                        "cms_confdb_id": {
+                          "description": "the cms_confdb_id (applies to CMS records)",
+                          "type": "string"
+                        },
+                        "process": {
+                          "description": "Indicating the process level (e.x. HLT, RECO etc.)",
+                          "type": "string"
+                        },
+                        "recid": {
+                          "description": "The internal ID of the related record, if it is another Open Data record",
+                          "type": "string"
+                        },
+                        "url": {
+                          "description": "External link to the configuration file",
+                          "type": "string"
+                        }
+                      },
+                      "type": "object"
+                    }
+                  ],
+                  "type": "array"
+                },
+                "generators": {
+                  "items": [
+                    {
+                      "description": "Generator used in production of dataset",
+                      "type": "string"
+                    }
+                  ],
+                  "type": "array"
+                },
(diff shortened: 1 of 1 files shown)


cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -286,6 +286,71 @@
...
+            {
+              "properties": {
+                "configuration_files": {
+                  "items": [
+                    {
+                      "properties": {
+                        "script": {
+                          "description": "A cmsdriver script",
+                          "type":"string"
+                        },
```
katilp: As this will contain the full production script including all commands for the job that was run for production (with cmsrel .... , cd ...., and then cmsdriver [parameters]), maybe we could call it "Production script".

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -286,6 +286,71 @@
...
+                    }
+                  ],
+                  "type":"array"
+                },
+                "generators": {
+                  "items": [
+                    {
+                      "description": "Generator used to in producing dataset",
+                      "type":"string"
+                    }
```
katilp: Generator used in production of dataset ?

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -286,6 +286,71 @@
...
+                  "type":"array"
+                },
+                "global_tag": {
+                  "description": "The global tag for generator",
+                  "type":"string"
+                },
+                "release": {
+                  "description": "CMS software used in releasing data",
+                  "type":"string"
+                },
```
katilp: To which CMSSW version does this refer? If it is the one used in production (i.e. the one which can be found in CMSDAS or in McM), then this would be "CMS software version used in production"

cernopendata/jsonschemas/records/record-v1.0.0.json
tiborsimko: Also, sometimes it will contain genfragment blobs, not only cmsdriver blobs, so we'd need a rather generic description.

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -286,6 +286,71 @@
...
+                        },
+                        "cms_confdb_id": {
+                          "description": "the cms_confdb_id (applies to CMS records)",
+                          "type":"string"
+                        },
+                        "process": {
+                          "description": "Indicating the process level (e.x. HLT, RECO etc.)",
+                          "type":"string"
+                        },
+                        "recid": {
```
tiborsimko: "The internal ID of the related record, if it is another Open Data record"

see the description string for `recid` elsewhere in the data model

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -286,6 +286,71 @@
...
+                },
+                "type": {
+                  "description": "Format of the dataset",
+                  "type":"string"
+                }
+              },
+              "type":"object"
+            }
+          ],
+          "type":"array"
```
tiborsimko: Cosmetics: please put spaces after ':', that is `"type": "array"` etc

PR #2546, opened 2019-04-01
fixtures: amend methodology steps

### LINKED ISSUE #2544: fixtures: amend 2012 `methodology.steps` for MC datasets
Add steps under methodology in fixtures (references #2506)

mantasavas:
Still missing parsed description into JSON (script can be used: references https://github.com/cernopendata/data-curation/issues/27)
* closes #2544


tiborsimko:
The `configuration_files` field is missing the `recid` so that links can be resolved.  (It is present in the HTML you deleted, but not in the generated JSON.)  It would be good to add it.

PR #2549, opened 2019-04-02
records: add record for Higgs candidate events

### LINKED ISSUE #1410: Visualization of Higgs candidates in Run2012 datasets
There are many Higgs candidate events which have been noted and documented that exist in the Run2012 public datasets. It would be nice to extract them into event display files for visualization. Some of the events (Htautau) have already been extracted in the course of usual event display production:

https://twiki.cern.ch/twiki/bin/view/CMSPublic/Hig13004PubTWiki#Event_Display

There are varied and rich event topologies available in:

- [ ] Hbb
- [ ] Htautau
- [ ] HWW
- [ ] HZZ
- [ ] Hgammagamma

ArtemisLav:
* (closes #1410)

Signed-off-by: Artemis Lavasa <artemis.lavasa@cern.ch

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-derived-csv-2012.json
+++ b/cernopendata/modules/fixtures/data/records/cms-derived-csv-2012.json
@@ -1,0 +1,69 @@
+{
+  "abstract": {
+    "description": "<p>This document contains Higgs candidate events that pass the selections found in the example Higgs-to-four-lepton analysis of CMS public data.</p> <p>These data were selected for use in education and outreach and contain a subset of the total event information. They are not suitable for a full physics analysis.</p>"
+  },
+  "accelerator": "CERN-LHC",
+  "authors": [
+    {
+      "name": "McCauley, Thomas",
+      "orcid": "0000-0001-6589-8286"
+    }
+  ],
+  "collections": [
+    "CMS-Derived-Datasets"
+  ],
+  "date_published": "2019",
+  "distribution": {
+    "formats": [
+      "csv",
+      "ig",
+      "json"
+    ]
+  },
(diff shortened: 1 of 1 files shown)


cernopendata/modules/fixtures/data/records/cms-derived-csv-2012.json
```diff
@@ -0,0 +1,69 @@
...
+    }
+  ],
+  "license": {
+    "attribution": "CC0"
+  },
+  "methodology": {
+    "description": "These events were selected in the example Higgs-to-four-lepton analysis of CMS public data."
+  },
+  "publisher": "CERN Open Data Portal",
+  "recid": "FIXME",
```
tiborsimko: We can take 5200, the next hundreds above `-*-cvs-*.json` files

cernopendata/modules/fixtures/data/records/cms-derived-csv-2012.json
```diff
@@ -0,0 +1,69 @@
...
+  ],
+  "date_published": "2019",
+  "distribution": {
+    "formats": [
+      "csv",
+      "ig",
+      "json"
+    ]
+  },
+  "doi": "10.7483/OPENDATA.CMS.FIX.ME",
```
tiborsimko: Maybe leave DOI fully empty for now, so that we avoid clashes.  We can always add one later.

ArtemisLav: I "reserved" a specific one; shall I remove it? It's not minted yet.

tiborsimko: If you reserved, please use the real one.... While we shall deploy on DEV only for now, we can already use the target one...

PR #2564, opened 2019-04-04
ui: file preview modal "copy to clipboard" button

### LINKED ISSUE #2561: Previewer: preview windows in new data records do not allow copying
It is not possible to copy anything (ctrl-c, mouse double-click, other...) from the preview windows for cmsDriver script or genfragment  i.e in http://opendata-dev.cern.ch/record/9790

preview

pamfilos:
* closes #2561

### DIFF
--- a/cernopendata/templates/cernopendata_records_ui/records/detail.html
+++ b/cernopendata/templates/cernopendata_records_ui/records/detail.html
@@ -150,7 +150,7 @@ $('#download_caution').on('show.bs.modal', function (event) {
         modal.find('#download-msg-uri').attr("href", file_uri)
 })
 
-var recordApp = angular.module('recordApp', ['ui.codemirror', 'ui.bootstrap' ]);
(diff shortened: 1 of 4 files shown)


cernopendata/templates/cernopendata_records_ui/records/detail.html
```diff
@@ -338,29 +338,32 @@ <h4> Export </h4>
...
 
+    $ctrl.msg = null;
     $scope.contents = content;
 
     $scope.codemirrorLoaded = function(_editor){
-      setTimeout(function(){
-        _editor.refresh()
-      });
+      $ctrl.copyEditor = function() {
+        clipboard.copyText(content);
```
tiborsimko: This seems to copy the full cmsDriver script etc.  Is it possible to copy only the parts user selectied? Use case: the script may be big and the person only wants to copy the highlighted part.

If it is too time consuming, we can merge this already :+1:

Issue #2447, opened 2018-10-27
CMS: Run2 QCD MC for data science jettuples


katilp:
In connection with #2440, this issue follows the jettuples to be produced from run2 AOD samples, to be made available on the portal.

The datasets:

Data science jettuples (contact Kimmo Kallonen HIP):
- /QCD_Pt-15to7000_TuneCUETP8M1_Flat_13TeV_pythia8/RunIISummer16MiniAODv2-PUMoriond17_magnetOn_80X_mcRun2_asymptotic_2016_TrancheIV_v6-v1/MINIAODSIM
   - Release: CMSSW_8_0_21 Global Tag: 80X_mcRun2_asymptotic_2016_TrancheIV_v6

To do:

- [x] transfer the MINIAODSIM dataset to T3_CH_CERN_OpenData 
   - transfer request [done](https://cmsweb.cern.ch/phedex/prod/Data::Subscriptions#state=create_since%3D1538039295%3Bnode%3D2241%3Brequest%3D1499073)  ok
- [x] prepare the data record for  MiniAODSIM see #2525
- [x] test that the production workflow works on CMS Open Data VM
   - the VM with encapsulated slc6  #2426 can be used (it is not yet on the portal record)
   - the correspondig GT is needed (#2443), now available (note the changes needed to read the GT on the config file + symbolic links: cfr http://opendata.cern.ch/docs/cms-guide-for-condition-database) 
- [ ] test that the production workflow works on ReANA (@alintulu)
- [x] prepare the CODP record for the production workflow
   - have it in a github repo (see with @caredg ): https://github.com/cms-legacydata-analyses/JetNtupleProducerTool
   - provide the corrsponding metadata for the CODP record (see #2584 )
- [x] produce the jettuple files and transfer to eospublic upload (see with @tiborsimko)
   - they will be in Kimmo's public area  to be copied over
- [x] provide the metadata for the CODP records for thejettuple files (see https://github.com/cernopendata/opendata.cern.ch/issues/2447#issuecomment-477700348 below)
  - NB the variable description is in https://github.com/cms-legacydata-analyses/JetNtupleProducerTool/blob/2016/README.md
- [x] if appropriate, provide an example of usage as a software record

For contributions, see also https://github.com/cernopendata/opendata.cern.ch/wiki/Contributing-content-to-CERN-Open-Data

katilp:
The path to files can be found through
[...]

PR #2582, opened 2019-04-08
records: fixes for record 12100

ArtemisLav:
* addresses #2447.


ArtemisLav:
Need to address a 'FIXME' in methodology. Connected to #2584.

Issue #2451, opened 2018-10-30
CMS: Run2010A collision data to be released


katilp:
The rest of 2010 data, RunA datasets, now to be released are
`dataset=/*/Run2010A-Apr21ReReco-v*/AOD`
https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset%3D%2F*%2FRun2010A-Apr21ReReco-v*%2FAOD

12  datasets - NB in RunB ZeroBias no events pass the validated events filter, but for RunA it is OK

```
/ZeroBias/Run2010A-Apr21ReReco-v1/AOD
/MuOnia/Run2010A-Apr21ReReco-v1/AOD
/MuMonitor/Run2010A-Apr21ReReco-v1/AOD
/Mu/Run2010A-Apr21ReReco-v1/AOD
/MinimumBias/Run2010A-Apr21ReReco-v1/AOD
/JetMETTauMonitor/Run2010A-Apr21ReReco-v1/AOD
/JetMETTau/Run2010A-Apr21ReReco-v1/AOD
/JetMET/Run2010A-Apr21ReReco-v1/AOD
/EGMonitor/Run2010A-Apr21ReReco-v1/AOD
/EG/Run2010A-Apr21ReReco-v1/AOD
/Commissioning/Run2010A-Apr21ReReco-v1/AOD
/BTau/Run2010A-Apr21ReReco-v1/AOD
```

To do

- [x] transfer the datasets
   - transfer request done https://cmsweb.cern.ch/phedex/prod/Data::Subscriptions#state=create_since%3D1542124296%3Bnode%3D2241%3Brequest%3D1557490 Transfer completed!
- [x] prepare the short description of the contents (as in Run2010B)
   - E.g.

>  Events stored in this primary dataset were selected because of presence of at least one high-energy jet in the event.

- [x] extract the metadata and provenance from CMSDAS (see #2450)

katilp:
@tiborsimko With the help of transfer teams, the last pending datasets have now been completed and the datasets above can be moved to `/eospublic/cms`

katilp:
To do:
- move files from upload
- create file listing indexes

tiborsimko:
Number of files and size on EOS:

```
/BTau/Run2010A-Apr21ReReco-v1/AOD 2067   2193478806310
/Commissioning/Run2010A-Apr21ReReco-v1/AOD 4716   5252574106904
[...]

Issue #2450, opened 2018-10-27
CMS: provenance information (configs) is missing for 2010 collision data records


katilp:
2011 and 2012 collision datasets have the full provenance information available in the record:s
![pdprov](https://user-images.githubusercontent.com/5859918/47604284-c7616b00-d9f7-11e8-9a69-95f8236223a7.JPG)
2010 collision datasets do not have it: http://opendata.cern.ch/record/14

Now, as the rest of 2010 data will be released and some RAW data samples will be added for selected 2010 AODs, add the information also to the 2010 records:
- [x] extract from confdb the config files listed in https://fwyzard.web.cern.ch/fwyzard/hlt/2010/summary
   - Run2010B configs in https://cernbox.cern.ch/index.php/s/SsvXZ08Ijj5fZbb (starting from runs 141956 - 142137), 
   - Run2010A configs (+ HI) now in https://cernbox.cern.ch/index.php/s/bvfSIUcioFAmYns
- [x] prepare a summary page with the run number ranges, CMSSW versions and trigger menu configs from https://fwyzard.web.cern.ch/fwyzard/hlt/2010/summary and prepare a record like in http://opendata.cern.ch/record/1700
   - NB: All the tags mentioned there are already available in https://github.com/cms-sw/cmssw-cvs github repository  e.g. https://github.com/cms-sw/cmssw-cvs/tree/CMSSW_3_8_5_onlpatch3
   - source for the 1701 page is: https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/modules/fixtures/data/records/cms-trigger-information-2012.json
   - Done (cms-trigger-information-2010.json in katilp/sandbox), record numbers for config files to be updated, PR #2461
- [x] list the possible triggers for each collision dataset (take the lists from https://fwyzard.web.cern.ch/fwyzard/hlt/2010/dataset)
- [x] extract the rereco config files from CMSDAS 
  - not available: [cmsdas query](https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=config+dataset%3D%2FJet%2FRun2010B-Apr21ReReco-v1%2FAOD) ?
  - production listed in https://twiki.cern.ch/twiki/bin/view/CMS/Collisions2011Analysis and announced in  https://hypernews.cern.ch/HyperNews/CMS/get/datasets/924.html
[...]

PR #2595, opened 2019-04-11
Adds Run2010A records

katilp:
- Adds 12 Run2010A primary data records, addresses #2451 
- Adds a methodology description for BTau Run2010A sample, addresses #2450 
   - needs a link to the record from PR #2461
   - if it shows OK, the same text can be added to all RunA and  RunB of 2010 in this files


tiborsimko:
Looks good :+1: I'll add record IDs and remove some FIXMEs so that we can deploy on DEV and I'll add the file listings later

Issue #2439, opened 2018-10-05
CMS: Add RAW samples for existing AOD datasets


katilp:
Add RAW samples of O(Mevent) corresponding to the existing CMS AODs

- [x] Select the samples (priority list Single/Double Electron/Muon Jets, 2012, 2011, (2010))
  - RunA 2011 run period 160404 to 173692:
    - [RAW for /SingleMu/Run2011A-12Oct2013-v1/AOD](https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset%3D%2FSingleMu%2FRun2011A-v1%2FRAW) - http://opendata.cern.ch/record/32 ok
    - [RAW for /SingleElectron/Run2011A-12Oct2013-v1/AOD](https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset%3D%2FSingleElectron%2FRun2011A-v1%2FRAW) - http://opendata.cern.ch/record/31 ok
    - [RAW for /DoubleElectron/Run2011A-12Oct2013-v1/AOD](https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset%3D%2FDoubleElectron%2FRun2011A-v1%2FRAW) - http://opendata.cern.ch/record/16 ok
    - [RAW for /DoubleMu/Run2011A-12Oct2013-v1/AOD](https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset%3D%2FDoubleMu%2FRun2011A-v1%2FRAW) - http://opendata.cern.ch/record/17 ok
    - [RAW for /Jet/Run2011A-12Oct2013-v1/AOD](https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset%3D%2FJet%2FRun2011A-v1%2FRAW) - http://opendata.cern.ch/record/21 ok
  - RunB 2012, run period 193833 to 196531:
    - [RAW for /MinimumBias/Run2012B-22Jan2013-v1/AOD](https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset%3D%2FMinimumBias%2FRun2012B-v1%2FRAW) http://opendata.cern.ch/record/6013 ok
[...]

PR #2596, opened 2019-04-12
Add Run1 RAW

katilp:
Adds Run1 RAW - addresses #2439 
  - needs recid's and DOIs for (4 new record 2010, 5 new records 2011, 6 new records 2012)
  - needs config files added, (edit:) get them from the list in the comment below
  - needs n of files and size

Adds also the methodology and a link to trigger configuration for 2010 RAW (in the same way as now in 2010 AOD)

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-raw-datasets-Run2010B.json
+++ b/cernopendata/modules/fixtures/data/records/cms-raw-datasets-Run2010B.json
@@ -1,0 +1,310 @@
+[
+  {
+    "abstract": {
+      "description": "<p>A sample from MinimumBias primary dataset in RAW format from RunB of 2010. Run range [146428,146514].</p> <p>This dataset contains selected runs from 2010 RunB. The list of validated lumi sections, which must be applied to all analyses, can be found in</p>",
+      "links": [
+        {
+          "recid": "1000"
+        }
+      ]
+    },
+    "accelerator": "CERN-LHC",
+    "collaboration": {
+      "name": "CMS collaboration",
+      "recid": "450"
+    },
+    "collections": [
+      "CMS-Primary-Datasets"
+    ],
+    "collision_information": {
+      "energy": "7TeV",
+      "type": "pp"
+    },
+    "date_created": [
+      "2010"
+    ],
+    "date_published": "2019",
+    "distribution": {
+      "formats": [
+        "root",
+        "raw"
+      ],
+      "number_events": 2451242
+    },
+    "experiment": "CMS",
+    "license": {
+      "attribution": "CC0"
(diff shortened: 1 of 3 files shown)


katilp:
Config files for each dataset are to be copied from (not linked to as this will move) :
- 2010 Electron: https://raw.githubusercontent.com/cms-legacydata-validation/RAWToAODValidation/2010/code/Electron/configDump_raw_Electron10.py
- 2010 Jet: https://raw.githubusercontent.com/cms-legacydata-validation/RAWToAODValidation/2010/code/Jet/configDump_raw_Jet10.py
- 2010 MinimumBias: https://raw.githubusercontent.com/cms-legacydata-validation/RAWToAODValidation/2010/code/Mu/configDump_raw_Mu10.py

(2010 config files are fully expanded because parts for estimating the luminosity - not needed in re-reco phase and not functinal in the open data environment - had to be removed in various positions in the file)

- 2011 DoubleElectron: https://raw.githubusercontent.com/cms-legacydata-validation/RAWToAODValidation/2011/code/DoubleElectron/raw_DoubleElectron11.py
- 2011 DoubleMu: https://raw.githubusercontent.com/cms-legacydata-validation/RAWToAODValidation/2011/code/DoubleMu/raw_DoubleMu11.py
- 2011 Jet: https://raw.githubusercontent.com/cms-legacydata-validation/RAWToAODValidation/2011/code/Jet/raw_Jet11.py
- 2011 SingleElectron: https://raw.githubusercontent.com/cms-legacydata-validation/RAWToAODValidation/2011/code/SingleElectron/raw_SingleElectron11.py
- 2011 SingleMu: https://raw.githubusercontent.com/cms-legacydata-validation/RAWToAODValidation/2011/code/SingleMu/raw_SingleMu11.py

for 2012 datasets, extensive rereco testing has not been done, only a quick verification, and we may want to indicate this with a generic config (without the input file name whereas the configs above have the example root file in the input file path that actually has been tested and compared to the original AOD)

katilp:
@tiborsimko If the config file above are attached to the record, what would be the appropriate json field to describe them? Probably `usage`?

tiborsimko:
In the "usage" field we only have "description" and "links" properties, so we'd need to add a new one to store config snippets, like we did for "script" in the "methodology" part... Alternatively, if some of them can become big, we can store them all as files on EOS, as for the collision data configs.

katilp:
2010 configs are very big as they are fully expanded, it would not be practical to show then like we do for "script"

katilp:
All the config files mentioned above are now uploaded in
`/eos/opendata/cms/upload/kati/RAWtoAODconfigs`

Issue #2588, opened 2019-04-08
CMS: SW record for ML sample production (using tracker-hit-enriched Run1 AOD)


katilp:
From #2444 
The structure can be as in http://opendata-dev.cern.ch/record/5104

Needed:
- title: TrackerRecHitProducerTool - an example workflow to add tracker hits to CMS Run1 AOD @emanueleusai The title should probably be extended to include the ML production part
- authors: Emanuele Usai @emanueleusai do we use the same list here as for the resulting ML samples #2575 
- @emanueleusai pls complete the brief description below in json
- tags/keywords: Software, Tool
- the datasets that are used as input: None as this explains to full production chain, or maybe the tracker-hit-enriched AODSIM #2526 can be indicated as an input for the ML production part? 
- System details 
   - VM recid 252
   - CMSSW version: CMSSW_5_3_32 (check)
- how can you use, the github repo is in 
https://github.com/cms-legacydata-analyses/TrackerRecHitProducerTool
but it will be a placeholder as it will be moved to https://github.com/cms-opendata-analyses/TrackerRecHitProducerTool and have a release tag there
- licence @emanueleusai do you agree with GNU General Public License (GPL) version 3

@ArtemisLav @tiborsimko  :
- this ML SW record produces datasets which will be on the portal, the ones in preparation in #2526 and #2575 
   - do we have an appropriate json field for that?
   - it should have as description "The ouput datasets of step2 and step3 of this software tool is available in:"

json could be:

```
{
  "created": "FIXME", 
  "id": FIXME, 
  "metadata": {
    "$schema": "http://opendata.cern.ch/schema/records/record-v1.0.0.json", 
    "abstract": {
      "description": " <p> An example workflow to produce CMS Run1 AODSIM enriched with tracker RecHits. All steps from event generation to reconstruction (step0-step2) are provided, and the tool to extract the information in a format appropriate to ML studies (step3) is included. The code can be run inside the CMS Open Data environment</p>"
    }, 
    "accelerator": "CERN-LHC", 
    "authors": [
      {
        "name": "Usai, Emanuele"
      }
FIXME, other authors as in #2575
    ], 
    "collections": [
      "FIXME Is this needed/necessary, can it be removed: CMS-Validation-Utilities"
    ], 
    "control_number": "FIXME", 
    "date_created": [
      "2019"
    ], 
    "date_published": "2019", 
    "distribution": {
      "formats": [
        "gz"
      ], 
      "number_files": 1, 
[...]

Issue #2575, opened 2019-04-07
CMS: Record for ML samples (tracker-hit-enriched Run1 AOD)


katilp:
From #2444
Prepare the record for the root/h5 output.

This will appear as a separate record (similar to http://opendata-dev.cern.ch/record/12100).
The underlying json will be in cms-derived-Run1-datascience.json
(to be created in https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/modules/fixtures/data/records/)

If you are comfortable with the json description, you can suggest the contents directly to @ArtemisLav in json format (see in http://opendata-dev.cern.ch/record/12100/export/json), or you can give the content to the different fields here.

Needed in particular:

- Dataset name from @emanueleusai
- Authors are Emanuele Usai, Michael Andrews, Bjorn Burkle, Sergei Gleyzer, Meenakshi Narain, orcid?
- @emanueleusai will provide a description (for "abstract")
- Dataset characteristics (@emanueleusai can provide N events/entries - @tiborsimko will check n files)
- Dataset semantics table from @emanueleusai
- Related datasets from #2526
 - "How were these data generated?" (or produced) with a link to SW record to be created from https://github.com/cms-legacydata-analyses/TrackerRecHitProducerTool
- How can you use these data?  @emanueleusai can you give a brief text and and eventually a link to an example code?

emanueleusai:
Hi @katilp ,
 
Here's the information needed:

> * Dataset name from @emanueleusai

Samples with high-momentum jets for tracking, ML, and top quark tagging studies

> * Authors are Emanuele Usai, Michael Andrews, Bjorn Burkle, Sergei Gleyzer, Meenakshi Narain, orcid?

Emanuele Usai
https://orcid.org/0000-0001-9323-2107

Michael Andrews
https://orcid.org/0000-0001-5537-4518

Bjorn Burkle
https://orcid.org/0000-0003-1645-822X

Sergei Gleyzer
https://orcid.org/0000-0002-6222-8102

Meenakshi Narain
https://orcid.org/0000-0002-7857-7403

> * @emanueleusai will provide a description (for "abstract")

The dataset consists of hits from the tracking detector, reconstructed tracks, simulated tracks, generated particles, and jets clustered from the generated particles. The various objects are matched in order to reconstruct the provenance of the various hits. 
[...]

Issue #2526, opened 2019-03-21
CMS: Prepare data records for tracker-hit-enriched Run1 AOD for ML


katilp:
prepare data records for the intermediate steps for tracker-hit-enriched Run1 AODs

(from #2444 )

Corresponding open datasets are

http://opendata.cern.ch/record/9582 (ttbar all jets)
http://opendata.cern.ch/record/8882 (QCD flat)
http://opendata.cern.ch/record/8898 (QCD EM Enriched)

The newly produced ones, tracker hits added, are being uploaded to
https://eospublichttp.cern.ch/eos/opendata/cms/upload/emanuele

and have the following configs (step0 SIM, step1 HLT, step2 RECO, step3 Demo):

record 9582 (ttbar all jets):
https://github.com/emanueleusai/opendatadnn/blob/master/step0_ttbar_V27.py
https://github.com/emanueleusai/opendatadnn/blob/master/step1_ttbar_V27.py
https://github.com/emanueleusai/opendatadnn/blob/master/step2_ttbar_V27.py
https://github.com/emanueleusai/opendatadnn/blob/master/step3_ttbar_V27.py

record 8882 (QCD flat):
https://github.com/emanueleusai/opendatadnn/blob/master/step0_QCDPt_15_3000_Flat_V27.py
https://github.com/emanueleusai/opendatadnn/blob/master/step1_QCDPt_15_3000_Flat_V27.py
https://github.com/emanueleusai/opendatadnn/blob/master/step2_QCDPt_15_3000_Flat_V27.py
https://github.com/emanueleusai/opendatadnn/blob/master/step3_QCDPt_15_3000_Flat_V27.py

record 8898 (QCD EM Enriched):
https://github.com/emanueleusai/opendatadnn/blob/master/step0_QCDPt_80_170_EM_V27.py
https://github.com/emanueleusai/opendatadnn/blob/master/step1_QCDPt_80_170_EM_V27.py
https://github.com/emanueleusai/opendatadnn/blob/master/step2_QCDPt_80_170_EM_V27.py
https://github.com/emanueleusai/opendatadnn/blob/master/step3_QCDPt_80_170_EM_V27.py

Release: CMSSW_5_3_32 Global Tag: START53_V27

In the dataset description of the record it should be mentioned that these datasets have been specially prepared in order to keep tracker hit that are needed for the ML studies.

To do:

- [x] prepare records
- [x] complete metadata (nevents)

katilp:
[...]

PR #2601, opened 2019-04-15
Add records for tracker-hit-enriched Run1 AOD and ML files, and SW

katilp:
(addresses #2526 & #2575 & #2588)

Adds 4 records for tracker-hit-enriched Run1 AOD, related  records 9582 (for the 1st record), 8882 (for the 3 others)

Adds records for the resulting ML files

Adds record for the SW

For AODSIM, needs:
- modified config files from the github repo (see also  #2588 ) either as records of their own or as files attached to their own record (now links are after FIXME in the config title field
- size of filese
- file indexes: files to be moved from the `/eospublic/cms/upload/emanuele`
   - `step2_ttbarOD_0*` for the 1st record
   - `step2_QCD300to600` for the 2nd record
   - `step2_QCD400to600` for the 3rd record
   - `step2_QCD_600to3000_0*` for the 4th record

For ML files, needs:
- remove end-of-lines in table html
- remove quotes if needed for type field in the dataset semantics

@emanueleusai Is that right, all files in szep2_ttbarOD_0* directories belong together? And the same for 600to3000?

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-derived-Run1-datascience.json
+++ b/cernopendata/modules/fixtures/data/records/cms-derived-Run1-datascience.json
@@ -1,0 +1,565 @@
+[
+  {
+    "abstract": {
+      "description": "<p>Samples in this record are in a custom root ntuple format and contain the position of the hits and information from the generator-level objects associated to the tracker hits. The samples can be used to study top quark identification algorithms that use low-level detector information such as tracker hits. Machine learning algorithms are suitable for this classification task.<\p><p> They have been produced from datasets, which consists of events extracted from simulated proton-proton collision events at a center-of-mass energy of 8 TeV generated with Pythia 6 (QCD) or MadGraph2.6 and Pythia6 (top-antitop pair sample). The particles emerging from the collisions traverse through a simulation of the CMS detector. The parent datasets of these samples contain light jets (QCD) in various energy ranges (FIXME add recids), or all-hadronic high transverse momentum decays of top quarks (FIXME recid), and consist of hits from the tracking detector, reconstructed tracks, simulated tracks, generated particles, and jets clustered from the generated particles. The various objects are matched in order to reconstruct the provenance of the various hits. Samples are produced from the standard CMS format \"AODSIM\" plus a series of low-level tracker-related collections that allow the extraction of the tracker hits. </p> <p><thead>
+<tr>
+<th align="left">Data set name</th>
+<th align="left">Description</th>
+<th align="right">Number of events</th>
+<th align="right">Number of files</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+<td align="left">QCD300to600</td>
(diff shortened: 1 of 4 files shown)


emanueleusai:
Hi @katilp ,

Yes, step2_ttbarOD_0* and step2_QCD_600to3000_0* belong together, respectively.
However, if you want to merge each of the datasets in a single folder, the files must be renamed, as the numbering re-starts from zero in each folder. 

Cheers,
Emanuele

katilp:
OK, thanks, I think they can stay in different folders, we have that for most regular datasets. The file index with the folder name will keep them separate i.e. for xrootd reading. However, I let @tiborsimko to decide. Would  it be a problem for the example code if they are renamed?

emanueleusai:
@katilp no, it's not a problem for me. I'll just have to change the file lists, but that's quick and easy to do. Plus the lists need to be re-done anyway once the files are moved from the /emanuele folder to the final location

katilp:
@emanueleusai OK, good to know. Note that the file indexes will generated and attached to the record (e.g. similar to http://opendata-dev.web.cern.ch/record/8882/files/CMS_MonteCarlo2012_Summer12_DR53X_QCD_Pt-15to3000_TuneZ2star_Flat_8TeV_pythia6_AODSIM_PU_RD1_START53_V7N-v1_00000_file_index.txt), so you can maybe wget them from there in the final instructions

emanueleusai:
@katilp ah that's good to know, thanks!

tiborsimko:
@emanueleusai @katilp One cosmetic comment: what about numbering the files with leading zero digits so that each file has exactly four digits (e.g. `_0004` instead of `_4`) so the alphabetical file sorting would be naturally nice? 

Currently one gets:

```console
$ eos ls /eos/opendata/cms/upload/emanuele/step2_QCD400to600 | head -15
step2_OpenData_1.root
step2_OpenData_10.root
step2_OpenData_100.root
step2_OpenData_1000.root
step2_OpenData_1001.root
step2_OpenData_1002.root
step2_OpenData_1003.root
step2_OpenData_1004.root
step2_OpenData_1005.root
step2_OpenData_1006.root
step2_OpenData_1007.root
step2_OpenData_1008.root
step2_OpenData_1009.root
step2_OpenData_101.root
step2_OpenData_1010.root
```

We could produce instead:

```console
step2_OpenData_0001.root
step2_OpenData_0002.root
step2_OpenData_0003.root
step2_OpenData_0004.root
step2_OpenData_0005.root
step2_OpenData_0006.root
step2_OpenData_0007.root
step2_OpenData_0008.root
step2_OpenData_0009.root
step2_OpenData_0010.root
step2_OpenData_0011.root
...
```

However, if this means some non-trivial changes in example code generating and using these files based on how many one produces, then it may be not worth it... I'll leave you to judge.

emanueleusai:
@tiborsimko sure I can do that

emanueleusai:
@tiborsimko I have renamed the files as suggested

Issue #2613, opened 2019-05-15
CMS: remaining fixes needed in provenance


katilp:
## Data:
- 2010:
  - [x] Trigger config listing and cmsdriver example #2450 
- 2011-2012 all good

## MC:
- 2010
  - [x] change the order of steps
  - [x] CMSSW version and GT missing for steps (e.g. http://opendata-dev.web.cern.ch/record/729)
  - [x] check if configs are there
    - e.g RECO and DIGI for https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=config+dataset%3D%2FQCD_Pt-15to1000_fwdJet_bwdJet_Tune4C_Flat_7TeV_pythia8%2FSummer12-LowPU2010_DR42_PU_S0_START42_V17B-v1%2FAODSIM
- 2011 looks OK
  - [x] amend "Production Overview" case 
  - [x] amend "Finding the parameters" phrase
- 2012 
  - [x] change the order of steps
  - [x] all(?) LHE files missing, the generator parameters link gives now hadronizer parameters (also nice to have..) but the information about the LHE step is not there (e.g. http://opendata-dev.web.cern.ch/record/7803: /GluGluTo2e2mu_Contin_8TeV-MCFM67-pythia6/Summer12_DR53X-PU_S10_START53_V19-v1/AODSIM)
  - [x] check if missing CMSSW version and GT can be extracted (e.g. http://opendata-dev.web.cern.ch/record/7724: /DYJets_0p0_1p2_2p10_3p15_4p15_CT10_8TeV-sherpa/Summer12_DR53X-PU_RD1_START53_V7N-v1/AODSIM)

- 2016
  - [x] what is now under RECO-HLT step is MiniAODSIM step 
  - [x] what is now under GEN-SIM is GEN-SIM and RECO
  - [x] config files not linked
  - [x] check with the UL legacy script
  ---> local cache creation OK with cms-YYYY script, including LHE for Hbb samples and MiniAODSIM step, files to be updated

tiborsimko:
WRT 2010 and 2012, all fixed including GT, except for CMSSW versions that will have to be retrieved anew from DAS, since McM is empty... I'll launch the necessary cache population jobs.

tiborsimko:
> I'll launch the necessary cache population jobs.

All finished last night so I'll merge 2012 changes too and deploy on DEV.

tiborsimko:
[...]

PR #2634, opened 2019-05-27
records: CMS 2010 MC provenance fixes

tiborsimko:
- Fixes provenance information for CMS 2010 simulated datasets. (addresses #2613)

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-simulated-datasets-2010.json
+++ b/cernopendata/modules/fixtures/data/records/cms-simulated-datasets-2010.json
@@ -1,7 +1,7 @@
 [
   {
     "abstract": {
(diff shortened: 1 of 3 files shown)


tiborsimko:
@katilp The CMS 2012 MC will follow shortly, the DAS gathering script will take many hours to finish... But you can already look at some 2010 records to see how the provenance looks like. Should be all ready there.

katilp:
I now realize (now it is easy to see as the generator/hardonizer parameters are nicely displayed) that the some 2010 MC (e..g. having madgraph in their name) have an LHE step, and it is available in DAS.
e.g. for /QCD_HT-250To500_TuneZ2Star_7TeV-madgraph-pythia6/Summer12-LowPU2010_DR42_BS2011_PU_S0_START42_V17B-v1/AODSIM record 735. Does the script reach grand parent in DAS f0r 2010 datasets? Or was it just record building which does not take them from cache (or maybe that record is not yet updated)

tiborsimko:
Yes, the script reaches to all parents in both McM and DAS, so e.g. the record 735 has well the LHE provenance step recognised.

Issue #2426, opened 2018-08-08
CMS: new VM for 2011-2012 with slc6 container


katilp:
A new version of VM for 2011-2012 is available for testing.
It has the CMS shell as slc6 shell as a singularity container (similar to VM 1.1.2 in http://opendata.cern.ch/record/250)

UPDATE, see below https://github.com/cernopendata/opendata.cern.ch/issues/2426#issuecomment-447997789

https://cernbox.cern.ch/index.php/s/omF4JNCPMDigQk8

It corresponds to the image

http://cernvm.cern.ch/releases/production/cernvm4-micro-2018.06-2.hdd

with the context from

https://github.com/cernvm/public-contexts/blob/master/cms-opendata-2011.context

Testing needed.

katilp:
@jblomer Were there some updates needed to this VM (image and/or contextualization) for the eos mount issue?

jblomer:
@katilp Yes, I'm still working on it.  The `df` issue is sorted out, eos still needs an update.

katilp:
@jblomer Thanks! It seems many of the testers do not open the slc6 shell, would it be possible to have the OS visible in the prompt for both shells?

katilp:
Update from @jblomer 

> a new version of the VM (CMS OpenData VM 1.5.0):
> 
>     https://cernbox.cern.ch/index.php/s/YJeRKF5hIXLpnIy
> 
> from the sources
> 
> 
> http://cernvm.cern.ch/releases/ucernvm-images.2018.10-2.cernvm.x86_64/ucernvm-v4testing.2018.10-2.cernvm.x86_64.hdd
> 
> and
> 
> https://github.com/cernvm/public-contexts/blob/master/cms-opendata-2011.context
> 
> This VM has the following changes/fixes:
> 
>   o) Update to and EL 7.6 based OS
> 
>   o) df in the CMS shell works
> 
>   o) The standard shell now has a prompt "[Outer Shell]".
> 
> For in-house use, EOS volume can be mounted:
> 
> sudo service eosd start
> kinit yourcernlogin@CERN.CH # <-- your CERN login
> eosfusebind -g

katilp:
Testing:

- [x] https://github.com/cms-opendata-analyses/trigger_examples/tree/2011/TriggerInfo/TriggerInfoAnalyzer (/record/5004) on CMSSW_5_3_32 reading GT FT_53_LV5_AN1_RUNA 
[...]

PR #2645, opened 2019-06-03
records: new CMS VM image 1.5.1

tiborsimko:
* Adds new CMS VM image 1.5.1 OVA file. (addresses #2426)

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-tools-vm-image-2012.json
+++ b/cernopendata/modules/fixtures/data/records/cms-tools-vm-image-2012.json
@@ -56,7 +56,7 @@
       "Run2012C",
       "Run2012D"
     ],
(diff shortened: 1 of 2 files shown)


tiborsimko:
@katilp I see there is a 1.5.1 context file linked in the issue, which will become `cms-user-data.txt` in the contextualisation record.

What about `cernvm-script`? Is it sufficient to replace HDD version or have some CLI parameters related to disk and memory also changed?

What about `opendata-desktop-settings`?  I guess no change is necessary there...

katilp:
We'll need to check with @jblomer for the contextualisation details:  Jakob, for reference, we used to have the scripts to produce the VM in a separate record, such as http://opendata-dev.cern.ch/record/253, are there changes in `cernvm-script` and `opendata-desktop-settings` between 1.3.0 and 1.5.1? Thanks!

Issue #2648, opened 2019-06-04
CMS: check list for the release (2010 records)


katilp:
# Collision data:

2010 collision data records:

- [x] build records for the 2010 HLT config files (now FIXMEs in http://opendata-dev.web.cern.ch/record/1699) see #2657 and https://github.com/cernopendata/data-curation/pull/48
- [x] update description of good lumi http://opendata-dev.web.cern.ch/record/1000, add mention to RunA (or remove public data = RunB), was not addressed in #2637 (check the run numbers for RunA, see also https://twiki.cern.ch/twiki/bin/view/CMS/PhysicsPrimaryDatasets)
   - min max run numbers from DAS (NB run numbers  not necessarily in order in DAS...):
    - BTau 141950 144114
    - EGMonitor 135821 144114
    - ZeroBias  135808 144431
    - MuOnia 140042 144114
    - JetMET 141950 144114
    - JetMETTau 135821 141887
    - EG 135821 144114
    - MuMonitor 136097 136782
    - MinimumBias 135808 144114
    - Commissioning 135821 144114
    - Mu 135821 144114
    - JetMETTauMonitor 135821 144114
    - ----> RunA is between 135808 and 144431 (-> #2651)
- [x] Add the selection description to RunA ZeroBias http://opendata-dev.web.cern.ch/record/74
- [x] Add the mention of JetMETTau split into  JetMET and BTau in the RunA datasets (see https://hypernews.cern.ch/HyperNews/CMS/get/online-selection/1034.html) 
- [x] All data records: Add a link to the cmsDriver description (coming in #2630) from the provenance description: "can be generated with the cmsdriver command line script" 
- [x] All data records: check if the cmsdriver script line in the data reocrds can be displayed better
- [x] All data records: fix the usage to point to 2010 VM and getting started 
   - /docs/cms-virtual-machine-2010
   - /docs/cms-getting-started-2010
- [x] create eos/opendata/cms/Run2010A and move RunA datasets from the upload area, add file index  (see #2451 )

2010 MC records:
- [x] config files to be added and linked to the records (under record/NNN/files ?) see #2642
- [x] check how many of the 100 dataset have no provenance (production script, generator parameters)  from mcm (as in QCD datasets like http://opendata-dev.web.cern.ch/record/755 in comparison to http://opendata-dev.web.cern.ch/record/798), did it get fixed/does it matter?
[...]

PR #2652, opened 2019-06-04
records: updates to 2010 data records

katilp:
(addresses #2648)

Updates:
- adds the selection description to RunA ZeroBias 
- adds a mention of JetMETTau split into JetMET and BTau in the RunA datasets
- adds a link to the cmsDriver description
- fixes the usage to point to 2010 VM and getting started


cernopendata/modules/fixtures/data/records/cms-primary-datasets.json
```diff
@@ -2598,6 +2598,9 @@
...
+    },
     "methodology": {
       "description": "<p> FIXME"
```
tiborsimko: I'll remove this extra `methodology` field containing FIXME while merging...

Issue #2655, opened 2019-06-05
CMS: Final check list for the release (ML samples)


katilp:
SW records:

- [x] add figure #2654 

ML files:

- [x] point to the usage examples

MC files:

- [x] Phase2 all: `date_created`: 2019 -> Phase2 
- [x] Phase2 ttbar GEN-SIM-DIGI-RAW http://opendata-dev.web.cern.ch/record/12301
   - remove ")" in "as the input in ML studies)." 
   - "in GEN-SIM-DIGI-RAW (see CMS Monte Carlo production overview) format" -> "in GEN-SIM-DIGI-RAW format (see CMS Monte Carlo production overview)  
  - starts with "This dataset" -> change to singular for the rest of the description
- [x] Phase2 ttbar GEN-SIM
   - "in GEN-SIM (see CMS Monte Carlo production overview) format" -> "in GEN-SIM format (see CMS Monte Carlo production overview) 
  -  starts with "This dataset" -> change to singular for the rest of the description
  - "These data were generated in several steps" -> "This is the first step in the production chain"
- [x] Phase2 ttbar FEVTHLTDEBUG
  - remove ")" in "as the input in ML studies)."
  - starts with "This dataset" -> change to singular for the rest of the description
- [x] Phase2 FEVTHLTDEBUG and GEN-SIM-DIGI-RAW fix system details (likely comes from the template as it goes right for GEN-SIM format but not for these two:
![systdetails](https://user-images.githubusercontent.com/5859918/58978601-722ada80-87cc-11e9-84d1-5859e35cfcef.JPG) (#2660)

Run2 MiniAODSIM 
  - [x] correct plural/singular in the description
  - [ ] check why LHE step is not there for `BulkGrav...` (@tiborsimko  maybe it was only tested locally, and not merged?)
    - LHE step is not in DAS but available in McM e.g. for /BulkGravTohhTohbbhbb_narrow_M-1000_13TeV-madgraph/RunIISummer15wmLHEGS-MCRUN2_71_V1_ext1-v1/GEN-SIM https://cms-pdmv.cern.ch/mcm/requests?prepid=B2G-RunIISummer15wmLHEGS-00781&page=0&shown=70368744177663
    - followed up in #2525
 - [x] mention of pile-up is missing for all MINIAODSIM 

Run1 tracker-hit-enriched datasets (cms-simulated-datasets-Run1-datascience.json)
- [x] correct plural/singular in the description
[...]

PR #2656, opened 2019-06-05
records: final fixes to ML records

### LINKED ISSUE #2654: CMS: figure for the SW record for ML sample production (tracker-hit-enriched Run1 AOD)
Figure to add to SW record /record/12210 to explain different steps

opendata_cartoon

See if the description reads well with it

katilp:
(addresses #2655)
(closes #2654)

Fixes to Phase2 MC records:
- date_created Phase2 (2019 makes no sense to indicate the year data taking)
- text improvements in the description

Fixes to Run1 SW record:
- add workflow figure
- add links to usage examples in cernopendata-datascience repositories

Fixes to Run1 MC records
- abstract text
- add url to configs (fixes legacydata -> opendata, and master -> v12.0.0 in config links)

Fixes to Run2 MiniAODSIM
- abstract text (plural/singular)
- add mention of pileup events

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-simulated-datasets-Phase2-datascience.json
+++ b/cernopendata/modules/fixtures/data/records/cms-simulated-datasets-Phase2-datascience.json
@@ -1,7 +1,7 @@
 [
   {
     "abstract": {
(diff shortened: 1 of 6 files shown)


cernopendata/modules/fixtures/data/records/cms-simulated-datasets-Phase2-datascience.json
```diff
@@ -22,7 +22,7 @@
       "type": "pp"
     },
     "date_created": [
-      "2019"
+      "Phase2"
```
tiborsimko: We are using `date_created` to mean the collision year, roughly... However, this field is also going to schema.org `dateCreated` field which needs to be ISO 8601 compliant... If we allow "Phase2" value to signify future, what YYYY value shall we use for the schema.org export? E.g. current year could be a fallback...

katilp: Phase2, according to the current LHC schedule, is 2026. We could use that...

katilp:
Added links to cernopendata-datascience repositories for three ML files records in this PR (not to the record from Adriano that was just updated on master)

PR #2769, opened 2019-07-18
Remove issuestats badges from README

RaoOfPhysics:
- The service no longer works
- The domain is no longer controlled by the original developers

Signed-off-by: Achintya Rao <achintya.rao@cern.ch>


RaoOfPhysics:
Will update the branch.

Issue #2777, opened 2019-07-29
CMS: Higgs boson to two tau leptons outreach analysis


stwunsch:
Hi all,

I'm currently implementing the records describing the outreach analysis in [this repository](https://github.com/cms-opendata-analyses/HiggsTauTauNanoAODOutreachAnalysis) and I'll use this issue to track the progress. See the plot below for an example result:

![](https://github.com/cms-opendata-analyses/HiggsTauTauNanoAODOutreachAnalysis/raw/master/plots/m_vis.png)

**First question:** The analysis uses 9 derived datasets in reduced NanoAOD format. Should I make a separate record for each of these or one for all of them? If I do a separate record for each derived dataset, I would have to copy the dataset semantics many times. Is it ok to put all of them into a single record?

Best
Stefan

katilp:
Hi Stefan, if the format is generic (i.e. usable also with other examples), I would be in favour of having them in separate records. It is then easier to describe the dataset provenance (i.e. derived from NNN dataset etc). 

Are these dataset the output of recid 12340 or the output of skim.cxx which is in https://github.com/cms-opendata-analyses/HiggsTauTauNanoAODOutreachAnalysis?

The skims (specific to this analysis) could eventually be in one record, but the "NanoAOD"s (outputs of the producer) would preferably have their own records.

Does that make sense?

stwunsch:
Alright, thanks! So the files are not the output of the skims, I do not plan to put them on the portal (?).

So mainly my problem is the following: Shall I copy the dataset semantics (about 40 variables) for each record in each json file? This is not maintainable on long term since you have to change things in 9 different places.

katilp:
If they are on the portal they will have a certain stability. We do not expect them to change that often, and if the contents change, there will be a new set of records.
Independent of that, pinging @tiborsimko to see if there's a possibility to have the dataset semantics for a set of files to be defined in a single source location. That would be the best in any case. Thinking forward also to the full NanoAOD: we will certainly not repeat full variable description for all records.

stwunsch:
@tiborsimko I could do a "description record" and link the record from all other records. Otherwise I don't see a solution (or I copy everything around).

stwunsch:
I've put an example setup in the PR #2778.

katilp:
Chatting with @tiborsimko yesterday: we could have the semantics description in the sister `data-curation` repository, which is used for records (most data records) that are generated by a script. And then have a simple script to generate the records.

[...]

PR #2778, opened 2019-07-30
CMS: Higgs boson to two tau leptons outreach analysis

stwunsch:
PR to include the outreach analysis of a Higgs boson decaying into two tau leptons

The issue for discussion is #2777.

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-derived-nanoaod-outreach-higgstautau.json
+++ b/cernopendata/modules/fixtures/data/records/cms-derived-nanoaod-outreach-higgstautau.json
@@ -1,0 +1,2756 @@
+[
+  {
+    "abstract": {
+      "description": "<p>Dataset in reduced NanoAOD format for education and outreach derived from the primary dataset in AOD format linked below.</p>"
+    },
+    "accelerator": "CERN-LHC",
+    "authors": [
+      {
+        "name": "Wunsch, Stefan"
+      }
+    ],
+    "collections": [
+      "CMS-Derived-Datasets"
+    ],
+    "dataset_semantics": [
+      {
+        "description": "Run number",
+        "type": "int",
+        "variable": "run"
+      },
+      {
+        "description": "Luminosity block number",
+        "type": "unsigned int",
+        "variable": "luminosityBlock"
+      },
+      {
+        "description": "Event number",
+        "type": "unsigned long",
+        "variable": "event"
+      },
+      {
+        "description": "Flag indicating whether the event passed the respective trigger",
+        "type": "bool",
+        "variable": "HLT_IsoMu24"
+      },
+      {
+        "description": "Flag indicating whether the event passed the respective trigger",
+        "type": "bool",
+        "variable": "HLT_IsoMu24_eta2p1"
+      },
+      {
+        "description": "Flag indicating whether the event passed the respective trigger",
+        "type": "bool",
+        "variable": "HLT_IsoMu17_eta2p1_LooseIsoPFTau20"
+      },
+      {
+        "description": "Number of primary vertices",
+        "type": "int",
+        "variable": "PV_npvs"
+      },
+      {
+        "description": "Number of muons in this event",
+        "type": "unsigned int",
+        "variable": "nMuon"
+      },
+      {
+        "description": "Transverse momentum of the muons",
+        "type": "float[nMuon]",
+        "variable": "Muon_pt"
+      },
+      {
+        "description": "Pseudorapidity of the muons",
+        "type": "float[nMuon]",
+        "variable": "Muon_eta"
+      },
+      {
+        "description": "Azimuth of the muons",
+        "type": "float[nMuon]",
+        "variable": "Muon_phi"
+      },
+      {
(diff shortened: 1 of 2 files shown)


cernopendata/modules/fixtures/data/records/cms-derived-nanoaod-outreach-higgstautau.json
```diff
@@ -0,0 +1,2693 @@
...
+        },
+        "validation": {
+            "description": "<p>These data were derived from the primary datasets linked below using only the validated runs. No further validation was done for the output.</p>",
+            "links": [
+                {
+                    "recid": "8001"
+                }
+            ]
+        }
+    },
```
katilp: It would be good to have:

```
"relations": [
      {
        "description": "This dataset was derived from:", 
        "recid": "8001", 
        "title": "/GluGluToHToTauTau_M-125_8TeV-powheg-pythia6-tauPolarOff/Summer12_DR53X-PU_S10_START53_V19-v1/AODSIM", 
        "type": "isChildOf"
      }
```
to indicate the direct relation. Now it is visible through "validation", but it would be better to have explicitly it (also for any eventual future work on the portal content display relying on parent-child relation of dataset)

stwunsch: Thanks for the review and the comments! Good point here, I've added the respective section in each record.

stwunsch:
@katilp Thanks for your review! I've added the relation to the primary dataset and changed the path as you suggested (makes totally sense!).

stwunsch:
@tiborsimko Is it possible to have this first version on the dev server for further reviews?

stwunsch:
@tiborsimko @katilp Could we proceed with merging this PR on the dev server?

cernopendata/modules/fixtures/data/records/cms-tools-nanoaod-outreach-higgstautau.json
```diff
@@ -0,0 +1,107 @@
...
+    },
+    "accelerator": "CERN-LHC",
+    "authors": [
+      {
+        "name": "Wunsch, Stefan"
+      }
+    ],
+    "collections": [
+      "CMS-Tools"
```
tiborsimko: "CMS-Derived-Datasets" is the value to use for datasets... I can change during merge.

cernopendata/modules/fixtures/data/records/cms-derived-nanoaod-outreach-higgstautau.json
```diff
@@ -0,0 +1,2756 @@
...
+            "number_events": 476963,
+            "number_files": 1,
+            "size": 196517149
+        },
+        "experiment": "CMS",
+        "files": [
+            {
+                "checksum": "adler32:2b025925",
+                "size": 196517149,
+                "uri": "root://eospublic.cern.ch//eos/opendata/cms/derived-data/AOD2NanoAODOutreachTool/GluGluToHToTauTau.root"
```
tiborsimko: Currently we have:

```console
$ eos ls /eos/opendata/cms/derived-data/AOD2NanoAODOutreachTool/
Run2012BC_DoubleMuParked_Muons.root
```

After the PR, we would have more files in the same directory:

```
DYJetsToLL.root
GluGluToHToTauTau.root
Run2012B_TauPlusX.root
Run2012BC_DoubleMuParked_Muons.root
Run2012C_TauPlusX.root
TTbar.root
VBF_HToTauTau.root
W1JetsToLNu.root
W2JetsToLNu.root
W3JetsToLNu.root
```

Are the new data also derived using the same `AOD2NanoAODOutreachTool`?  If so, how would people know how they can generate the former and how the latter data? (E.g. parameters to use, commands to call, etc.)  If the same tool can be used to produce many different derived datasets for different purposes, then it would make sense invent some scalable storage schema, e.g. to have subdirectories there for each use case or some such?

katilp: This is indeed the same tool, for the previous files some parts were commented in the code to produce only muons, here instead, the full output is kept.

tiborsimko: OK, I'll move them to the same place, but we should think how to capture the tool parametrisation leading to the various outputs. Could be simply via tags and using something like:

- `AOD2NanoAODOutreachTool/v1.0/Run2012BC_DoubleMuParked_Muons.root` 
- `AOD2NanoAODOutreachTool/v1.1/Run2012C_TauPlusX.root`
 
and the information about "provenance" will be found within the given tool's tag.

Or, if the same version of the tool is used, one could attach "provenance" configuration files and/or shell scripts next to each file:

- `AOD2NanoAODOutreachTool/Run2012C_TauPlusX.root` 
- `AOD2NanoAODOutreachTool/Run2012C_TauPlusX_generation/run.sh` 
- `AOD2NanoAODOutreachTool/Run2012C_TauPlusX_generation/config.h` 


that the given version of the tool uses. We can muse later about these...

Issue #2785, opened 2019-08-21
facets: Add 'event number' -facet


JuhaTeuho:
In preparation a new search facet for the Open Data Portal. This facet is called "event number" and it searches the records for a range of events in the record file. In the current jsonschema there is a field called 'distribution.number_events' which will be used to determine the number of events in the record. Facet would look like this:

![Screenshot from 2019-08-21 12-17-31](https://user-images.githubusercontent.com/51369786/63426287-bd8b7700-c412-11e9-906b-4af2a51edf50.png)

This facet requires changes to the following files:

- cernopendata/config.py    addition of the facet itself and the filter
- cernopendata/templates/cernopendata/search.html    addition of the facet
- Possibly some other changes for filter implementation

tiborsimko:
Closed by #2787.

PR #2787, opened 2019-08-21
facets: Add event_number facet

JuhaTeuho:
- Adds 'event_number'-facet to Open Data Portal.
- NOTE! Implementation for the range_filter is still required. At the moment, filter is missing and therefore the facet doesn't work correctly.
(addresses #2785)

Signed-off-by: Juha-Matti Teuho <juha.teuho@gmail.com>

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -315,6 +315,38 @@ RECORDS_REST_FACETS = {
             signature=dict(terms=dict(
                 field='signature.keyword',
                 order=dict(_term='asc'))),
+            event_number= {
+                'range': {
+                    'field': 'distribution.number_events',
+                    'ranges': [
+                        {
+                            'key' : '0 - 999',
+                            'from': 0,
+                            'to': 999
+                        },
+                        {
+                            'key' : '1000 - 9999',
+                            'from': 1000,
+                            'to': 9999
+                        },
+                        {
+                            'key' : '10 000 - 99 999',
+                            'from': 10000,
+                            'to': 99999
+                        },
+                        {
+                            'key' : '100 000 - 999 999',
+                            'from': 100000,
+                            'to': 999999
+                        },
+                        {
+                            'key' : '1 000 000 -',
+                            'from': 1000000,
+                            'to': 100000000
+                        }
+                    ]
+                }
+            }
         ),
         'post_filters': dict(
             experiment=terms_filter('experiment.keyword'),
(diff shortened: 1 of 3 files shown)


cernopendata/config.py
```diff
@@ -312,6 +312,38 @@
...
+                            'from': 10000,
+                            'to': 99999
+                        },
+                        {
+                            'key' : '100 000 - 999 999',
+                            'from': 100000,
+                            'to': 999999
+                        },
+                        {
+                            'key' : '1 000 000 -',
```
diegodelemos: I think we might be able to reuse the existing `range_filter`. Can you try to use double dashes for the ranges (`--`) as the default [`range_filter`](https://github.com/inveniosoftware/invenio-records-rest/blob/b8db18e3c0a0943d99050c224ed858888d8a7595/invenio_records_rest/facets.py#L38) [expects the two dashes](https://github.com/inveniosoftware/invenio-records-rest/blob/b8db18e3c0a0943d99050c224ed858888d8a7595/invenio_records_rest/facets.py#L52). Also, are you sure the numbers should have spaces in between?

cernopendata/config.py
```diff
@@ -328,6 +360,7 @@
             file_type=terms_filter('distribution.formats.keyword'),
             collections=terms_filter('collections.keyword'),
             availability=terms_filter('distribution.availability.keyword'),
+            event_number= None,
```
diegodelemos: So with the above change, we should just import the default `range_filter` and use it here:

```python
from invenio_records_rest.facets import range_filter
...
            event_number=reange_filter('distribution.number_events') 
```
Can you try that and share the results? Thank you :)

JuhaTeuho:
Thanks for the suggestions, the filter is working now! I added a new commit, where the filter is working. The problem was indeed the additional spaces and missing dashes. However, this changes the overall look of the filter

Screenshot from 2019-08-28 13-48-15

It would be nice to have those titles to be of the form "1 000 000 - 9 999 999" so it would be easier to tell how many zeros there are. I guess this could be done by implementing a new range_filter but perhaps it is not worth the time just for this aesthetic change.

Another thing that came up was that the filter name "event number" might be mixed up with the dataset parameter "event number" so maybe a title called "number of events" would be better. However, if I try to change the name, only the first underscore is replaced with a whitespace:

Screenshot from 2019-08-28 13-25-17

I tried to solve this by changing line 40 in [facets.html](https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/static/templates/cernopendata_search_ui/facets.html) from
`<div class="title"> Filter by {{ aggr.key.replace('_', ' ') }} </div>`
to
`<div class="title"> Filter by {{ aggr.key.replace(/_/g, ' ') }} </div>`
but it had no effect at all.

Maybe there is a way to change that? Alternatively we can keep the name "event number" as it is.

PR #2829, opened 2020-02-28
records: CMS dimuon spectrum with reduced NanoAOD

stwunsch:
* Update to release v1.3 of the software
* Use tags instead of branches for the links


tiborsimko:
Thanks, rebased against QA, I'll take care of amending files in the same PR.

tiborsimko:
@stwunsch Deployed in production as well.

PR #2885, opened 2020-09-14
records: Update cms-derived-Run2-datascience

jmduarte:
- Small update to add descriptions of labels
- Fix to one of the data types


tiborsimko (approved):
Thanks, fixed trailing white space and merging...

PR #2895, opened 2020-10-08
records: add remaining orcid ids

### LINKED ISSUE #2888: orcid: add orcid ids globally
ORCID iDs can be added retrospectively for all records that have individuals as authors. Currently they are only present in a subset or records. 

We should first establish: 
- how many records there are with individual authors
- how many of those records have an ORCID iD

ArtemisLav:
* (closes #2888)

Signed-off-by: Artemis Lavasa artemis.lavasa@cern.ch


tiborsimko (approved):
Great, thanks :rocket:

PR #2907, opened 2020-11-17
docker: fix yum clean instructions

tiborsimko:


Dockerfile
```diff
@@ -113,4 +110,5 @@ RUN adduser --uid 1000 invenio --gid 0 && \
 USER 1000
 
 # Start the CERN Open Data Portal application
-CMD ["uwsgi --module ${UWSGI_WSGI_MODULE} --socket 0.0.0.0:${UWSGI_PORT} --master --processes ${UWSGI_PROCESSES} --threads ${UWSGI_THREADS} --stats /tmp/stats.socket"]
+# hadolint ignore=DL3025
+CMD uwsgi --module ${UWSGI_WSGI_MODULE} --socket 0.0.0.0:${UWSGI_PORT} --master --processes ${UWSGI_PROCESSES} --threads ${UWSGI_THREADS} --stats /tmp/stats.socket
```
audrium, line 114: Was this change causing any errors? I actually did similar change to some `reana` repos and now wondering if it will break anything

Dockerfile
```diff
@@ -48,20 +48,17 @@ RUN yum install -y \
         openssl-devel \
         jq \
         python-devel \
-        python-pip \
-        yum clean all
+        python-pip && \
+    yum clean all
 
 # Install latest stable xrootd 4.12.5 version and its dependencies
 # hadolint ignore=DL3033
```
audrium, line 55: This `hadolint ignore` rule seems to be not needed any longer since all the versions are pinned

Dockerfile
tiborsimko: Yes, there was a problem. The image was building, but the site was not working. It might be related to expansion of `${...}` variables. I have tried also `CMD ["uwsgi", "--module", ..."` which is the usual way, but to no avail.  So reverted back due to lack of time.

Dockerfile
tiborsimko: Updated. I have actually joined it now with the previous `yum install` command, so that we don't have to do unnecessarily `yum clean all` twice.

audrium (approved):

PR #2918, opened 2020-11-24
Run2011A pp HI ref datasets

### LINKED ISSUE #2916: CMS - build data records for 2011 pp HI reference data
Build records for 2011 pp HI reference data cms-primary-datasets cms-primary-datasets-HIpp-Run2011A.json
- trigger paths available (see https://github.com/cernopendata/data-curation/issues/82)
- no reco configs available in DAS (see https://github.com/cernopendata/data-curation/issues/78)

### pp 2.76 TeV:

/AllPhysics2760/Run2011A-16Jul2011-v1/RECO 
/AllPhysics2760/Nov2011_HI-SD_JetHI-276TeV_ppRereco/RECO 
/AllPhysics2760/Nov2011_HI-SD_MuHI-276TeV_ppRereco/RECO 
/AllPhysics2760/Nov2011_HI-SD_PhotonHI-276TeV_ppRereco/RECO

### pp 7 TeV

/MinBias0Tesla0/Run2011A-PromptReco-v5/RECO 
/MinBias0Tesla1/Run2011A-PromptReco-v5/RECO 
/MinBias0Tesla2/Run2011A-PromptReco-v5/RECO 

### interfill (noise studies)

/ForwardTriggers/HIRun2011-PromptReco-v1/RECO (this should go to cms-primary-datasets-HIRun2011.json)

katilp:
(close #2916)

Adds 6 pp records from Run2011A, reference data for HI analysis

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-primary-datasets-HIpp-Run2011A.json
+++ b/cernopendata/modules/fixtures/data/records/cms-primary-datasets-HIpp-Run2011A.json
@@ -1,0 +1,737 @@
+[
+  {
+    "abstract": {
+      "description": "<p>AllPhysics2760 primary dataset from the 2.76 TeV proton-proton run of 2011, reference data for heavy-ion data analysis</p> <p> The list of validated runs, which must be applied to all analyses, either with the full validation or for an analysis requiring only muons, can be found in</p>",
+      "links": [
+        {
+          "description": "Validated runs, full validation",
+          "recid": "14208"
+        },
+        {
+          "description": "Validated runs, muons only",
+          "recid": "14209"
+        }
+      ]
+    },
+    "accelerator": "CERN-LHC",
+    "collaboration": {
+      "name": "CMS collaboration",
+      "recid": "450"
+    },
+    "collections": [
+      "CMS-Primary-Datasets"
+    ],
+    "collision_information": {
+      "energy": "2.76TeV",
+      "type": "pp"
+    },
+    "date_created": [
+      "2011"
+    ],
+    "date_published": "2020",
+    "date_reprocessed": "2011",
+    "distribution": {
+      "formats": [
+        "root",
+        "reco"
+      ],
+      "number_events": 56168211,
+      "number_files": 1491,
+      "size": 6570934150548
+    },
+    "experiment": "CMS",
+    "license": {
+      "attribution": "CC0"
+    },
+    "methodology": {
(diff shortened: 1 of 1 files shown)


cernopendata/modules/fixtures/data/records/cms-primary-datasets-HIpp-Run2011A.json
```diff
@@ -0,0 +1,737 @@
...
+    ],
+    "run_period": [
+      "Run2011A"
+    ],
+    "system_details": {
+      "global_tag": "GR_R_44_V15::All",
+      "release": "CMSSW_4_4_7"
+    },
+    "title": "/AllPhysics2760/Run2011A-16Jul2011-v1/RECO",
+    "title_additional": "AllPhysics2760 primary dataset in RECO format from the 2.76 TeV p-p run of 2011 (/AllPhysics2760/Run2011A-16Jul2011-v1/RECO)",
```
tiborsimko, line 65: Perhaps we can use "proton-proton" instead of "p-p"? 

But the most important question is what are we going to do with the data files? I don't think we have them on EOSPUBLIC already? Shall we merge with an "on demand" flag?

katilp: Proton-proton is better indeed 👍 .
According to my book-keeping (in https://github.com/cernopendata/data-curation/issues/73), the datasets were transferred. The only two missing are the RECO versions of already existing 2011 MinimumBias datasets, but I did not build the records for them.

tiborsimko: Ah, they are living under `data`, while I had looked under `hidata`. Where do we place them? E.g. `Nov2011_HI` would create a new top-level directory in `/eos/opendata/cms`...

katilp: Ah, OK, that makes sense as they are pp data and they ended up there because of their naming in CMS system. They are part of Run2011A so they can go under there as the other Run2011A datasets.

tiborsimko: Run2011A looks good, but Nov2011_HI is living outside, e.g.:

```console
$ eos ls /eos/opendata/cms/upload/data/Nov2011_HI
AllPhysics2760
$ eos ls /eos/opendata/cms/upload/data/Run2011A
AllPhysics2760
MinBias0Tesla0
MinBias0Tesla1
MinBias0Tesla2
MuEG
MuOnia
```

Hence probably the need for a new top-level directory Nov2011_HI...

katilp: OK, that makes sense as well. The three datasets which appear under ther three different directories under RECO there are different skims from the full dataset which resides under Run2011A. Let's keep this same logic if it does not hurt.

tiborsimko: I've started the data transfer and update this after merge...

PR #2968, opened 2020-12-16
installation: upgrade invenio packages

### LINKED ISSUE #2922: backend: migrate to latest Invenio 3 stable
CERN Open Data currently uses older versions of Invenio packages:

```console
$ grep invenio requirements-production.txt
invenio-assets==1.0.0b7
invenio-base==1.0.0a16
invenio-celery==1.0.0b3
invenio-config==1.0.0b3
invenio-db==1.0.0b8
invenio-i18n==1.0.0b4
invenio-indexer==1.0.0a10
invenio-jsonschemas==1.0.0a5
invenio-pidstore==1.0.0b2
invenio-previewer==1.0.0a11
invenio-query-parser==0.6.0
invenio-records==1.0.0b2
invenio-records-files==1.0.0a10
invenio-records-rest==1.0.0b1
invenio-records-ui==1.0.0b1
invenio-rest==1.0.0b2
invenio-search==1.0.0a11
invenio-search-ui==1.0.0a9
invenio-theme==1.0.0b2
```

Investigate migrating to latest Invenio stable releases, e.g. `invenio-records` from `1.0.0b2` to `1.3.2`.

The aim is to migrate the backend first, reusing JSON Schema model and Elasticsearch mappings, etc.

The goal is to load all records and files OK into an Invenio 3.3 or 3.4 instance.

The front-end migration and any necessary display adaptations will follow as part of another issue.

diegodelemos:
Closes #2922.


cernopendata/config.py
```diff
@@ -57,6 +57,12 @@
         "LOGGING_SENTRY_CELERY", False
     )
 
+# Security
+# ========
+#: Flask-Talisman secure headers
+APP_ENABLE_SECURE_HEADERS = False
```
diegodelemos, line 63: Disabling HTTPS temporarily. Flask-Talisman would force HTTPS on the clients but we don't have SSL setup in docker-compose production, so installation would break now.

docker-compose-dev.yml
```diff
@@ -41,24 +45,25 @@ services:
...
-      - APP_PIDSTORE_DATACITE_USERNAME=CERN.OPENDATA
-      - APP_PIDSTORE_DATACITE_PASSWORD=CHANGE_ME
-      - APP_PIDSTORE_LANDING_BASE_URL=http://opendata.cern.ch/record/
+      - INVENIO_SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://cernopendata:dbpass123@postgresql:5432/cernopendata
+      - INVENIO_CACHE_REDIS_HOST=redis
+      - INVENIO_CACHE_REDIS_URL=redis://redis:6379/0
+      - INVENIO_ACCOUNTS_SESSION_REDIS_URL=redis://redis:6379/1
+      - INVENIO_BROKER_URL=amqp://guest:guest@rabbitmq:5672/ # Celery 3
+      - INVENIO_CELERY_BROKER_URL=amqp://guest:guest@rabbitmq:5672/ # Celery 4
+      - INVENIO_CELERY_RESULT_BACKEND=redis://redis:6379/2
```
diegodelemos, line 60: Getting warning messages about upgrade path to Celery 6.X.Y:
```console
worker_1         | /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/celery/app/utils.py:206: CDeprecationWarning: 
worker_1         |     The 'CELERY_RESULT_BACKEND' setting is deprecated and scheduled for removal in
worker_1         |     version 6.0.0. Use the result_backend instead
worker_1         | /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/celery/app/utils.py:206: CDeprecationWarning: 
worker_1         |     The 'CELERY_TASK_SERIALIZER' setting is deprecated and scheduled for removal in
worker_1         |     version 6.0.0. Use the task_serializer instead
worker_1         | /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/celery/app/utils.py:206: CDeprecationWarning: 
worker_1         |     The 'CELERY_ACCEPT_CONTENT' setting is deprecated and scheduled for removal in
worker_1         |     version 6.0.0. Use the accept_content instead
worker_1         | /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/celery/app/utils.py:206: CDeprecationWarning: 
worker_1         |     The 'BROKER_URL' setting is deprecated and scheduled for removal in
worker_1         |     version 6.0.0. Use the broker_url instead
```

setup.py
```diff
@@ -67,41 +67,43 @@
...
+    'invenio-records-ui==1.0.0b1',
+    'invenio-records==1.4.0a3',
+    'invenio-search-ui==1.0.0a2',
+    # OD adapted Invenio `files` bundle
+    'invenio-previewer==1.0.0a11',
+    'invenio-records-files==1.2.1',
+    # OD adapted Invenio `postgresql` bundle
+    'invenio-db[versioning,postgresql]==1.0.5',
+    # OD adapted Invenio `elasticsearch5` bundle
+    'invenio-search[elasticsearch5]==1.4.0',
```
diegodelemos, line 93: Deprecation warnings about Elasticsearch5:
```console
web_1            | /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/invenio_search/ext.py:59: DeprecationWarning: Elasticsearch v2 and v5 support will be removed.
web_1            |   DeprecationWarning)
```

cernopendata/config.py
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of CERN Open Data Portal.
-# Copyright (C) 2017, 2018 CERN.
+# Copyright (C) 2017, 2018, 2020 CERN.
```
diegodelemos, line 4: Secret key configuration to be updated:
```console
web_1            | /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/invenio_config/default.py:74: UserWarning: Set configuration variable SECRET_KEY with random string
web_1            |   UserWarning)
web_1            | 
```

docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of CERN Open Data Portal.
-# Copyright (C) 2015, 2016, 2017 CERN.
+# Copyright (C) 2015, 2016, 2017, 2020 CERN.
```
diegodelemos, line 4: CSRF validation will be performed by default, are we affected?
```console
web_1            | /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/invenio_rest/ext.py:30: FutureWarning: CSRF validation will be enabled by default in the version 1.3.x
web_1            |   self.init_app(app)
```

diegodelemos:
Using this PR following [DEVELOPING.rst#installation](https://github.com/cernopendata/opendata.cern.ch/blob/master/DEVELOPING.rst#installation) with both `docker-compose.yml` and `docker-compose-dev.yaml` . Found issues:
1. [Missing HTTPS setup in `docker-compose.yaml`](https://github.com/cernopendata/opendata.cern.ch/pull/2968#discussion_r545061884)
2. [Warning: Celery version 5 to 6 deprecation warnings](https://github.com/cernopendata/opendata.cern.ch/pull/2968#discussion_r545062593)
3. [Warning: Elasticsearch 5 deprecation](https://github.com/cernopendata/opendata.cern.ch/pull/2968#discussion_r545063894)
4. [Warning: Secret key config](https://github.com/cernopendata/opendata.cern.ch/pull/2968#discussion_r545064402)
5. [Warning: CSRF validation by default coming in next Invenio-REST versions](https://github.com/cernopendata/opendata.cern.ch/pull/2968#discussion_r545065487)
6. Event visualiser http://localhost/visualise/events/cms not loading files:
image
7. Histograms http://0.0.0.0/visualise/histograms seem to not work
image
8. Search seems to work, needs to be double-checked in depth.
9. File download: working in dev mode, not in production mode (502 from nginx).

I believe 4. and 5. are not something we should take into account. For 4., we do not store any encrypted data in the database and 5., I am not aware of forms/data submission that needs to be validated.

requirements-production-local-forks.txt
```diff
@@ -1,3 +1,3 @@
 git+https://github.com/lepture/mistune-contrib.git#egg=mistune-contrib
-git+https://github.com/pamfilos/invenio-files-rest.git@eos-storage#egg=invenio-files-rest
+git+https://github.com/diegodelemos/invenio-files-rest.git@eos-storage#egg=invenio-files-rest
```
diegodelemos, line 2: Rebased on top of [latest master](https://github.com/inveniosoftware/invenio-files-rest/commits/master). Last release was 1.2.0 and commits on top of it are only related to docs and tests.

Dockerfile
```diff
@@ -99,11 +98,7 @@ ENV DEBUG=${DEBUG:-""}
 # Install Python packages needed for development
 RUN if [ "$DEBUG" ]; then pip install -r requirements-dev.txt; fi;
 
-# Change user from root to invenio for better security
-RUN adduser --uid 1000 invenio --gid 0 && \
-    chown -R invenio:root /code
-USER 1000
```
tiborsimko, line 105: `inveniosoftware/centos7-python:3.6` uses `root` UID/GID by default. Removing this bit will keep using `root`. Are we OK with that?  (Well, deploying on OpenShift should be fine anyway I guess...)

Dockerfile
```diff
@@ -22,9 +22,9 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
-# Use CentOS7
-FROM centos:7
+FROM inveniosoftware/centos7-python:3.6
 
+ENV XROOTD_VERSION=4.12.2
```
tiborsimko, line 27: Perhaps leave a small note about why we pin to 4.12.2 because of those XRootD  `-v` versions. Just so that it would be obvious for next time.

docker-compose.yml
tiborsimko: IIRC it should be OK.

Dockerfile
```diff
@@ -77,10 +74,12 @@ RUN pip install -r /tmp/requirements-production.txt
...
 COPY . /code
 
+RUN pip install -e ".[all]"
+
 # Create instance
 RUN /code/scripts/create-instance.sh && \
-    chgrp -R 0 ${APP_INSTANCE_PATH} && \
-    chmod -R g=u ${APP_INSTANCE_PATH}
+    chgrp -R 0 "${INVENIO_INSTANCE_PATH}" && \
+    chmod -R g=u "${INVENIO_INSTANCE_PATH}"
```
tiborsimko, line 82: The `PATH` changes are affecting also the discoverability of the `cernopendata` executable.

I see that it is now installed in:

```
/opt/rh/rh-python36/root/usr/bin/cernopendata
```

Let's add `/opt/rh/rh-python36/root/usr/bin/` path to default `PATH` so that the old commands like:

```
docker exec -i -t opendatacernch_web_1 cernopendata fixtures records -f /code/cernopendata/modules/fixtures/data/records/cms-derived-csv-Run2011A.json --mode insert
```

will keep working.

requirements-production-local-forks.txt
tiborsimko: WRT custom invenio-files-rest hacks, is file download supposed to work? I've tried to run in the production mode:

```console
$ docker-compose build
$ docker-compose up -d
$ docker exec -i -t opendatacernch_web_1 /code/scripts/clean-instance.sh
$ docker exec -i -t opendatacernch_web_1 ./scripts/populate-instance.sh --skip-records
$ docker exec -i -t opendatacernch_web_1 /opt/rh/rh-python36/root/usr/bin/cernopendata fixtures records -f /code/cernopendata/modules/fixtures/data/records/cms-derived-csv-Run2011A.json --mode insert
$ firefox http://localhost/record/5208
```

which works well :+1: but accessing the file itself (`http://localhost/record/5208/files/Zmumu.csv`) fails for now...

Dockerfile
```diff
@@ -22,9 +22,9 @@
 # waive the privileges and immunities granted to it by virtue of its status
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 
-# Use CentOS7
-FROM centos:7
+FROM inveniosoftware/centos7-python:3.6
```
tiborsimko, line 25: I noticed that there is [only one tag](https://hub.docker.com/r/inveniosoftware/centos7-python/tags) for this Docker image while there have been [several builds](https://hub.docker.com/r/inveniosoftware/centos7-python/builds) in the past six months.  Is the current `3.6` tag stable enough for production use? Shall we try to push for adopting semver2 image numbering in `inveniosoftware/centos7-python` for better safety and reproducibility?

Dockerfile
diegodelemos: This is because we are using the RH software collections and the python bin folder won't be exported until `scl enable` is run...

I am adding it to the path, but, could we also just run everything through bash `docker exec -i -t opendatacernch_web_1 bash -c cernopendata` and don't rely on the specific path?

Dockerfile
diegodelemos: Good catch: This is actually good in the base image (see [here](https://github.com/inveniosoftware/docker-invenio/blob/b9f65fdae6758d2efdb1f303e1d94ecb8d65434e/python3.6/Dockerfile#L79) the problem is that we do not set the user properly in our Dockerfile. I am adding it, thanks!

Dockerfile
tiborsimko: Adding `bash -c ...` also works, please just edit `DEVELOPING.rst` to amend the example commands...

requirements-production-local-forks.txt
diegodelemos: This should be fixed now. Regarding the problem itself, [we were creating a bucket for records](https://github.com/cernopendata/opendata.cern.ch/blob/a06b6bfdfa603dad2605405e121b5ea128d9c9c2/cernopendata/modules/fixtures/cli.py#L102) (which was correct in the past) but now [records come with buckets by default](https://invenio-records-files.readthedocs.io/en/latest/usage.html#creating-a-record), so we would populate the wrong bucket leaving the Record without files (explanation in commit message too).

The logic for update record needs to be updated too, shall we take this in a different PR?

diegodelemos: > The logic for update record needs to be updated too, shall we take this in a different PR?

Tested with current code and all works fine (because we clean the buckets and the records already exist, it is fine to create the buckets ourselves).

setup.py
```diff
@@ -67,41 +66,43 @@
 ]
 
 install_requires = [
-    'Flask-BabelEx>=0.9.3',
-    'Flask-Breadcrumbs>=0.4.0',
-    'Flask-CeleryExt>=0.2.2',
+    # OD adapted Invenio
```
tiborsimko, line 69: Regarding OD, we have COD3 and COD in the code base already. It might be good to harmonise this, e.g. just say COD3 here. 

However, in the series of  "OD adapted..." comments,  we could simply skip using acronyms, since the context is clear, we are building Invenio web app. So we could simply say "Custom Invenio `files` bundle`" and stuff.  (otherwise we'd have to explain how "adapted" the modules are, isnce most of them we are use vanilla, I think. Only a few required adaptations.)

Just a cosmetic naming comment.

requirements-production.txt
```diff
@@ -1,154 +1,173 @@
...
-webencodings==0.5.1
-Werkzeug==0.14.1
-whichcraft==0.4.1
-WTForms==2.1
-xrootdpyfs==0.1.6
+#
+# This file is autogenerated by pip-compile
+# To update, run:
+#
+#    pip-compile --output-file=requirements-production.txt
```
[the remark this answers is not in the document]
diegodelemos: I suspect I run it inside macOS since I can see the command in my history. Which version of pip-tools did you have to generate this list? With the latest `pip-compile` I get a very different format.

Dockerfile
diegodelemos: This hasn't been closed due to an incompatibility with `./run-tests.sh check-pytest`. To be addressed in a different PR.

requirements-production-local-forks.txt
tiborsimko: Works locally via browser :+1:  Will need adaptations to `cernopendata-client` though, CC @ParthS007. We can take this later.

requirements-production.txt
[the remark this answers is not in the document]
diegodelemos: Updated using:
```
$ docker-compose exec web bash
[root@2c3de0693db4 code]# pip install pip-tools==5.4.0
[root@2c3de0693db4 code]# requirements-production.txt
```

docker-compose-dev.yml
diegodelemos: This should be handled by [Invenio-Celery](https://github.com/inveniosoftware/invenio-celery/blob/6186820328aa2e1a70565f31beab7a19f8700092/invenio_celery/config.py). Probably not happening any time soon, see [Celery version pinning](https://github.com/inveniosoftware/invenio-celery/blob/6186820328aa2e1a70565f31beab7a19f8700092/setup.py#L45).

PR #2972, opened 2020-12-17
Add image for HI release note

katilp:


RaoOfPhysics:
Do you think we should reduce the resolution of the image and optimise it for the web? At the moment, it's nearly 5 MB. I can help with this if you wish.

RaoOfPhysics:
This is ~280 KB in size. Reduced width to 1440 pixels and the ran it through an image-optimisation software.

cms_hi_event_1

katilp:
Thanks! It looks great still. I'll push that to this same PR with _light at the end of the file name

tiborsimko:
Thanks! Since we shall serve only the light version via the portal, I'll merge only that one in order to make site container images slimmer...

PR #2983, opened 2021-01-06
installation: ssl config and secure headers

### LINKED ISSUE #2980: docker: add HTTPS setup on production mode
Disabling HTTPS temporarily. Flask-Talisman would force HTTPS on the clients but we don't have SSL setup in docker-compose production, so installation would break now.

_Originally posted by @diegodelemos in https://github.com/cernopendata/opendata.cern.ch/pull/2968#discussion_r545061884_

Configure production docker-compose to serve the application with HTTPS, as the production instance https://opendata.cern.ch/. Once done we can enable SSL enforcement from Invenio, see [here](https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/config.py#L63).

diegodelemos:
* Close #2980.


nginx/cernopendata.crt
```diff
@@ -0,0 +1,20 @@
+-----BEGIN CERTIFICATE-----
```
tiborsimko, line 1: The certificate says "cod-test.org" and "CERN Opendata".  Can you please update with better values?

- WRT "cod-test.org", that domain may look too broken to people trying this out. If this is for local developments only, then what about "localhost" or "localhost.localdomain" or somesuch; otherwise something like "opendata-dev.cern.ch" would suit better.
- WRT "CERN Opendata", the usual name is "CERN Open Data".

Something like:

```
Country Name (2 letter code) [GB]: CH
State or Province Name (full name) [Berkshire]: Switzerland
Locality Name (eg, city) [Newbury]: Geneva
Organization Name (eg, company) [My Company Ltd]: CERN
Organizational Unit Name (eg, section) []: IT-CDA-DR
Common Name (eg, your name or your server's hostname) []: localhost
Email Address []: opendata-team@cern.ch
```

nginx/cernopendata.conf
```diff
@@ -22,11 +22,16 @@ map $upstream_http_content_length $nocache {
...
 
 server {
-    listen 80 default;
+    listen 80;
+    listen 443 ssl; # remove in production, SSL termination managed by OpenShift
     server_name _ "";
     server_tokens off;
     charset utf-8;
 
+    # remove in production, SSL termination managed by OpenShift
```
tiborsimko, line 31: This may get unnoticed as the time goes by... Two possibilities to improve the visibility:

- e.g. starts with `# FIXME remove in production...` since we usually grep for FIXME, so this won't get unnoticed.
- e.g. rename file to `ngix/localhost.conf` to indicate this is for local developments only.

docker-compose.yml
```diff
@@ -201,6 +201,7 @@ services:
     image: cernopendata/nginx
     ports:
       - "80:80"
+      - "443:443"
```
tiborsimko, line 204: Speaking of SSL, there is quite a difference in the number of warnings seen in the web browser console. Shall we address them now or later?

HTTP:

http

HTTPS:

https

diegodelemos: They seem very localised, `csrftoken` and static assets. Perhaps we can fix it in a separate issue as it is "independent" from this task (altough it might be related with the Flask-Talisman config, there are chances we would hit this once we deploy and serve the upgraded Invenio site through HTTPS). Also it would allow us to move faster merging what we have today in the pipeline, what do you think?

tiborsimko: Sure, we can ticketise this for later... Please update the other two things and we can merge.

tiborsimko (approved):

nginx/cernopendata.conf
tiborsimko: If we go for `localhost.conf` in `Dockerfile`, can we also rename `cernopenadat.crt` and `cernopendata.key` to `localhost.crt` and `localhost.key`, so that they would be consistent?

We could actually even remove `cernopendata.conf` if you keep those two `FIXME-remove-in-production` comment lines to decorate the snippets... So that we can test locally the exact file that will be deployed in production.

```diff
25c25,26
<     listen 80 default;
---
>     listen 80;
>     listen 443 ssl;
28a30,32
> 
>     ssl_certificate     /etc/ssl/cernopendata.crt;
>     ssl_certificate_key /etc/ssl/cernopendata.key;
```

PR #2984, opened 2021-01-08
config: enable CORS

### LINKED ISSUE #2811: Make files available to in-browser applications (CORS)
I am working on a [proof-of-concept](https://github.com/cbourjau/alice-rs) to run an entire analysis involving(ALICE's) ROOT files in the browser using webasambly (wasm). Unfortunately, opendata.cern.ch does not serve files with a [CORS header](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) so they cannot be consumed by in-browser applications. Given the explicit purpose of the the open data being public, I would argue that the expected behavior is to serve these files with a CORS header allowing any origin to access the data.

Examples of files served without the `Access-Control-Allow-Origin: <origin> | *` headers:

http://opendata.cern.ch/record/1102/files/ALICE_LHC10h_PbPb_ESD_139038_file_index.txt
http://opendata.cern.ch//eos/opendata/alice/2010/LHC10h/000139038/ESD/0001/AliESDs.root

diegodelemos:
* Addresses #2811.

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -63,6 +63,9 @@ LOGGING_SENTRY_CELERY = os.environ.get(
 APP_ENABLE_SECURE_HEADERS = False
 
 
+#: CORS
+REST_ENABLE_CORS = True
+
 # Assets
 # ======
 #: Switch of assets debug.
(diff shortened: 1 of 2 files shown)


diegodelemos:
This PR activates CORS only for REST API (`/api`)  however files are download through `/records/<recid>/files/<filename>`.
Closing it as the solution doesn't cover the use case of downloading big files, for that we have to first tackle https://github.com/inveniosoftware/invenio-files-rest/issues/209.

PR #2985, opened 2021-01-08
docker: upgrade node version

### LINKED ISSUE #2981: docker: upgrade node version
We are currently using the unsupported Node version 6:
```
$ docker-compose build
...
                     DEPRECATION WARNING                            

  Node.js 6.x LTS Boron is no longer actively supported!

  You will not receive security or critical stability updates for this version.
```

The invenio base image [already includes Node 14](https://github.com/inveniosoftware/docker-invenio/blob/b9f65fdae6758d2efdb1f303e1d94ecb8d65434e/python3.6/Dockerfile#L30). We have to just remove [the version downgrade from the Dockerfile](https://github.com/cernopendata/opendata.cern.ch/blob/66b2ebab85a9c8fd96b993d6ffa0acf45990aa4a/Dockerfile#L55-L60) and make the app work.

audrium:
- Upgraded `node` version from 6 to 14
- Upgraded `npm` version from `3.10.10` to `6.14.5`
- Upgraded `node-sass` and `d3` npm packages to make it compatible with the newer `node` version
- Added `--unsafe-perm` param for installing npm packages to suppress permission error
- Temporary commented `opera` bundles

closes #2981

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,18 +50,9 @@ RUN yum install -y \
         https://storage-ci.web.cern.ch/storage-ci/xrootd/release/cc-7-x86_64/v${XROOTD_VERSION}/xrootd-client-devel-${XROOTD_VERSION}-1.el7.x86_64.rpm \
         https://storage-ci.web.cern.ch/storage-ci/xrootd/release/cc-7-x86_64/v${XROOTD_VERSION}/python3-xrootd-${XROOTD_VERSION}-1.el7.x86_64.rpm && \
     yum clean -y all
-# Downgrade Invenio base image to Node 6.x
-# hadolint ignore=DL3033
-RUN yum remove -y nodejs npm && \
-        curl -sL https://rpm.nodesource.com/setup_6.x -o /tmp/install-node6.sh && \
-        bash /tmp/install-node6.sh && \
-        rm /tmp/install-node6.sh && \
-        yum install -y nodejs && \
-        npm install -g npm@3.10.10 && \
-    yum clean -y all
 
 RUN pip uninstall pipenv -y && pip install --upgrade pip==20.2.4 setuptools==51.0.0 wheel==0.36.2 && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -50,18 +50,9 @@ RUN yum install -y \
...
-        curl -sL https://rpm.nodesource.com/setup_6.x -o /tmp/install-node6.sh && \
-        bash /tmp/install-node6.sh && \
-        rm /tmp/install-node6.sh && \
-        yum install -y nodejs && \
-        npm install -g npm@3.10.10 && \
-    yum clean -y all
 
 RUN pip uninstall pipenv -y && pip install --upgrade pip==20.2.4 setuptools==51.0.0 wheel==0.36.2 && \
-    npm install -g node-sass@3.8.0 clean-css@3.4.24 requirejs@2.3.6 uglify-js@3.12.1 jsonlint@1.6.3 d3@3.3.13
+        npm install -g --unsafe-perm node-sass@4.14.1 clean-css@3.4.24 requirejs@2.3.6 uglify-js@3.12.1 jsonlint@1.6.3 d3@6.3.1
```
diegodelemos, line 55: Minor: indentation should be aligned with previous line.

Not for this PR: I've tried to figure out what is going on here because in case we spot something we should fix it upstream. I am writing down what I found for future lookup:
- Couldn't find anything other than prefixing this command with `runuser -u invenio -- npm install -g node-sass@....`. This is because the [global npm directory belongs to `invenio`].(https://github.com/inveniosoftware/docker-invenio/blob/b9f65fdae6758d2efdb1f303e1d94ecb8d65434e/python3.6/Dockerfile#L82)
- `npm_config_[user|group]` seem to be correct by default:
  ```
  $ docker run -ti --rm inveniosoftware/centos7-python:3.6 bash
  > npm config ls -l | grep "user =\|group ="
  group = 0
  user = 1000
  ```
  According to docs ([`user`](https://docs.npmjs.com/cli/v6/using-npm/config#user), [`group`](https://docs.npmjs.com/cli/v6/using-npm/config#group)) the error shouldn't appear, even if npm is run as root it should know which user ID to use?

diegodelemos (approved):

PR #2987, opened 2021-01-08
frontend: removed all angular packages from npm bundles

### LINKED ISSUE #2923: frontend: migrate to latest Invenio 3 stable
Following up the backend migration #2922, the goal of this task is to adapt latest front end packages and record templates to display the records.

This task won't necessarily have to concern itself with home page, search UI, etc. The task will be fulfilled if `/record/1` is displayed nicely using latest Invenio packages.

audrium:
- commented out all `angular` dependencies so that it would be visible which ones was used while working on the `react` migration
- tested this PR on [upgraded `node` version](https://github.com/cernopendata/opendata.cern.ch/pull/2985)
- compiled a list of pages which still work and which broke after removing the `angular` packages
    - in summary all the pages which are using `jinja` templates only seems to work as expected
    - all the other ones which uses: `(i) search, (ii) file download box, (iii) visualise events (opera), (iv) visualise histograms` were broken and could be tackled individually
    
List of pages which are working as expected:
- "/visualise/events" 
- "/visualise/events/cms" 
- "/glossary/json" 
- "/sitemap.xml"
- "/ping"
- "/"
- "/docs/<pid_value>"
- 404 page

List of pages which are broken:
- "/visualise/histograms"
- "/visualise/histograms/cms"
- "/visualise/events/opera"
- "/md/debug"
- "/experiments" - same as /search
- "/resources" - same as /search
- "/education" - same as /search
- "/glossary" - same as /search
- "/research" - same as /search
- "/search"
- "/record/1002" - file download box is broken

closes #2923

### DIFF
--- a/cernopendata/modules/theme/bundles.py
+++ b/cernopendata/modules/theme/bundles.py
@@ -30,25 +30,25 @@ from flask_assets import Bundle
 from invenio_assets import NpmBundle
 
 js = NpmBundle(
-    'node_modules/angular/angular.js',
+    # 'node_modules/angular/angular.js',
     'node_modules/jquery/jquery.js',
     'node_modules/popper.js/dist/umd/popper.js',
     'node_modules/bootstrap/dist/js/bootstrap.js',
-    'node_modules/angular-ui-bootstrap/dist/ui-bootstrap-tpls.js',
+    # 'node_modules/angular-ui-bootstrap/dist/ui-bootstrap-tpls.js',
     output='gen/cernopendata.theme.%(version)s.js',
     npm={
-        'angular': '~1.4.9',
+        # 'angular': '~1.4.9',
         'jquery': '~1.9.1',
         'popper.js': '~1.11.0',
         'bootstrap': '~4.0.0-beta',
-        'angular-ui-bootstrap': '~2.2.0',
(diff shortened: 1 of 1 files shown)


cernopendata/modules/theme/bundles.py
```diff
@@ -30,25 +30,25 @@
 from invenio_assets import NpmBundle
 
 js = NpmBundle(
-    'node_modules/angular/angular.js',
+    # 'node_modules/angular/angular.js',
```
tiborsimko, line 33: What about the Angular directives? We have 'ng-if` and friends in many Jinja places. If we "close" the issue and not only "address" it, then I guess we'd have to remove all `ng-if`, `ng-controller` and friends from the pages that are declared working.  Is that OK already for all the checked endpoints? 

P.S. For `/search`, `/visualise/...` etc this will be naturally addressed elsewhere.

audrium: After scanning through the code, I only found Angular directives used in `/search` related templates:
- `cernopendata/templates/cernopendata_records_ui`
- `cernopendata/templates/cernopendata/search.html`

and also in components used by `/search` under the `/cernopendata/static`. So I guess it could be addressed while converting to `react`?

Issue #2919, opened 2020-11-26
ui: investigate Angular -> React options


tiborsimko:
Establish list of heavy Angular use cases in the current code base:
- search UI and facets
- paginated list of files for download

Estimate migration efforts:
- for search, adoption of `react-searchkit`
  - is faceting behaving the same? any adptations needed?
- for paginated list of files for download
  - is custom code still necessary? most probably yes, since we may have O(15,000) files in a given dataset, see http://opendata.cern.ch/record/1395

Form a quantitative opinion about migration effors, for example:
- option 1: keep current Angular 2.0 code and switch only Invenio backend and adapt Angular as needed;
- option 2: switch to React keeping custom search facet and paginated list code;
- option 3: switch to React and `react-searchkit` and develop any necessary additions;
- option 4: ...

tiborsimko:
One more special feature of search to take into account: `[ ] include on-demand datasets` checkbox on top of search facets.

tiborsimko:
Option 3 selected and implemented. The on-demand datasets not necessary as we shall be having more of these (e.g. on tape storage). Closing the epic.

PR #2994, opened 2021-01-14
elasticsearch: upgrade to Elasticsearch 7

### LINKED ISSUE #2978: elasticsearh: upgrade to Elasticsearch 7
Deprecation warnings about Elasticsearch5:
```console
web_1            | /opt/rh/rh-python36/root/usr/lib/python3.6/site-packages/invenio_search/ext.py:59: DeprecationWarning: Elasticsearch v2 and v5 support will be removed.
web_1            |   DeprecationWarning)
```

_Originally posted by @diegodelemos in https://github.com/cernopendata/opendata.cern.ch/pull/2968#discussion_r545063894_

Upgrade to a supported Elasticsearch version. At CERN the latest recommended version is `7`.

Note: maybe we can use [these docs](https://invenio.readthedocs.io/en/master/releases/upgrading/upgrade-3_1-3_2.html#upgrade-to-v7-by-reindexing) (:warning:  -  I haven't tested myself).

ParthS007:
closes #2978


docker-compose-dev.yml
```diff
@@ -173,7 +173,7 @@ services:
 
   elasticsearch:
     restart: "unless-stopped"
-    image: elasticsearch:5
+    image: elasticsearch:7.10.1
```
tiborsimko, line 176: Shall we just use 7.10 in order to have latest bugfix updates? (Or even 7?) The docker compose versions used only for development, the production upgrades are handled differently.

cernopendata/mappings/v7/records/record-v1.0.0.json
```diff
@@ -0,0 +1,166 @@
+{
+  "mappings": {
```
tiborsimko, line 2: Perhaps you can remove v5 mappings now that we have v7 everywhere?

cernopendata/mappings/v7/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+# You should have received a copy of the GNU General Public License
+# along with CERN Open Data Portal; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""CERN Open Data Elasticsearch v7 Mappings."""
```
tiborsimko, line 25: Please note the type in the commit log (`searh`).

I'll get to testing/merging after the OPERA UI stuff; just leaving some notes for now.  You'll be able to rebase against updated `master` soon...

cernopendata/mappings/v7/records/record-v1.0.0.json
ParthS007: I can remove it. I wanted to know if it is working fine on your side as well. If everything is good to go, I will remove it.

docker-compose-dev.yml
ParthS007: I used `elasticsearch:7` but I got this error: `ERROR: manifest for elasticsearch:7 not found: manifest unknown: manifest unknown`.

when I looked into `elasticsearch` docker hub, there are two tags that are available: `7.10.1` and `6.8.13`. I thought of putting `latest` but this tag is not supported.

https://hub.docker.com/_/elasticsearch

cernopendata/mappings/v7/records/record-v1.0.0.json
tiborsimko: It does not actually work out of the box. I cannot load records, as the ES container keeps restarting, with messages like:

```console 
$ docker-compose -f docker-compose-dev.yml down -v
$  docker-compose -f docker-compose-dev.yml build
$ docker-compose -f docker-compose-dev.yml up -d
$ docker exec -i -t opendatacernch_web_1 /code/scripts/populate-instance.sh --skip-records
...
$ docker logs opendatacernch_elasticsearch_1 
...
ERROR: [3] bootstrap checks failed
[1]: memory locking requested for elasticsearch process but memory is not locked
[2]: max virtual memory areas vm.max_map_count [65530] is too low, increase to at least [262144]
[3]: the default discovery settings are unsuitable for production use; at least one of [discovery.seed_hosts, discovery.seed_providers, cluster.initial_master_nodes] must be configured
ERROR: Elasticsearch did not exit normally - check the logs at /usr/share/elasticsearch/logs/docker-cluster.log
```

Seems like we need to set some new params that are not appearing in this PR's `docker-compose-dev.yml`?

docker-compose-dev.yml
diegodelemos: Perhaps we should be using the same ES as Invenio, see [here](https://github.com/inveniosoftware/cookiecutter-invenio-instance/blob/55645e9b56db0f52ae0235efa68c72bf173c1d40/%7B%7Bcookiecutter.project_shortname%7D%7D/docker-services.yml#L78), it seems to be the [recommended way](https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html).

ParthS007: Using the same as `invenio` sounds good. It is `7.9.3` which was released 3 months ago and `7.10.1` was released one month ago.

cernopendata/mappings/v7/records/record-v1.0.0.json
tiborsimko: Tested loading full site, and things mostly work, but I see some discrepancies comparing against production, e.g. see the total count:

```
/api/records

- localhost: "total": 7621
- DEV: "total": 5109
```

Some collections seem OK, for example:

```
/api/records?q=alice

- localhost; "total": 35
- DEV; "total": 35

/api/records?q=experiment:opera

- localhost; "total": 910
- DEV; "total": 910
```

The extra count seems to come from CMS records:

```
/api/records?q=experiment:cms

- localhost; "total": 6506
- DEV; "total": 3994
```

I haven't looked more deeply to see what is different. Perhaps we need to amend fixture loading or something... Can you please have a look?

P.S. If we need Elasticsearch 7 to unblock the upcoming work on the search UI adaptations (CC @mvidalgarcia), then we could merge "as is" and create a new issue for the "extra count" observations.

tiborsimko: OK, mystery understood :smile: The extra records (7621 vs 5109) are coming from "on-demand datasets" that we usually don't index by default in the QA and PROD and that we show in the results only when the corresponding "include on-demand datasets" check box in the UI is ticked. See https://github.com/cernopendata/opendata.cern.ch/issues/2919#issuecomment-734362529

Hence we can finalise this PR as is and address the on-demand parts later as part of the search UI and facets.

tiborsimko (approved):

PR #3000, opened 2021-01-19
global: initial Semantic UI config

diegodelemos:


cernopendata/modules/theme/assets/semantic-ui/scss/styles.scss
```diff
@@ -886,4 +886,4 @@ invenio-search-facets {
     p {
         margin: 0;
     }
-}
+}
```
mvidalgarcia, line 889: Minor: Missing end line.

cernopendata/modules/theme/webpack.py
```diff
@@ -27,21 +27,22 @@
...
     themes={
-        'bootstrap3': dict(
+        'semantic-ui': dict(
             entry={
-                'cernopendata_js': './js/app.js',
+                'cernopendata_css': './scss/styles.scss'
             },
             dependencies={
-                'angular': '~1.4.9',
                 'jquery': '~1.9.1',
```
mvidalgarcia, line 40: Moderate: How is this working exactly? Is `jquery` included in the CSS bundle? Don't we need a JS bundle to load these deps?

PS. Isn't `jquery` being loaded by [`invenio-theme`](https://github.com/inveniosoftware/invenio-theme/blob/d94bf4a5452a31214b1485bf2225b57de93799bb/invenio_theme/webpack.py#L70)?

cernopendata/modules/theme/webpack.py
```diff
@@ -27,21 +27,22 @@
...
+                'open-iconic': '~1.1.1',
                 'popper.js': '~1.11.0',
```
mvidalgarcia, line 42: Same here.

setup.py
```diff
@@ -127,13 +127,11 @@
...
         ],
         'invenio_assets.webpack': [
-            'cernopendata_theme = cernopendata.modules.theme.webpack:js',
+            'cernopendata_theme = cernopendata.modules.theme.webpack:theme',
         ],
         # 'invenio_assets.bundles': [
-            # 'cernopendata_js = cernopendata.modules.theme.bundles:js',
             # 'cernopendata_front_js = cernopendata.modules.theme.bundles'
             # ':front_js',
-            # 'cernopendata_theme_css = cernopendata.modules.theme.bundles:css',
```
mvidalgarcia, line 136: Minor: Maybe we can start deleting bundles from `bundles.py`? so we can keep track of what is migrated.

cernopendata/templates/cernopendata_theme/page.html
```diff
@@ -1,6 +1,8 @@
 {% extends 'invenio_theme/page.html' %}
 
 {% block css %}
+    {{ super() }}
+    {{ webpack['cernopendata_css.css'] }}
     {# {%- assets "cernopendata_theme_css" %}<link href="{{ ASSET_URL }}" rel="stylesheet"> {%- endassets %} #}
```
mvidalgarcia, line 6: Minor: We can probably remove this one.

cernopendata/modules/theme/webpack.py
```diff
@@ -27,21 +27,22 @@
 from invenio_assets.webpack import WebpackThemeBundle
 
 
-js = WebpackThemeBundle(
```
mvidalgarcia, line 30: Question: why we got rid of the JS bundle? How is e.g. `popper.js` being loaded?

cernopendata/modules/theme/assets/semantic-ui/less/cernopendata/theme.config
```diff
@@ -0,0 +1,120 @@
...
+
+/*******************************
+            Folders
+*******************************/
+
+/* Path to theme packages */
+@themesFolder : 'themes';
+
+/* Path to site theme folder */
+@siteFolder  : '../../less/my_site/site';
```
mvidalgarcia, line 110: Moderate: I think this path is incorrect.

mvidalgarcia (commented):
Minor: s/SematicUI/SemanticUI/ in commit message.

cernopendata/modules/theme/webpack.py
diegodelemos: Thanks for the catch, removing `jquery` as it indeed comes from `invenio-theme`.

diegodelemos: P.S. We will probably be breaking things because Invenio is installing `jquery` `3.2.1`, while here we require `1.9.1`. Perhaps to be investigated on a per-case basis?

cernopendata/modules/theme/webpack.py
diegodelemos: It is now included inside the `theme` bundle (I am basically following the recommended way in cookiecutter invenio instance, see [here](https://github.com/inveniosoftware/cookiecutter-invenio-instance/blob/master/%7B%7Bcookiecutter.project_shortname%7D%7D/%7B%7Bcookiecutter.package_name%7D%7D/theme/webpack.py)).

The usage ends up being `{{ webpack['cernopendata_css.css'] }}` for CSS and `{{webpack['cernopendata_js.js'] }}` for JS, both being picked from the [`entry`](https://github.com/cernopendata/opendata.cern.ch/pull/3000/files#diff-8cbc780ffa24ccb6fdbb0958a7ce997f655647976c428e0d25d19ebe80d17e19L36) dictionary.

mvidalgarcia: That I understand. But I don't see a JavaScript entry bundle, `cernopendata_js` has been deleted if I'm not wrong.

diegodelemos: When I meant now I meant with the force push, `cernopendata_js.js` was indeed missing, thanks for spotting.

mvidalgarcia: OK, I see that [you've just added it](https://github.com/cernopendata/opendata.cern.ch/compare/503f0ab7c4126fad19760609b6cf5222768f9a7c..3ebe5f5804860cd9f75a4770d8a4e068844f6ff5) back...

mvidalgarcia (approved):

Issue #2997, opened 2021-01-18
frontend: fix UI according to new webpack build system


diegodelemos:
1. Upgrade last Invenio UI packages:
   - [x] Upgrade to latest Invenio-Search-UI (https://github.com/cernopendata/opendata.cern.ch/pull/3015 @mvidalgarcia)
   - [x] Upgrade to latest Invenio-Records-UI (https://github.com/cernopendata/opendata.cern.ch/pull/2998 @diegodelemos)
   - [x] Upgrade to latest Invenio-Previewer (https://github.com/cernopendata/opendata.cern.ch/pull/3002 @audrium)

2. Upgrade bundles to new webpack build system (documentation [here](https://invenio-assets.readthedocs.io/en/latest/upgrade.html))
   - [x] [Global CSS bundle](https://github.com/cernopendata/opendata.cern.ch/blob/76abc14f0f49f58a14a9c3b792c6daaeab67824a/cernopendata/modules/theme/bundles.py#L48) (https://github.com/cernopendata/opendata.cern.ch/pull/3000 @diegodelemos)
   - [x] [Glossary bundle CSS and JS](https://github.com/cernopendata/opendata.cern.ch/blob/76abc14f0f49f58a14a9c3b792c6daaeab67824a/cernopendata/modules/theme/bundles.py#L112-L125) (@diegodelemos)
   - [x] [Front JS bundle (not working, do we want it?)](https://github.com/cernopendata/opendata.cern.ch/blob/76abc14f0f49f58a14a9c3b792c6daaeab67824a/cernopendata/modules/theme/bundles.py#L68) (@mvidalgarcia) - Removed from now
   - [x] [Visualise bundle](https://github.com/cernopendata/opendata.cern.ch/blob/76abc14f0f49f58a14a9c3b792c6daaeab67824a/cernopendata/modules/theme/bundles.py#L91-L110)
   - [x] [Opera viewer](https://github.com/cernopendata/opendata.cern.ch/blob/76abc14f0f49f58a14a9c3b792c6daaeab67824a/cernopendata/modules/theme/bundles.py#L127-L168)
   - [x] [ispy bundle](https://github.com/cernopendata/opendata.cern.ch/blob/76abc14f0f49f58a14a9c3b792c6daaeab67824a/cernopendata/modules/theme/bundles.py#L170-L223)
[...]

PR #3015, opened 2021-01-21
frontend: upgrade search UI

### LINKED ISSUE #3014: frontend: upgrade to latest Invenio-Search-UI
The goal of this issue is having a MVP of search using React and Semantic-UI. Further adaptations such as custom code or facets will be tackled in other issues.

- [x] Upgrade to latest `invenio-search-ui` Python package
- [x] ~Adapt bundles accordingly~ Using default bundle
- [x] Use `react-searchkit` components to have MVP search results
- [x] Get rid of all unused JS (angular), templates, and CSS.

mvidalgarcia:
- upgrade to latest invenio-search-ui (2.0.0)
- use default search template
- create react components for various result types
- get rid of unused templates, JS, styles...
- MVP funcionality

closes #3014
addresses https://github.com/cernopendata/opendata.cern.ch/issues/2997

image

### To test

- Search for a term in the search bar (e.g. http://0.0.0.0:5000/search?q=higgs)
- Check different facets
- Use pagination
- ... try to break it and report issues :)


cernopendata/config.py
```diff
@@ -341,42 +350,43 @@
...
-                            'to': 9999999
-                        },
-                        {
-                            'key': '10000000--',
-                            'from': 10000000
-                        }
-                    ]
-                }
-            }
+            # TODO: Breaks the search (KeyError: 'query'), adapt or remove?
```
mvidalgarcia, line 353: @ParthS007: This is one the issues I had. Do you have a clue how to adapt it? cc/ @tiborsimko @diegodelemos

cernopendata/config.py
```diff
@@ -295,11 +295,20 @@
...
+#     }
+# }
+
+# TODO: based on https://github.com/inveniosoftware/invenio-records-rest/blob/1c44f4c1490e48064cb3288168ff9c63cb5872a9/invenio_records_rest/config.py#L300
+RECORDS_REST_DEFAULT_SORT = dict(
+    _all=dict(
+        query='bestmatch',
+        noquery='mostrecent',
+    )
+)
```
mvidalgarcia, line 311: @ParthS007: Also this one

cernopendata/config.py
diegodelemos: Since the range filter hasn't changed since the last versions, I think the problem is merely presentational which is good. Invenio-Search-UI does not return config to React-SearchKit other than for [term aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html) (see [here](https://github.com/inveniosoftware/invenio-search-ui/blob/44b6122d8f066a92344f1c1a58b4d18e888b690c/invenio_search_ui/views.py#L191-L211)), but we also need [range aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-range-aggregation.html).

From what I understand the fix needs to be done in Invenio-Search-UI, allowing to return config for other kinds of aggregations. A _dirty local fix_ to see if things work now is to modify `/opt/invenio/var/instance/python/lib/python3.6/site-packages/invenio_search_ui/views.py` as follows:
```diff
        return [
            {
                "title": k.capitalize(),
                "agg": {
                    "aggName": k,
-                   "field": v["terms"]["field"]
+                   "field": (v.get("terms") or v.get("range")).get("field")
                },
            }
            for k, v in aggs.items()

```
Note: Of course this is not a good solution, I guess we need a function that gets `field` for any of the supported aggregations? What do you think?

cernopendata/modules/theme/assets/semantic-ui/templates/search/utils.js
```diff
@@ -0,0 +1,27 @@
...
+ * along with CERN Open Data Portal; if not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+ * MA 02111-1307, USA.
+ *
+ * In applying this license, CERN does not
+ * waive the privileges and immunities granted to it by virtue of its status
+ * as an Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+export const stripHtml = (input) => input?.replace(/<[^>]*>?/gm, "");
```
diegodelemos, line 27: 🪄

diegodelemos (commented):
Tested locally and basic functionality works great 🚀 Just a few minor comments and [the range filters trouble](https://github.com/cernopendata/opendata.cern.ch/pull/3015/files#r561991754) which seems a bit more complicated, perhaps we can move to a new separate issue.

cernopendata/modules/theme/assets/semantic-ui/templates/search/ResultsGrid.item.jsx
```diff
@@ -0,0 +1,50 @@
...
+ * In applying this license, CERN does not
+ * waive the privileges and immunities granted to it by virtue of its status
+ * as an Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+import React from "react";
+import { Card, List } from "semantic-ui-react";
+
+export const CODResultsGridItem = ({ result, index }) => {
+  const contributors = result.metadata.contributors || [];
```
audrium, line 31:
```suggestion
  const { contributors } = result.metadata;
```

cernopendata/modules/theme/assets/semantic-ui/templates/search/ResultsGrid.item.jsx
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * In applying this license, CERN does not
+ * waive the privileges and immunities granted to it by virtue of its status
+ * as an Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+import React from "react";
+import { Card, List } from "semantic-ui-react";
+
+export const CODResultsGridItem = ({ result, index }) => {
```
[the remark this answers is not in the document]
mvidalgarcia: This comes from `cookiecutter-invenio-instance` by default, that's why is there, I haven't adapted it though. I see we have a condensed view on PROD so it might not be that misguided? I can get rid of it but I'm not sure what's gonna happen with the view selector. I guess the question is, do we want to support the condensed view? Also, should it be part of this PR?

cernopendata/modules/theme/assets/semantic-ui/templates/search/components/DefaultListItem.js
```diff
@@ -0,0 +1,105 @@
...
+          <p>
+            {truncate(stripHtml(metadata.abstract?.description), {
+              length: 200,
+            })}
+          </p>
+        </Item.Description>
+        <Item.Extra className="badges-box">
+          <Label
+            className="badge badge-type"
+            href={`/search?type=${metadata.type.primary}`}
```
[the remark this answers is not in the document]
mvidalgarcia: Interesting.. didn't know this. Out of curiosity, where is this configured? and why it was [like this](https://github.com/cernopendata/opendata.cern.ch/pull/3015/files#diff-92503841c3a88215b2b39df14d48f9afabb8ef11ae211f1e03d1d47dea4e4d5cL11) before?

cernopendata/modules/theme/assets/semantic-ui/templates/search/utils.js
mvidalgarcia: I'm not very proud of this TBH.. it's probably a temporary solution... Angular (I think) was [taking care of this](https://github.com/cernopendata/opendata.cern.ch/pull/3015/files#diff-f669327ae6143a58c0da8767f3f3b62263e36cbab79c346f8965380fc8109546L27), otherwise, HTML tags are visible, and since we trim the result is not safe to actually print it as HTML.

cernopendata/modules/theme/assets/semantic-ui/templates/search/ResultsGrid.item.jsx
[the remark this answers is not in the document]
diegodelemos: Thanks for the clarification!

> Also, should it be part of this PR?

Agreed, not to be included here. Once we merge this PR we have to create new issues for the different search features, one of them can be about this.

cernopendata/modules/theme/assets/semantic-ui/templates/search/utils.js
diegodelemos: It worked well, I've checked a few pages of search and I didn't find any inconsistency. In the end is just one line, well isolated and what it does is well known, so not a big deal.

cernopendata/config.py
mvidalgarcia: Thanks for the detailed explanation, now I understand better the issue. I've tested your snippet and it stopped failing.
I guess there's not much we can do apart from contributing upstream to `invenio-search-ui`? How come that it was working well with the old version of `invenio-search-ui`?

cernopendata/modules/theme/assets/semantic-ui/templates/search/ResultsGrid.item.jsx
mvidalgarcia: Actually, this is a leftover because we don't print contributors anywhere in the search results.

cernopendata/config.py
mvidalgarcia: OK, I see that's being used [here](https://github.com/inveniosoftware/invenio-search-ui/blob/44b6122d8f066a92344f1c1a58b4d18e888b690c/invenio_search_ui/assets/semantic-ui/js/invenio_search_ui/components/SearchApp.js#L37) to display the multiple bucket aggregation components.
image

It shouldn't be a big deal to sync with Invenio team to contribute upstream, should it?

diegodelemos: > It shouldn't be a big deal to sync with Invenio team to contribute upstream, should it? 

No big deal, I just wanted to first double-check with you. Since you can reproduce it and it seems like a reasonable request we can ask and contribute to extend this behavior.

> How come that it was working well with the old version of `invenio-search-ui`?

Probably because all this info was populated [through Jinja templates](https://github.com/inveniosoftware/invenio-search-ui/blob/master/invenio_search_ui/templates/invenio_search_ui/search.html), and now there is this new function which is [specific to React SearchKit](https://github.com/inveniosoftware/invenio-search-ui/blob/44b6122d8f066a92344f1c1a58b4d18e888b690c/invenio_search_ui/views.py#L194) to render the facets through an API call.

diegodelemos (approved):

Issue #3007, opened 2021-01-21
frontend: migrate markdown renderer to Semantic UI


diegodelemos:
Make markdown renderer looklike production, e.g. [content rendered](http://opendata.cern.ch/docs/about-alice) and [`.md` source](https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/modules/fixtures/data/docs/alice-about/alice-about.md). Related template [here](https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/templates/cernopendata_pages/md_template.html).

diegodelemos:
See second two points from https://github.com/cernopendata/opendata.cern.ch/pull/3016#issuecomment-767462412, tl;dr tables do not render plus improvable code block rendering.

PR #3016, opened 2021-01-21
frontend: migrate docs record view to Semantic UI

### LINKED ISSUE #3004: frontend: migrate docs record view to Semantic UI
Make docs record view look like [production](http://opendata.cern.ch/docs/welcome). Related template [here](https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/templates/cernopendata_records_ui/docs/detail.html).

diegodelemos:
Closes #3004

This can be tested with any of the records [listed here](http://opendata.cern.ch/search?page=1&size=20&q=&type=Documentation&subtype=About&subtype=Activities&subtype=Authors&subtype=Guide&subtype=Help&subtype=Policy&subtype=Report). To populate them locally:
```
$ docker exec -i -t opendatacernch_web_1 /code/scripts/populate-instance.sh --skip-records
```
Special case with related records:  http://opendata.cern.ch/docs/cms-physics-objects-2010

### DIFF
--- a/cernopendata/templates/cernopendata_records_ui/docs/detail.html
+++ b/cernopendata/templates/cernopendata_records_ui/docs/detail.html
@@ -1,64 +1,62 @@
 {%- extends "invenio_records_ui/detail.html" %}
 
 {%- block page_body %}
-<div class="container-fluid background">
-    <div class="container detail-page" ng-app="recordApp">
-        <div class="row">
-            <div class="{% if record.related %} col-md-10 {% else %} col-md {% endif %}">
-                <div class="card card-style">
-                    <div class="card-body">
-                        <h1 class="d-inline"> {{ record.title }} </h1>
-                        <p>
-                        {% if record.get('created') %} <small>{{ record.get('created') }} by </small> {% endif %}
-                        <small>{{ record.author}} </small>
-                        </p>
-                        <p class="badges-box">
-                        {% if record.type %} <a class="badge badge-type" href="/search?type={{ record.type.primary }}">{{ record.type.primary }}</a>
-                        {% if record.type.secondary %}
-                        {% for type in record.type.secondary %}
-                        <a class="badge badge-subtype" href="/search?type={{ record.type.primary }}&subtype={{ type }}">{{ type }}</a>
-                        {% endfor %}
+<div class="ui {% if record.related %}two{% else %}one{% endif %} centered column stackable grid container detail-page">
+    <div class="row">
+        <div class="{% if record.related %}thirteen wide column{% else %}sixteen wide column{% endif %}">
+            <div class="ui segment">
+                <div class="card-body">
+                    <h1> {{ record.title }} </h1>
+                    <p>
+                    {% if record.get('created') %} <small>{{ record.get('created') }} by </small> {% endif %}
+                    <small>{{ record.author}} </small>
+                    </p>
+                    <p class="badges-box">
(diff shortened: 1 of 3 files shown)


cernopendata/templates/cernopendata_records_ui/docs/detail.html
```diff
@@ -1,64 +1,62 @@
...
-                        {% if record.type.secondary %}
-                        {% for type in record.type.secondary %}
-                        <a class="badge badge-subtype" href="/search?type={{ record.type.primary }}&subtype={{ type }}">{{ type }}</a>
-                        {% endfor %}
-                        {% endif %}
+<div class="ui {% if record.related %}two{% else %}one{% endif %} centered column stackable grid container detail-page">
+    <div class="row">
+        <div class="{% if record.related %}thirteen wide column{% else %}sixteen wide column{% endif %}">
+            <div class="ui segment">
+                <div class="card-body">
```
diegodelemos, line 8: @mvidalgarcia I left this `card-body` class as is because I saw usages in search, however, now that https://github.com/cernopendata/opendata.cern.ch/issues/3014 is out, I see we don't really need it (just asking for confirmation). If that is the case, then I can perhaps amend it.

audrium:
I noticed that for example in [`/docs/alice-virtual-machine`](http://opendata.cern.ch/docs/alice-virtual-machine) titles of paragraphs are rendered as links in a blue colour. This happens because it's an actual link: `<a name="vbox">Step 1: Installing VirtualBox</a>`. Here's what it looks like:
image

audrium:
Small things which I spotted after comparing multiple different pages against production:
- header elements are bolder now, I noticed that in production we use: `font-weight: 500;` to make it thinner. But I'm not sure if it's super important to keep the look and feel exactly the same as before, now it also looks good.
- text in paragraphs are rendered in smaller font size now. (before it was `font-size: 1.1rem;`, now `1rem`)
- `<code>` blocks in some pages rendered differently (it also misses horizontal scrollbar). Example in [/docs/cms-getting-started-2011](http://localhost:5000/docs/cms-getting-started-2011) and in [/docs/cms-mc-production-overview](http://localhost:5000/docs/cms-mc-production-overview)
- tables are not rendered now. Example in [/docs/cms-simulated-dataset-names](http://localhost:5000/docs/cms-simulated-dataset-names) page.

diegodelemos:
For the first two points perhaps we can look at it in a separate issue once the site is more stable?

Regarding the second two points, they will most probably have to be fixed as part of https://github.com/cernopendata/opendata.cern.ch/issues/3007 as the HTML is generated by a markdown extension.

audrium (approved):

cernopendata/templates/cernopendata_records_ui/docs/detail.html
mvidalgarcia: It's being used [here](https://github.com/cernopendata/opendata.cern.ch/pull/3020/files#diff-e57c82b46c275a58fdeb7b4dc6500177cc1afbe84ec909d3f861e38c58d24d4cR9) so we need it.

mvidalgarcia:
> For the first two points perhaps we can look at it in a separate issue once the site is more stable?

Agreed, global stuff that changed can be tackled via `site.variables` file, as I started doing [here](https://github.com/cernopendata/opendata.cern.ch/pull/3020/files#diff-bd7e287accd552886f99dfe3b5841049c8c60b94ad73fa9db3e966fc0d6dd40a). To be completed on the way.

Issue #3009, opened 2021-01-21
frontend: migrate OPERA event visualiser to Semantic-UI


diegodelemos:
Make OPERA event visualiser look like [production](http://opendata.cern.ch/visualise/events/opera). Related template [here](https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/templates/cernopendata_pages/visualise_events_opera.html) and [bundle](https://github.com/cernopendata/opendata.cern.ch/blob/32b59d0b52fa639e2bb5824aa932f810f849c20d/cernopendata/modules/theme/bundles.py#L91-L132).

PR #3019, opened 2021-01-25
ui: migrate Opera visualiser bundle from using NpmBundle

audrium:
addresses #3009


diegodelemos (commented):
Works fine locally on dev and prod mode. It seems like a legit solution for the time being.

Regarding how this may affect production: taking into account that we are copying the whole `$INVENIO_INSTANCE_PATH/static` folder from the web image inside the production `nginx` container, this solution should work also on our OpenShift deployment.

diegodelemos (approved):

PR #3020, opened 2021-01-25
layout: migrate frontpage elements to semantic-ui

### LINKED ISSUE #3006: frontend: migrate frontpage to Semantic UI
Make front page look like [production](http://opendata.cern.ch/). Related [template](https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/templates/cernopendata_pages/front) and [bundle](https://github.com/cernopendata/opendata.cern.ch/blob/32b59d0b52fa639e2bb5824aa932f810f849c20d/cernopendata/modules/theme/bundles.py#L32-L38).

mvidalgarcia:
- layout: adapt header to semantic-ui
- layout: adapt footer to semantic-ui
- layout: adapt main.html to semantic-ui
- layout: adapt LVA section to semantic-ui
- layout: adapt news section to semantic-ui

closes #3006

https://user-images.githubusercontent.com/8863238/105725170-eaf22380-5f28-11eb-9508-c8d21b328f88.mov

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/less/cernopendata/site/globals/site.overrides
+++ b/cernopendata/modules/theme/assets/semantic-ui/less/cernopendata/site/globals/site.overrides
@@ -25,21 +25,6 @@
  */
 
 
-// This is a LESS file. It is automatically imported because of the @siteFolder
-// definition in theme.config.
-//
-// Use this file to provide additional CSS rules. For example:
-
-//  footer{
-//     margin-top: 3px;
-//   }
-
-//   html.cover-page{
-//     background-color: @brandColor;
-//   }
-
-//   .content{
-//     .header{
-//         font-size: xx-large !important;
-//       }
-//   }
+a:hover {
+    text-decoration: underline;
+}
--- a/cernopendata/modules/theme/assets/semantic-ui/less/cernopendata/site/globals/site.overrides
+++ b/cernopendata/modules/theme/assets/semantic-ui/less/cernopendata/site/globals/site.overrides
@@ -28,3 +28,7 @@
 a:hover {
     text-decoration: underline;
 }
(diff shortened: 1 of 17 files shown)


audrium:
Very nice work! :rocket:  Just couple of minor observations:
- I noticed that header is quite huge now compared to what we have in production. `77px` compared to `65px`
- Also regarding the header:
    - if I click "About" multiple times fast, it selects the text
    - On mobile view "About" section is hidden
    - If I open "About" dropdown and select for example CMS on the side (when text is not underlined) it gets rendered in bold and does not redirect:
Screenshot from 2021-01-26 16-24-38
- Search bar lost nice border colour when focused. On Chrome it's even rendered in black for me:
Screenshot from 2021-01-26 15-56-40 (FF renders it in blue)
- Also "Start typing..." placeholder is very near the search bar border. Is it possible to add some padding?
- In "Get Started" section I noticed some extra space between `header` and `body`. By removing `margin-bottom: 2em;` from `header` it worked in the same way as in production. This value comes from `invenio` and was placed in `invenio_theme/theme/globals/site.overrides`
- "News & Updates" section looks way better now :rocket: But for me hovering on CMS label makes it very light:
Screenshot from 2021-01-26 16-17-49

audrium (approved):

mvidalgarcia:
@audrium I've addressed all the issues you mentioned. Thanks for the detailed report!

The mobile view is quite broken ATM, as we get many default styles from Invenio-Theme, but probably we can create a different issue to tackle responsiveness,  what do you think?

audrium:
Just retested, everything works nicely! Turbo :1st_place_medal: 
Totally agree that we shouldn't worry to much about the mobile view now and we can address it later with different issue

Issue #3023, opened 2021-01-26
search: render facet subcategories


mvidalgarcia:
After upgrading to the latest `invenio-search-ui`, facet subcategories are not displayed.

Before:
![image](https://user-images.githubusercontent.com/8863238/105857398-8d261000-5fea-11eb-9ce5-06455ef453ef.png)

After:
![image](https://user-images.githubusercontent.com/8863238/105857550-b050bf80-5fea-11eb-851a-5bc309363775.png)

~We might need to override the [`BucketAggregation`](https://github.com/inveniosoftware/invenio-search-ui/blob/44b6122d8f066a92344f1c1a58b4d18e888b690c/invenio_search_ui/assets/semantic-ui/js/invenio_search_ui/components/SearchApp.js#L37) component to print out subcategories properly.~ Child aggregations [are supported by React-SearchKit](https://github.com/inveniosoftware/react-searchkit/blob/b5d0b560f9387fa5f6e248da5a87ac3185da624b/src/lib/components/BucketAggregation/BucketAggregationValues.js#L34).

PR #3026, opened 2021-01-28
search-ui: support multiple aggregation types

### LINKED ISSUE #3021: search: range aggregations fails on latest invenio-search-ui
In our PROD config, we have a facet with [range aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-range-aggregation.html) (apart from typical [term aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html)):

https://github.com/cernopendata/opendata.cern.ch/blob/101c8942b621c69a7d57e85f5bf28314c50c0a86/cernopendata/config.py#L318-L353

After upgrading to the latest `invenio-search-ui`, which uses `react-searchkit` when using the `semantic-ui` theme, this term aggregation breaks the search as [it's not taken into account](https://github.com/inveniosoftware/invenio-search-ui/blob/44b6122d8f066a92344f1c1a58b4d18e888b690c/invenio_search_ui/views.py#L207) when generating the Search App configuration. Note how it's always taking `terms` as key, thus producing an exception when a range agg is parsed.

In order to fix this, we need to contribute upstream to make it more flexible and accept multiple [bucket aggregations](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket.html).

[_Original discussion here._](https://github.com/cernopendata/opendata.cern.ch/pull/3015#discussion_r562573689)

mvidalgarcia:
- needed in particular for event number filter of type range

closes #3021

### DIFF
--- a/cernopendata/modules/search_ui/helpers.py
+++ b/cernopendata/modules/search_ui/helpers.py
@@ -1,0 +1,60 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of CERN Open Data Portal.
+# Copyright (C) 2021 CERN.
+#
+# CERN Open Data Portal is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# CERN Open Data Portal is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with Invenio; if not, write to the Free Software Foundation, Inc.,
+# 59 Temple Place, Suite 330, Boston, MA 02111-1307, USA.
+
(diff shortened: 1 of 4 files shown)


cernopendata/modules/search_ui/helpers.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        search_index = self._rest_config["search_index"]
+        aggs = (
+            current_app.config.get("RECORDS_REST_FACETS", {})
+            .get(search_index, {})
+            .get("aggs", {})
+        )
+        aggs_list = []
+        for k, v in aggs.items():
+            agg_type = [k for k in v.keys() if k != "aggs"][0]
```
diegodelemos, line 49: Minor: I would perhaps give more descriptive names to the variables like `agg_name` instead of `k`, `agg_definition` instead of `v`, otherwise it gets a bit confusing.

Regarding the implementation, I believe it will be more explicit and understandable when we will extend this to support sub-aggregations (https://github.com/cernopendata/opendata.cern.ch/issues/3023).

diegodelemos (approved):
Works nicely 🚀 (needs rebase)

PR #3027, opened 2021-01-28
frontend: fix markdown rendered pages

### LINKED ISSUE #3007: frontend: migrate markdown renderer to Semantic UI

diegodelemos:
* Fixes horizontal scrolling for code blocks.

* Removes code background color coming in new versions of Python
  markdown package.

Used `find ./ -name "*.md" | xargs grep -e "</.*>" ` to find markdown files using HTML

Closes #3007

### DIFF
--- a/cernopendata/modules/fixtures/data/docs/cms-simulated-dataset-names/cms-simulated-dataset-names.md
+++ b/cernopendata/modules/fixtures/data/docs/cms-simulated-dataset-names/cms-simulated-dataset-names.md
@@ -40,7 +40,7 @@ To unify this part CMS uses the following conventions:
 
 The following conventions are applied for the most common particles:
 
-<table class="table table-condensed">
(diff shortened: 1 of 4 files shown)


mvidalgarcia (approved):
I assume that all anchor tags working as titles should be addressed ad-hoc, is that correct?

I noticed all of them have `name` attribute (not sure why), so with CSS we can do something like:

```css
a[name] {
    color: inherit;
}
```

and forget about them.

PR #3033, opened 2021-01-28
search: add analyzer for title search

### LINKED ISSUE #2930: search: exact dataset title
Revise tokenisers and analysers so that exact searches would work.

For example, for the dataset http://opendata.cern.ch/record/1, the following query does not give anything:

```
/BTau/Run2010B-Apr21ReReco-v1/AOD
```

even though this is exact match with record title.

It even leads to *Error: Invalid query", most probably due to the use of slashes.

The users have to use double quotes:

```
"/BTau/Run2010B-Apr21ReReco-v1/AOD"
```

which finds the record.

However, forcing the use of double quotes is not very user-friendly.

Let's see if we can amend Elasticsearch mappings and analysers so that exact matches will be returned directly.

ParthS007:
addresses #2930

### DIFF
--- a/cernopendata/mappings/v7/records/record-v1.0.0.json
+++ b/cernopendata/mappings/v7/records/record-v1.0.0.json
@@ -1,12 +1,36 @@
 {
-  "mappings": {
-    "properties": {
-      "doi": {
-        "type": "keyword"
-      },
-      "title": {
-        "type": "text"
+    "mappings": {
+      "properties": {
+        "title": {
+            "type": "text",
+            "analyzer": "title_analyzer",
+            "fields": {
+            "exact": {
+                "type": "keyword"
+            }
+            }
+        }
       }
+    },
+    "settings": {
+        "analysis": {
+          "analyzer": {
+            "title_analyzer": {
+              "tokenizer": "keyword",
+              "filter": [
+                "title_filter"
+              ]
+            }
+          },
+          "filter": {
+            "title_filter": {
+              "type": "pattern_capture",
+              "preserve_original": true,
+              "patterns": [
+                "(/[a-zA-Z-_0-9]+)"
+              ]
+            }
+          }
+        }
     }
-  }
 }


crazyjay0516 (approved):

ParthS007:
1. Tokens generated for: `/BTau/Run2010B-Apr21ReReco-v1/AOD`

- /AOD
- /BTau
- /BTau/Run2010B-Apr21ReReco-v1/AOD
- /Run2010B-Apr21ReReco-v1

2. Query for `JetMETTauMonitor`: `/api/records/?q=%22JetMETTauMonitor%22` gives 3 hits:

- recid: 79
- recid: 6
- Documentation: "cms-guide-for-research"

cernopendata/mappings/v7/records/record-v1.0.0.json
```diff
@@ -1,12 +1,39 @@
...
+                "title_filter"
+              ]
+            }
+          },
+          "filter": {
+            "title_filter": {
+              "type": "pattern_capture",
+              "preserve_original": true,
+              "patterns": [
+                "(/[a-zA-Z-_0-9]+)"
```
tiborsimko, line 33: We cannot "close" the referenced issue, since this PR does not touch forward slashes. We can use "addresses" though.

BTW note that we have slashes in other fields, e.g. description. They would need the same treatment, so more fields would be treated the same WRT forward slashes.

But I'm not really sure what practical difference this PR makes against production. They seem to give usually the same results, which is good... But what search exactly did you compare against production system that this pattern capture addresses?

For example, I noticed a possible problem for DOIs. The DOI values also contain slashes. It is interesting to compare a search like `/api/records/?q=10.7483` which gives thousands of records on PROD but only dozens with this PR. I think we have about 2,000 DOIs minted, so we should find more here?  Note the dot...

ParthS007: 1. Yes, `addresses` is more apt here. 👍 

2. I have to add the other fields in the mappings with this type of analyzer, searching of `/` without quotes will work only when we are done with query parsing on python side before sending it to ES.

3. In progress...

4. `doi` was not present in the  previous mapping and were mapped by default settings in ES. Currently I have set it to `keyword` so exact search will work similar to what is their in zenodo search.

tiborsimko: WRT 4, the worry is that 10.7483 will find much less records than it should... The default mapping seems to work better there. Ideally, searching for `10.7483` should find all records that have DOI minted (of the form `10.7483/OPENDATA.CMS.HWS8.QPHJ` and the like).  The default mapping does it better, so I guess we need to address this as part of this PR.

BTW looking for exact DOI is working locally for me (does not lead to query error), but returning too much stuff:

doi-search

tiborsimko (approved):
Thanks, works well locally

PR #3040, opened 2021-02-03
global: remove usage of Invenio-Previewer

### LINKED ISSUE #3011: frontend: migrate file preview to Semantic UI
Move embedded file preview to Semantic UI. Related template [here](https://github.com/cernopendata/opendata.cern.ch/blob/5e76c4b5c6aee057ec0786e1c133352926b68bce/cernopendata/templates/cernopendata_records_ui/records/components/previewer.html).

diegodelemos:
* I doesn't remove Invenio-Previewer from `setup.py` as it is
  required by base Invenio packages.
* Closes #3011


mvidalgarcia:
May I ask what is the reasoning for the deletions of this PR? Could you share some record examples of embedded file previewer?  For instance, http://opendata.cern.ch/record/453/preview/cms-author-list-2012.pdf won't be available after this PR? 
The code looks code but I'm not sure I understand why is being removed.

diegodelemos:
Sorry for the lack of context, we [discussed it internally](https://mattermost.web.cern.ch/it-dep/pl/dcwj3iquh7rufqsjns9tzmeath) and then the PR went stale for a week, I should have written it here instead. We didn't think of this case, so the PR can be closed. Thanks for spotting it.

mvidalgarcia (approved):

diegodelemos:
I have to recreate the PR since I can't update this one (there seems to be a bug, I've updated [diegodelemos:3011-file-preview-to-semantic](https://github.com/diegodelemos/opendata.cern.ch/tree/3011-file-preview-to-semantic) a couple of times and this didn't get updated.)

PR #3041, opened 2021-02-04
ui: small improvements on search page styling

### LINKED ISSUE #3029: ui: improve styling of search pages
Current status after moving to Semantic UI:

search-styling

Possible improvements:

- The 7621 results found is styled in a blue box, which mixes with Dataset blue labels indicating collection type. Let's simply print 7621 without any decoration.

- The title of the result seems to use a too prominent font, it seems to be quite bigger then font size of other pages.

- The title of the facet is also quite huge, could be made smaller

audrium:
closes #3029

Add small styling improvements to search page described [here](https://github.com/cernopendata/opendata.cern.ch/issues/3029#issue-795964768).

Edit:
- In addition removed underline hover effect on search items and dropdown list

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/js/search/components/DefaultListItem.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/search/components/DefaultListItem.js
@@ -33,7 +33,7 @@ import { stripHtml } from "../utils";
 const CODefaultListItem = ({ result: { metadata, id }, index }) => {
   const link = `/record/${id}`;
   return (
-    <Item key={index}>
+    <Item key={index} className="results-item">
       <Item.Content>
         <Item.Header href={link}>{metadata.title}</Item.Header>
         <Item.Description href={link}>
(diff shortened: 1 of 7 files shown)


diegodelemos (approved):

cernopendata/modules/theme/assets/semantic-ui/js/search/components/DefaultListItem.js
```diff
@@ -33,7 +33,7 @@ import { stripHtml } from "../utils";
 const CODefaultListItem = ({ result: { metadata, id }, index }) => {
   const link = `/record/${id}`;
   return (
-    <Item key={index}>
+    <Item key={index} className="results-item">
```
diegodelemos, line 36: I have a doubt, it might be something I don't know. I can't see this class (`results-item`) in the output HTML, however, the changes are applied (there is no `text-decoration`). Could it be that the change is applied by the other CSS changes you did? Either that or I am missing something.

audrium: Hm, the class should be there. Here is what my browser shows me:
Screenshot from 2021-02-05 14-55-49

diegodelemos: I can see it in an incognito window, thanks for double-checking!

Issue #3030, opened 2021-01-28
UI: Support old URLs recognition


ParthS007:
What is happening?
- All records are listed when we search by this URL - `http://localhost:5000/search?keywords=education`.
- When querying in the search bar, we are getting expected results.

What is expected?
- Searching by entering direct URL and querying in the search bar should give same number of results.

Example:
- new: `localhost:5000/search?q=&f=type%3ADataset`
- old: `localhost:5000/search?type=Dataset`

tiborsimko:
Example from the home page: the Glossary link points to: `/search?type=Glossary&sort=title`, while with the new search UI we should rather point to: `?q=&f=type%3AGlossary` (plus sorting by title that is currently broken; see another issue).

While we could change the links internally, there may be many in CMS web site, CERN web site, etc.

Hence it would be good to make a URL redirector to recognise old URLs arguments (such as `type=Glossary`, `keywords=education`, etc) and redirect to the new URLs to ensure they are working.

PR #3044, opened 2021-02-05
ui: use Semantic UI labels instead of custom badge classes

### LINKED ISSUE #3024: frontend: use Semantic UI labels instead of custom badge classes
Use [semantic-ui label classes](https://semantic-ui.com/elements/label.html#colored) that handle what we currently implement by ourselves with [custom badges](https://github.com/cernopendata/opendata.cern.ch/blob/50e78727219347732bdfb9819fc9e8ee27970207/cernopendata/modules/theme/assets/semantic-ui/scss/styles.scss#L398-L465) so we have less code to maintain. For example, we could use `ui label blue` which would have a similar appearance as the current [`badge-subtype`](https://github.com/cernopendata/opendata.cern.ch/blob/50e78727219347732bdfb9819fc9e8ee27970207/cernopendata/modules/theme/assets/semantic-ui/scss/styles.scss#L409).

_Stems from https://github.com/cernopendata/opendata.cern.ch/pull/3017#discussion_r565363208_

audrium:
closes #3024

Removed some of the custom badge CSS classes and started to use Semantic UI labels instead.
Renamed other CSS classes to reflect colours and make it compatible with SUI.

To test please check:
- search results
- /glossary/AOD <- terms example
- /docs/about-cms <- docs example
- /record/22 <- records example


cernopendata/modules/theme/assets/semantic-ui/js/search/components/DefaultListItem.js
```diff
@@ -54,15 +54,15 @@ const CODefaultListItem = ({ result: { metadata, id }, index }) => {
             metadata.type.secondary.map((subtype) => (
               <Label
                 key={subtype}
-                className="badge badge-subtype"
+                className="tiny teal"
                 href={`/search?f=subtype:${subtype}`}
               >
                 {subtype}
               </Label>
```
mvidalgarcia, line 61: I wonder whether we should extract all these labels to a separate component and thus avoid all the repetition we have, what do you think?
So we end up with something like `<ListItemLabels metadata={metadata} />`

I know it's not part of this task, just speaking my mind here.

cernopendata/modules/theme/assets/semantic-ui/scss/styles.scss
```diff
@@ -409,48 +402,11 @@ a {
...
-
-        &:hover {
-            background-color: darken($medium-blue, $darken-on-hover);
-            color: white;
-        }
-    }
 
-    &.badge-category {
+    &.dark-cyan {
         background-color: #0f6d7c;
```
mvidalgarcia, line 410: Minor: what do you think about moving these colors to [`variables.scss`](https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/modules/theme/assets/semantic-ui/scss/variables.scss) for the sake of reusability?

mvidalgarcia (approved):
It looks good to me! 💅 Feels good to get rid of custom code. Left a minor comment and one reflection that doesn't need to be addressed in this PR.

Plus, I think you can delete [`index.html`](https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/templates/cernopendata_pages/index.html), as you've already mentioned on MM, it seems not to be used.

cernopendata/modules/theme/assets/semantic-ui/js/search/components/DefaultListItem.js
```diff
@@ -54,15 +54,15 @@ const CODefaultListItem = ({ result: { metadata, id }, index }) => {
             metadata.type.secondary.map((subtype) => (
               <Label
                 key={subtype}
-                className="badge badge-subtype"
+                className="tiny teal"
                 href={`/search?f=subtype:${subtype}`}
```
tiborsimko, line 58: Shouldn't we also amend search URLs as part of the PR?

Note that old URLs might be supported as part of #3030, but it would be good if the site itself does not generate old URLs, only new ones.

Example: when testing this PR, the links work but the search UI facet checkboxes are not always ticked, try to click on "Dataset" badge as opposed to "Simulated" badge for example:

simulated

The number of results is well 3437, but the facet checkbox is not selected. (Unlike for "Dataset" badge, which is top level, and where this works.)

mvidalgarcia (approved):
Nice! Looks much better now, we avoid lots of repetition.

tiborsimko (approved):
Thanks for the changes, please rebase against `latest` master

PR #3047, opened 2021-02-05
glossary: ignore terms inside MathJax expressions

### LINKED ISSUE #3043: ui: possible interference between TeX display and glossary link display
There are two issues regarding LaTeX display:

(1) The font seems oversized in my Firefox setup compared to the usual text, see:

latex2

(2) When you reload the page a few times, you get to a situation where the display gets broken, see:

latex

The latter is most probably due to interplay of "tau" that is being displayed both as part of TeX and underlined as part of glossary term link.  (This problem happens also on production server.)  

Let's find out how to better co-exist TeX display and glossary display.

diegodelemos:
* Avoids that strings like `\tau` get rendered as a glossary
  term tooltips, instead of a MathJax expressions.

* Closes #3043.

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/js/glossary/jquery.zglossary.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/glossary/jquery.zglossary.js
@@ -27,13 +27,13 @@ import jquery from "jquery/dist/jquery";
 				var skip = 0;
 
 				// Check the element is a text node
-				if (e.nodeType == 3) {
+				if (e.nodeType == 3 && e.textContent.trim().length) {
 
 					// Case insensistive matching option
 					if (options.ignorecase) {
-						var pos = e.data.toLowerCase().search(new RegExp('\\b'+patfmt.toLowerCase()+'\\b'));
+						var pos = e.data.toLowerCase().search(new RegExp('(?<!\\\\)\\b'+patfmt.toLowerCase()+'\\b'));
 					} else {
-						var pos = e.data.search(new RegExp('\\b'+patfmt+'\\b'));
+						var pos = e.data.search(new RegExp('(?<!\\\\)\\b'+patfmt+'\\b'));
 					}
 
 					// Check if the term is found
--- a/cernopendata/templates/cernopendata_theme/page.html
+++ b/cernopendata/templates/cernopendata_theme/page.html
@@ -56,14 +56,13 @@
   {%- block mathjax_support %}
     <script type="text/x-mathjax-config">
       MathJax.Hub.Config({
-        extensions: ["tex2jax.js"],
         jax: ["input/TeX", "output/HTML-CSS"],
         tex2jax: {
           inlineMath: [ ['$','$'], ["\\(","\\)"] ],
           displayMath: [ ['$$','$$'], ["\\[","\\]"] ],
           processEscapes: true
         },
-        "HTML-CSS": { availableFonts: ["TeX"] }
+        "HTML-CSS": { availableFonts: ["TeX"], matchFontHeight: false }
       });
     </script>
 


cernopendata/modules/theme/assets/semantic-ui/js/glossary/jquery.zglossary.js
```diff
@@ -27,13 +27,13 @@ import jquery from "jquery/dist/jquery";
...
-				if (e.nodeType == 3) {
+				if (e.nodeType == 3 && e.textContent.trim().length) {
 
 					// Case insensistive matching option
 					if (options.ignorecase) {
-						var pos = e.data.toLowerCase().search(new RegExp('\\b'+patfmt.toLowerCase()+'\\b'));
+						var pos = e.data.toLowerCase().search(new RegExp('\\b'+patfmt.toLowerCase()+'\\b(?!^\)'));
 					} else {
-						var pos = e.data.search(new RegExp('\\b'+patfmt+'\\b'));
+						var pos = e.data.search(new RegExp('\\b'+patfmt+'\\b(?!^\)'));
```
tiborsimko, line 36: Helps for the example record cited in the issue, but for another record (see below) I have seen the following with this PR:

charged

for the record http://localhost:5000/record/10107 coming from `cernopendata/modules/fixtures/data/records/opera-detector-events-tau.json`.

Can you please have a look at refining the regexp some more?

tiborsimko: BTW do you also see this "overemphasised TeX style" when compared to the style of the main text?  I'm seeing it also with Chromium, but it is less pronounced there. The letter sizes are more comparable, but seem to end up "bolder" than the main text. (Note how In the middle of reload, the "2nd" part is handled by Mathjax and it looks OK, and then it is only at the end of the reload when it turns "bolder"):

record-10107-chromium

cernopendata/modules/theme/assets/semantic-ui/js/glossary/jquery.zglossary.js
```diff
@@ -27,13 +27,13 @@ import jquery from "jquery/dist/jquery";
 				var skip = 0;
 
 				// Check the element is a text node
-				if (e.nodeType == 3) {
+				if (e.nodeType == 3 && e.textContent.trim().length) {
 
 					// Case insensistive matching option
 					if (options.ignorecase) {
-						var pos = e.data.toLowerCase().search(new RegExp('\\b'+patfmt.toLowerCase()+'\\b'));
+						var pos = e.data.toLowerCase().search(new RegExp('(?<!\\\\)\\b'+patfmt.toLowerCase()+'\\b'));
```
diegodelemos, line 34: AFIAK we have no other option than this ugly [escaping](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping) since we have to use a string literal for the regexp (as a regular expression literal it looks like [`'(?<!\\)\btau\b'`](https://regex101.com/r/R5mDa3/1)).

cernopendata/modules/theme/assets/semantic-ui/js/glossary/jquery.zglossary.js
diegodelemos: With https://github.com/cernopendata/opendata.cern.ch/pull/3047/commits/6bc14993d76c539475a7dfd84bad9ca1800cb1f8 the font doesn't get bigger, but still there is a second font size change... let me know how it looks for you. Otherwise we can merge without this commit and tackle this problem separately since apparently, it has to do with MathJax only.

cernopendata/modules/theme/assets/semantic-ui/js/glossary/jquery.zglossary.js
tiborsimko: Thanks, seems to fix the several tau troubles I have tried :+1: 

The font size jumping is also behaving better. I guess we can accept the "similar size"

tiborsimko: > this ugly escaping

... BTW your sigh reminds me of the quote by Jamie Zawinski (see [history](http://regex.info/blog/2006-09-15/247)): _"Some people, when confronted with a problem, think “I know, I'll use regular expressions.”   Now they have two problems."_ :wink: I agree this is a forced move here...

tiborsimko (approved):

PR #3050, opened 2021-02-08
devops: dev deployment preparation

diegodelemos:


mvidalgarcia (approved):
Just one thing to take into account, if you pass `ANNOUNCEMENT_BANNER_MESSAGE = "<script>alert('hacked')</script>"`, it's executed. It shouldn't be a big deal as this value comes from the server and is set by admins. Just in case we expose this configuration to the UI in the future.

PR #3054, opened 2021-02-09
search: configure query parser for handling special characters

### LINKED ISSUE #2930: search: exact dataset title

### LINKED ISSUE #3042: search: configure query parser internally for handling special characters
Using Luqum -> https://luqum.readthedocs.io/en/latest/about.html

ParthS007:
closes #3042 
closes #2930

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -207,6 +207,8 @@ RECORDS_UI_ENDPOINTS = dict(
 RECORDS_REST_ENDPOINTS['recid']['search_index'] = '_all'
 
 RECORDS_REST_ENDPOINTS['recid'].update({
+    'search_factory_imp': 'cernopendata.modules.records.search.query'
+                          ':cernopendata_search_factory',
     'pid_minter': 'cernopendata_recid_minter',
     'pid_fetcher': 'cernopendata_recid_fetcher',
     'record_class': _Record,
--- a/cernopendata/modules/records/search/query.py
+++ b/cernopendata/modules/records/search/query.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of CERN Open Data Portal.
+# Copyright (C) 2021 CERN.
+#
+# CERN Open Data Portal is free software; you can redistribute it
+# and/or modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation; either version 2 of the
+# License, or (at your option) any later version.
+#
+# CERN Open Data Portal is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with CERN Open Data Portal; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place, Suite 330, Boston,
+# MA 02111-1307, USA.
+#
+# In applying this license, CERN does not
+# waive the privileges and immunities granted to it by virtue of its status
+# as an Intergovernmental Organization or submit itself to any jurisdiction.
+
+"""Cernopendata Query factory for REST API."""
+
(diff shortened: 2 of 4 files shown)


cernopendata/config.py
```diff
@@ -281,7 +281,7 @@
             default_order='asc',
             order=2,
         ),
-        'title': dict(fields=['title.keyword'],
+        'title': dict(fields=['title.exact'],
```
tiborsimko, line 284: Sorting is well fixed :+1: Special characters appear in the front, which is OK for now, but we can also see later whether we could classify them into the alphabet, e.g. `/Btau` amongst `B`s rather than in front of the search results.

cernopendata/modules/records/search/query.py
```diff
@@ -0,0 +1,70 @@
...
+
+    :param self: REST view
+    :param search: Elastic search DSL search instance
+
+    :return: Tuple with search instance and URL arguments
+    """
+    def _default_parser(_query_string=None):
+        """Default parser that uses the Q() from elasticsearch_dsl."""
+        if '"' not in _query_string:
+            _query_string = _query_string.replace("/", "//")
```
tiborsimko, line 45: This fixes the internal error for regexp-like queries, but it also leads to some unexpected findings, such as:

- searching for `10.7483/OPENDATA.9S5F.BY3B` will find `/glossary/Barn` as one of the matching hits.  Do you know why was this matched?  

- searching for `"10.7483/OPENDATA.9S5F.BY3B"` will find hundreds of results which are not related, like `record/349`.

An alternative approach could be to quote expressions... We can see IRL

tiborsimko: Here is a table showing an overview of the above findings:

query | opendata.cern.ch (production) | opendata-dev.cern.ch (master) | PR 3054
----- | ----------------------------- | ----------------------------- | -------
10.7483/OPENDATA.9S5F.BY3B | error | 357 results | 359 results
"10.7483/OPENDATA.9S5F.BY3B" | 1 result | error | 5444 results 
10.7483 | 2123 results | 357 results | 359 results 

Basically the PROD system has the best result, the only thing we want to correct is the "error"  when people don't use quotes.  The PR fixes the error, but finds too much hits for quoted query, and not enough hits for partial query.

ParthS007: Yes, I am currently testing this one.

tiborsimko: Ditto for CMS dataset title search; while the PR fixes the error situations, it also finds way too many "unrelated" results:

query | opendata.cern.ch (production) | opendata-dev.cern.ch (master) | PR 3054
----- | ----------------------------- | ----------------------------- | -------
/BTau/Run2010B-Apr21ReReco-v1/AOD | error | error | 4851 results
"/BTau/Run2010B-Apr21ReReco-v1/AOD" | 2 results | error | 4851 results 
BTau | 5 results | 5 results | 5 results

tests/test_cernopendata_query_parser.py
```diff
@@ -0,0 +1,37 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of CERN Open Data Portal.
+# Copyright (C) 2017 CERN.
```
tiborsimko, line 4: s/2017/2021/g ?

cernopendata/modules/records/search/query.py
ParthS007: Ping @tiborsimko

tiborsimko (approved):

PR #3059, opened 2021-02-10
ui: small improvements on search page styling

### LINKED ISSUE #3045: ui: more style improvements for search page
Following upon #3029, we should also improve the following:

- There are two search boxes (see screenshot); we should keep only one.
- Interestingly, note how the search pattern typed into the upper box got changed in the lower box after clicking on the search icon. This issue will disappear once we introduce only one input box, but we should check how the input query pattern is massaged internally, because itlooks like a part of it was lost in the processing, so to speak. (Perhaps just a display issue.)
- If we keep the upper search box, then we could push up a bit the facets on the lhs so that they start at an even level with 18 results found on the rhs.

search-two-boxes

### LINKED ISSUE #3057: ui: styling of "availability" box
Improve styling on the "availability" box, to have nicer bigger title, see for example record http://localhost:5000/record/13072:

availability-ondemand

audrium:
closes #3045
closes #3057

- increased general font size in the records page to unify it across the application
- fixed pagination bug
- removed additional search box
- pushed facets to the top and aligned to `results found`

Screenshot:
Screenshot from 2021-02-10 16-56-53

- also fixed styling of "availability" box (https://github.com/cernopendata/opendata.cern.ch/issues/3057)
To test, please use:
```
docker-compose exec web cernopendata fixtures records -f /code/cernopendata/modules/fixtures/data/records/cms-simulated-datasets-2012.json --mode insert-or-replace
```
and check `http://localhost:5000/record/13072`

For comparison on the left side is the new style, on the right is what we currently have:
Screenshot from 2021-02-11 12-37-46

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
+++ b/cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
@@ -1,6 +1,6 @@
 .content {
   .header {
-    font-size: 1em !important;
+    font-size: 1.2em !important;
   }
 }
 
@@ -8,4 +8,21 @@
   a {
     text-decoration: none;
   }
+}
+
+.result-options {
+  margin-top: 20px;
+}
+
+.four.wide.column {
+  margin-top: -70px !important;
+}
+
+.center.aligned.eight.wide.column {
+  margin-left: -130px;
+}
+
+.right.aligned.four.wide.column {
+  margin-top: 0px !important;
+  margin-left: 60px;
 }
--- a/cernopendata/templates/cernopendata_records_ui/records/record_detail.html
+++ b/cernopendata/templates/cernopendata_records_ui/records/record_detail.html
@@ -147,15 +147,14 @@
         </div>
 
         {% if record.distribution.get("availability", "online") != "online" %}
-
-        <div class="file-availability-disclaimer">
-            <h6>Availability: <strong style="color: {{ '#a24d4d' if record.distribution.availability else '#af7c2a' }}">{{record.distribution.availability | upper}}</strong>
-            </h6>
+        <div class="ui info message">
+            <div class="header">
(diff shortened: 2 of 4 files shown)


cernopendata/templates/cernopendata_records_ui/records/record_detail.html
```diff
@@ -147,16 +147,15 @@ <h2>Characteristics</h2>
         </div>
 
         {% if record.distribution.get("availability", "online") != "online" %}
-
-        <div class="file-availability-disclaimer">
-            <h6>Availability: <strong style="color: {{ '#a24d4d' if record.distribution.availability else '#af7c2a' }}">{{record.distribution.availability | upper}}</strong>
-            </h6>
+        <div class="ui info message">
```
diegodelemos, line 150: @tiborsimko is this new color ok?

cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
```diff
@@ -9,3 +9,20 @@
     text-decoration: none;
   }
 }
+
+.result-options {
+  margin-top: 20px;
+}
+
+.four.wide.column {
```
diegodelemos, line 17: Works nicely. I guess this will become a problem when we try to make the web responsive. I think we should open an issue on `inveniosoftware/React-SearchKit` to solve it upstream, what do you think?

diegodelemos (commented):
LGTM @tiborsimko if the new "ONDEMAND" box is fine we can merge.

tiborsimko (approved):

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
```diff
@@ -147,16 +147,15 @@ <h2>Characteristics</h2>
...
+            <div class="header">
+                Availability: <strong style="color: {{ '#a24d4d' if record.distribution.availability else '#af7c2a' }}">{{record.distribution.availability | upper}}</strong>
```
mvidalgarcia, line 152: 🤦‍♂️

PR #3060, opened 2021-02-12
search: support old search querystring format

### LINKED ISSUE #3030: UI: Support old URLs recognition

mvidalgarcia:
closes #3030

🆕 I accidentally found a `search_wrapper` function that does what we want so there's no need to have changes on Invenio-Search-UI.

#### Steps to test:

- All links following the old format e.g. `/search?type=Dataset` should be translated into the new one (`/search?f=type:Dataset`), thus making the search properly work.
- All the links displayed on the home page follow this old format, so clicking on them should perform the proper search.
- Play with different query params, you can get inspiration on the test included in this PR (`test_old_search_qs`).


cernopendata/views.py
```diff
@@ -44,6 +44,15 @@ def redefine_search_endpoint(blueprint_setup):
 
 def search_wrapper():
     """Wrap default invenio search endpoint."""
+    # translate old search query params to new format
+    # e.g. type=Dataset => f=type:Dataset
+    facets = current_app.config['RECORDS_REST_FACETS']
+    facet_keys = facets['_all']['aggs'].keys()
```
audrium, line 50: I think storing `facet_keys` in memory on app load could speed things up and save us from parsing  `RECORDS_REST_FACETS` to get the keys on each request. What do you think?

mvidalgarcia: Where do you suggest storing them?

cernopendata/views.py
```diff
@@ -44,6 +44,15 @@ def redefine_search_endpoint(blueprint_setup):
...
+    args = request.args.to_dict()
```
audrium, line 51: I noticed that queries where multiple types/subtypes are defined didn't work as expected. Example:
```/search?page=1&size=20&type=Dataset&type=Documentation&subtype=Collision&subtype=About```

Expected: `Dataset/Collision` and `Documentation/About` to be selected, but instead it selects only `Dataset/Collision`.

This happens because `request.args.to_dict()` overrides values for the same keys. To fix it `request.args.to_dict(flat=False)` could be used which returns:
```
{ ...
'subtype': ['Collision', 'About'],
'type': ['Dataset', 'Documentation']
... }
```

File: cernopendata/views.py (hunk 1 above)
Comment (audrium) [reply cernopendata/views.py:50]: This can be ignored, since the facets tree is also needed in `translate_search_url`

**Review by audrium — APPROVED** [review]
Tested carefully and if works great :rocket: Just spotted one minor issue which can be addressed

File: cernopendata/views.py (hunk 2 above)
Comment (mvidalgarcia) [reply cernopendata/views.py]: Well spotted! I hadn't noticed this use case. I modify a little bit the logic to take this into account and added a test case covering it. Thanks!

Comment (audrium) [reply cernopendata/views.py]: Retested and works great now. But I found another small issue. Once all `subtypes` of a certain `type` are selected, it does not work correctly. Example:
```
/search?page=1&size=20&q=&type=Dataset&subtype=Collision&subtype=Derived&subtype=Simulated
```
It should select `Collision, Derived, Simulated`, but selects only `Collision`

Comment (mvidalgarcia) [reply cernopendata/views.py]: This is getting trickier... when you have multiple subtypes it's hard to know who's the parent type as they are part of the same array...

For instance, if you have 2 types and 4 subtypes, you'd get something like `{'type': ['A', 'B'], 'subtype': ['a1', 'a2', 'b1', 'b2']}`, how do you know that `a1` belongs to `A`? From the code we can't access the different values an agg has, i.e. I can't know that `Collision` is a subtype of `Dataset`...  

With my last change, I was relying on the order of the array but of course, I wasn't taking into account that the might be more than one subtype selected for the same type.

I honestly don't know how to solve this. Do we have to support all of these use cases? How was the old code knowing the subtype-type relationships? @tiborsimko @diegodelemos

Comment (tiborsimko) [reply cernopendata/views.py]: Since our facet hierarchy is not changing often, actually hardly ever at all, and since we have all the fixtures available in our code base, we could take the following "outside path" to solve the problem.

(1) Write a helper script called say `scripts/generate_facet_hierarchy.py` that would go through all the JSON fixtures, look for type/subtype JSON fields such as:

```json
    "type": {
      "primary": "Dataset",
      "secondary": [
        "Collision"
      ]
    },
```

and would output a Python configuration variable indicated all primary/secondary type pairs, for example:

```python
FACET_HIERARCHY = {
    'Dataset':  ['Collision', 'Derived', 'Simulated'], 
    'News': [], 
    ....
}
```

Please see existing helper scripts such as `scripts/check_fixtures.py` or `scripts/clean_json_file.py`.

(2) We'll commit manually the above `FACET_HIERARCHY` variable into `cernopendata/config.py`.  We could actually even write a test case for `run-tests.sh` that would check whether this variable isn't outdated.  So that CI could alert us to update it should something change in the fixtures.  (This is somewhat similar to REANA's generation of OpenAPI specs.)

(3) The `FACET_HIERARCHY` could then be used by the URL type/subtype manipulator in order to generate proper `f=` arguments out of the passed subtype values.

P.S. The above `FACET_HIERARCHY` data structure was mentioned just for illustration; it may be good to store pairs for more runtime lookup efficiency, e.g. in the form `subtype: type`.

tiborsimko: P.P.S. The above should also be more nicely named, for example `FACET_TYPE_HIERARCHY`, because this concerns only `type`, not the other facets...

mvidalgarcia: I've modified the logic creating a config variable, could you please check again?

audrium: I've noticed that there is an issue now when selecting couple of items which does not have any subtypes. Example:
- `/search?page=1&size=20&q=&experiment=ALICE&experiment=ATLAS`
  should select ALICE and ATLAS, but selects only ALICE

- `/search?page=1&size=20&q=&type=News&type=Glossary`
  should select News and Glossary, but selects only News

cernopendata/views.py
```diff
@@ -54,6 +65,48 @@ def search_wrapper():
         return invenio_search_view()
 
 
+def translate_search_url(args, facets):
+    """Translate old search querystring args to new ones."""
+    subagg_agg_mapping = {}
```
audrium, line 70: What do you think of breaking down this function into smaller pieces? It got pretty heavy and it's quite hard to follow all the logic, maybe some comments would be beneficial as well. In case someone has to debug it in the future

cernopendata/views.py
mvidalgarcia: Wow good catch, it was a leftover from changing the `flat=False`, it should be better now. I also added lots of comments to understand the algorithm as it got complex... Thank you so much for checking and sorry for all the mess, it seemed like a simple task in the beginning but go supercomplex in the end...

audrium: I've just retested, works nicely! Thanks for the fixes, it's much easier to follow the logic now with the comments

PR #3061, opened 2021-02-12
histograms: load first plot by default

### LINKED ISSUE #3052: histogramming: load first plot by default
(1) One issue in the new histogramming interface is the following. When users enters the `/visualise/histograms`  page for the first time, e.g. check fresh browser, they will see the following:

h1

It does not feel convenient not to see any plot at this stage. In spite of seeing the help text, people would have to think what to do.

For this reason, the PROD was pre-selecting E1, so that "at least one example plot" would be shown to newcomers.

(1a) Can we pre-select E1 (or some other reasonable entity), so that people would see a plot when entering the page?
 
(1b) Note that the behaviour from PROD was not ideal either, e.g. if you change the event types from the menu, the plot may not be shown... But at least it was upon first page view.

(2) BTW a minor stylistic issue observed on the side: (Qutebrowser, Firefox)

h2

h3

see how "Log Y" is hidden because "Set bin width" takes more horizontal and vertical space.

audrium:
closes #3052

This fix loads first plot by default once a new event type is selected and on initial page load. It solves the race condition described [here](https://github.com/cernopendata/opendata.cern.ch/issues/3052#issuecomment-777310981).

Also, some styling improvements are done. Here is how it looks like now:
Screenshot_2021-02-12 CERN Open Data Portal

To test please see the instructions [here](https://github.com/cernopendata/opendata.cern.ch/pull/3032)

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/scss/visualise.scss
+++ b/cernopendata/modules/theme/assets/semantic-ui/scss/visualise.scss
@@ -1,49 +1,71 @@
-#vis-histograms {
-  color: #606D75;
-}
+$histograms-dark-grey: #6c757d;
 
-#vis-histograms .previewer {
-  background-color: #f4f4f4;
-  border-radius: 4px;
-  margin-bottom: 35px;
-  color: #606D75;
-  margin-bottom: 20px;
-}
+.vis-histograms {
+  color: $histograms-dark-grey;
 
-#vis-histograms #flot-plots {
-  background-color: #fff;
-  margin-bottom: 15px;
-  min-height: 300px;
-  height: auto;
-  margin-right: 0;
-  margin-left: 0;
-}
+  .previewer {
+    border-radius: 4px;
+    color: $histograms-dark-grey;
+    margin-bottom: 20px;
+  }
 
-#vis-histograms #howto > div{
-  background-color: #f4f4f4;
-  border-radius: 4px;
-  pre {
-    overflow-y: hidden;
-    overflow-x: auto;
+  #flot-plots {
+    background-color: white;
+    margin-bottom: 15px;
+    min-height: 300px;
+    height: auto;
+    margin-right: 0;
+    margin-left: 0;
+  }
+
+  #howto > div {
+    border-radius: 4px;
+    pre {
+      overflow-y: hidden;
+      overflow-x: auto;
+    }
+  }
+
+  .plot-control {
+    margin-left: 2em !important;
+  }
+
+  .ui.button {
+    padding: 0.8em;
+    margin: 0;
+    font-weight: 500;
+
+    &.toggle.active {
+      background-color: $histograms-dark-grey !important;
+    }
+
+    &.undoselect {
+      margin-left: 4em;
+    }
+  }
+
+  .ui.action.input {
+    margin-left: 1em;
+    max-width: 150px;
+  }
+
+  .ui.grid {
+    margin-top: 0rem;
+    .row {
+      padding-left: 1rem;
+    }
+  }
+
+  .ui.selection.dropdown {
+    background-color: $histograms-dark-grey;
+    .text {
+      color: white !important;
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


cernopendata/modules/theme/assets/semantic-ui/scss/visualise.scss
```diff
@@ -28,6 +28,29 @@
...
+  margin: 0;
+  font-weight: 500;
+}
+
+#vis-histograms .ui.action.input {
+  margin-left: 1em;
+  max-width: 150px;
+}
+
+#vis-histograms .undoselect {
```
mvidalgarcia, line 50: You can move this up under `#vis-histograms .ui-button` and thus you avoid the `!important`:

```css
#vis-histograms .ui.button {
  padding: 0.8em;
  margin: 0;
  font-weight: 500;

  &.undoselect {
    margin-left: 4em;
  }
}
```

cernopendata/modules/theme/assets/semantic-ui/scss/visualise.scss
```diff
@@ -28,6 +28,29 @@
   }
 }
 
+#parameter-button-row .item {
+  margin-left: 0.5em;
+}
+
+#vis-histograms .plot-control {
```
mvidalgarcia, line 35: Since you're modifying this file, what do you think about arranging it a little bit better? Developers that touch this file in the future will be grateful :)

- We can use SASS nesting capabilities to move things under a single `#vis-histograms`
- Colors can be moved to variables file to avoid _magic numbers_
- It's [not recommended to use `id`s](https://dev.to/clairecodes/reasons-not-to-use-ids-in-css-4ni4) unless it's completely necessary, they can be converted to classes. If you do this, remember to change JS selectors as well.

mvidalgarcia (approved):
It works properly 👍 I'll leave some notes on how to improve a bit the CSS file.

cernopendata/modules/theme/assets/semantic-ui/scss/visualise.scss
audrium: Thanks for the tips, I've addressed most of the items, however still left some id`s which were queried in jQuery quite frequently

cernopendata/modules/theme/assets/semantic-ui/scss/visualise.scss
```diff
@@ -1,49 +1,74 @@
-#vis-histograms {
-  color: #606D75;
-}
+$histograms-text: #606d75;
+$histograms-dropdown: #6c757d;
+$histograms-button: #6c757d;
```
mvidalgarcia, line 3: I'd name these variables to describe the color itself as we have in `variables.scss`. Unless one is able to read HEX, they won't tell you much.

PS. `$histograms-dropdown` and `$histograms-button` have the same value.

cernopendata/modules/theme/assets/semantic-ui/scss/visualise.scss
```diff
@@ -1,49 +1,74 @@
...
-  min-height: 300px;
-  height: auto;
-  margin-right: 0;
-  margin-left: 0;
-}
+  .previewer {
+    border-radius: 4px;
+    margin-bottom: 35px;
+    color: $histograms-text;
+    margin-bottom: 20px;
```
mvidalgarcia, line 12: `margin-bottom` twice.

cernopendata/modules/theme/assets/semantic-ui/scss/visualise.scss
audrium: Since all these colors were different variants of grey I took functional naming approach, but after playing a bit more with these I've reduced the colors to s single grey color, because the differences were very minimal

cernopendata/modules/theme/assets/semantic-ui/scss/visualise.scss
audrium: Nice catch, leftover from the old code

tiborsimko (approved):

PR #3071, opened 2021-02-16
ui: fix CMS visualiser documentation arrangement

### LINKED ISSUE #3069: previewer: fix CMS documentation box-like arrangement
Go to http://opendata.cern.ch/visualise/events/cms and select "Need HELP?" link.

In production, you get seven screenshots, aligned in 3x2 plus an extra row:

c1

In development, you get 4x4, then 0-1, then 2-0 and 2-0. 

c2

Can we improve the style so that the page reads a bit better?

audrium:
closes #3069

- Removed some css leftovers from global styles and created a separate file for custom ispy css

Result:
Screenshot from 2021-02-16 11-03-34

### DIFF
--- a/cernopendata/modules/theme/static/assets/ispy/ispy.css
+++ b/cernopendata/modules/theme/static/assets/ispy/ispy.css
@@ -1,0 +1,12 @@
+.ispy-help-img {
+    width: 100%;
+    height: auto;
+}
+
+.ispy-help-item {
+    padding: 5px !important;
+}
+
+.ispy-help-row {
+    margin: 0 !important;
+}
--- a/cernopendata/modules/theme/static/assets/ispy/ispy.css
+++ b/cernopendata/modules/theme/static/assets/ispy/ispy.css
@@ -3,10 +3,10 @@
     height: auto;
 }
 
-.ispy-help-item {
-    padding: 5px !important;
+.bootstrap-ispy .ispy-help-item {
+    padding: 5px;
 }
 
-.ispy-help-row {
-    margin: 0 !important;
+.bootstrap-ispy .ispy-help-row {
+    margin: 0;
 }
--- a/cernopendata/templates/cernopendata_ispy/ispy_help.html
+++ b/cernopendata/templates/cernopendata_ispy/ispy_help.html
@@ -34,7 +34,7 @@
             </div>
             <div class="col-sm-6 ispy-help-item">
               <i class="fa fa-home"></i> Returns the event display to its initial orientation. <i
-              class="fa fa-search-plus"></i> and <i class="fa fa-search-minus"></i> Zoom in and out.<br/><img
+              class="fa fa-search-plus"></i>/<i class="fa fa-search-minus"></i> Zoom in/out.<br/><img
               class="ispy-help-img" src="{{ url_for('static', filename='img/iSpy-Online-graphics/start.gif')}}">
             </div>
             <div class="col-sm-6 ispy-help-item">
--- a/cernopendata/templates/cernopendata_ispy/ispy_help.html
+++ b/cernopendata/templates/cernopendata_ispy/ispy_help.html
@@ -22,41 +22,41 @@
             Each button comes with a tool-tip: hover over the button to know its functionality.
           </p>
           <p>
-          <div class="row">
-            <div class="col-sm-6">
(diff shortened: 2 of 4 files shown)


tiborsimko:
Hmm, I still seem to see the unaligned rows on a freshly built system?!

cernopendata/modules/theme/static/assets/ispy/ispy.css
```diff
@@ -0,0 +1,12 @@
...
+    height: auto;
+}
+
+.ispy-help-item {
+    padding: 5px !important;
+}
+
+.ispy-help-row {
+    margin: 0 !important;
+}
```
mvidalgarcia, line 12: [Using `!important` is a bad practice](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity#the_!important_exception) as it breaks the natural cascading and makes debugging harder.

This can be avoided with something like:

```css
.bootstrap-ispy .ispy-help-item {
    padding: 5px;
}

.bootstrap-ispy .ispy-help-row {
    margin: 0;
}
```

Tested and working properly.

mvidalgarcia (approved):
Working properly on my side! ✨ Just left a minor comment with some advice.

image

cernopendata/modules/theme/static/assets/ispy/ispy.css
tiborsimko: Interesting! I had just rebuilt afresh this PR to confirm/infirm the problem, and I do get the same non-arranged setup described in the issue. Seeing this with Firefox, Chromium on Fedora 33. What could be the reason for the rendering difference on our two platfroms? Perhaps the font setup or some such? BTW, do you see the original problem on `master` branch e.g. if you check [on DEV](https://opendata-dev.cern.ch/visualise/events/cms)?

mvidalgarcia: Yes, I see the problem on DEV. FF on macOS 11.2 Big Sur.

tiborsimko: OK, here is the problem. Using this PR, I'm seeing the following:

out

Note how "out." is displayed on the next line. This is not what you see in the screenshot, and this is exactly what makes the arrangement to go wild. 

Using browser inspector, I shortened the text, and the boxes became well aligned. So the workaround solution could be to simply shorten the help text and write "Zoom in/out", as follows:

out3

This fixes the arrangement of the boxes on my setup as well.

Perhaps you can add this improvement to the PR alongside your `!important` changes?

(Alternatively, we could find out why the font size is not "the same" on our two platforms, but it does not seem worth it...)

mvidalgarcia: I see.. tricky indeed.. this might be related to the problem we had with mathjax and so... In any case, we should have a better grid that prevent these arrangement issues, I guess we can tackle it on a different issue together with the font size problem.

I updated the PR, could you have a look now and merge if it's ok?

cernopendata/templates/cernopendata_ispy/ispy_help.html
```diff
@@ -22,41 +22,41 @@ <h3>Top-row buttons</h3>
...
-              events in the <kbd>.ig</kbd> file you loaded.<br/><img class="screen-s"
+              events in the <kbd>.ig</kbd> file you loaded.<br/><img class="ispy-help-img"
                                                                      src="{{ url_for('static', filename='img/iSpy-Online-graphics/prevnext.gif')}}">
             </div>
-            <div class="col-sm-6">
+            <div class="col-sm-6 ispy-help-item">
               <i class="fa fa-home"></i> Returns the event display to its initial orientation. <i
-              class="fa fa-search-plus"></i> and <i class="fa fa-search-minus"></i> Zoom in and out.<br/><img
-              class="screen-s" src="{{ url_for('static', filename='img/iSpy-Online-graphics/start.gif')}}">
+              class="fa fa-search-plus"></i> and <i class="fa fa-search-minus"></i> Zoom in/out.<br/><img
```
tiborsimko, line 37: Please replace also the first "and" with "/", so that plus/minus icons would have a slash between themselves, analogously to the later "in/out" string.

mvidalgarcia: Sorry I didn't notice this in your message. Fixed.

tiborsimko (approved):

PR #3073, opened 2021-02-16
ui: improve styling of facets

### LINKED ISSUE #3062: search: improve styling of facets
Current status:

search

Possible improvements:

- The facet name ("Experiment") is graphically too prominent. Facets should just aid the navigation, so not be the main focus point for eyes. 
  Possible improvements: (1) keep facet title in bold, but in same size as facet values, and not separate it by underline. (A bit like Amazon facet style.) Alternatively, (2) print it in lighter regular font, and colour the background. (A bit like Zenodo facet style.) Personally I like the former, since the facet list is long, so it would save also the vertical screen space.
  
- The subfacets ("Collision") are graphically mixing with facets ("Dataset"). This could be improved by having more indendation, so that subfacet checkboxes are aligned with facet value.  A bit like GitHub does, see:

    - [ ] Dataset
      - [ ] Collision
      - [ ] Derived
      - [ ] Simulated
    - [ ] Documentation
      - [ ] About
      - [ ] ...
              
- The facet value followed by count ("Dataset (4,711)") is a blob printed in the same regular font. Consider printing facet value counts in light grey, to better distinguish facet value ("Dataset"), printed in the common black, from facet value ("4711"), printed in light grey. In the PROD, the count was displayed on the right-hand-side, which ensured the separation; this was also chosen for INSPIRE for example. In DEV, we can keep the facet counts on the lhs, provided that we alter their appearance a bit, so that the numbers would be graphically and visually better separated from the facet values.

audrium:
closes #3062

- Reduced facet name font size
- Added more indendation for subfacets
- Colored facet value counts in grey

New look:
Screenshot from 2021-02-16 14-17-04

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/js/search/components/FacetItem.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/search/components/FacetItem.js
@@ -24,13 +24,39 @@
  * as an Intergovernmental Organization or submit itself to any jurisdiction.
  */
 
-import { createSearchAppInit } from "@js/invenio_search_ui";
-import { CODLayoutSwitcher, ResultsCount, CODFacetItem } from "./components";
+import React from "react";
+import { List, Checkbox } from "semantic-ui-react";
 
-const initSearchApp = createSearchAppInit({
-  "LayoutSwitcher.element": CODLayoutSwitcher,
-  "ResultsGrid.container": null,
-  "Count.element": ResultsCount,
-  "SearchApp.searchbarContainer": () => null,
-  "BucketAggregationValues.element": CODFacetItem,
-});
+const CODFacetItem = (props) => {
+  const {
+    bucket,
+    isSelected,
+    onFilterClicked,
+    getChildAggCmps,
+    keyField,
+  } = props;
+  const label = bucket.label ? (
+    bucket.label
+  ) : (
+    <label>
+      {keyField}{" "}
+      <span className="facet-item-count">
+        ({bucket.doc_count.toLocaleString("en-US")})
+      </span>
+    </label>
+  );
+  const childAggCmps = getChildAggCmps(bucket);
+  return (
+    <List.Item key={bucket.key}>
+      <Checkbox
+        label={label}
+        value={bucket.key}
+        onClick={() => onFilterClicked(bucket.key)}
+        checked={isSelected}
+      />
+      {childAggCmps}
+    </List.Item>
+  );
+};
+
+export default CODFacetItem;
--- a/cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
+++ b/cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
@@ -47,3 +47,9 @@
   margin-top: 0px;
   margin-left: 60px;
 }
+
+.twelve.wide.column {
+  .ui.grid.relaxed {
+    background-color: white;
+  }
+}
--- a/cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
+++ b/cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
@@ -1,7 +1,28 @@
-.content {
-  .header {
-    font-size: 1.2em !important;
+@import "variables";
+
+.ui.items > .item > .content > .header:not(.ui) {
+  font-size: 1.2em;
+}
+
+.ui.card {
+  .content {
+    padding: 10px 15px;
+
+    .header:not(.ui) {
+      font-size: 0.9em;
+      margin: 0;
+    }
   }
+}
+
+.item {
+  .ui.list {
+    padding-left: 1.2em;
+  }
+}
+
+.facet-item-count {
+  color: $dark-grey;
 }
 
 .results-item {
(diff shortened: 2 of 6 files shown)


tiborsimko:
One interesting colour observation.  (1) On the search results page, the lhs facets (supposedly less important) are printed on whitish background, while the rhs search results (supposedly more important) are printed on greyish background. However, when you click on a search hit, and you are brought (2) onto the detailed record page, the record content is printed on the whitish background, while the not important page margins are displayed on the greyish background. This creates a bit inconsistent look and feel, which may disturb user experience.  Can we harmonise this, so that the most important page content would be always printed on the same background, and less important page border areas on another background, consistently? 
(E.g. compare with PROD where the search results were printed on a whitish background too.)

cernopendata/modules/theme/assets/semantic-ui/js/search/components/FacetItem.js
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * In applying this license, CERN does not
+ * waive the privileges and immunities granted to it by virtue of its status
+ * as an Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+import React from "react";
+import { List, Checkbox } from "semantic-ui-react";
+
+const FacetItem = (props) => {
```
mvidalgarcia, line 30: Minor: What do you think about prefixing this component with `COD`? As it's a custom COD one.

cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
```diff
@@ -4,6 +4,27 @@
   }
 }
 
+.ui.card {
+  .content {
+    padding: 10px 15px !important;
+
+    .header {
+      font-size: 0.9em !important;
```
mvidalgarcia, line 12: This value was set already on line 1 of this very same file.

cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
```diff
@@ -4,6 +4,27 @@
   }
 }
 
+.ui.card {
+  .content {
+    padding: 10px 15px !important;
```
mvidalgarcia, line 9: As mentioned [here](https://github.com/cernopendata/opendata.cern.ch/pull/3071/files#r577455237), it'd be nice to get rid of `!important`s.

cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
```diff
@@ -4,6 +4,27 @@
...
+}
+
+.item {
+  .ui.list {
+    padding-left: 1.2em;
+  }
+}
+
+.facet-item-count {
+  color: grey;
```
mvidalgarcia, line 25: We have `$dark-grey` color in `variables.scss`. It might fit well here?

cernopendata/modules/theme/assets/semantic-ui/js/search/components/FacetItem.js
```diff
@@ -0,0 +1,62 @@
...
+    keyField,
+  } = props;
+  const label = bucket.label ? (
+    bucket.label
+  ) : (
+    <label>
+      {keyField}{" "}
+      <span className="facet-item-count">
+        ({bucket.doc_count.toLocaleString("en-US")})
+      </span>
```
mvidalgarcia, line 45: It's a pity that we have to override the whole component only because of separating the count in a different HTML element, but I don't see any other easy solution for this... 😕

mvidalgarcia (commented):
Looks good, left some minor comments.

mvidalgarcia:
A) White background for results (resembling current PROD).

image

image

B) Grey background

image

image

I personally prefer A, it makes the main content stand out.

ParthS007:
I liked the A option. 👍

mvidalgarcia:
Heads up: Beware of the new commit I've created to make the background white.

audrium:
I also like the A) option, however I think by adding a greyish border it looks even nicer and consistent with other UI elements:
Screenshot from 2021-02-18 11-39-12

WDYT?

mvidalgarcia:
Looks good IMO!

cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
audrium: That value on line 1 controls the header font size of search results, while this line sets facets header font size. Definitely not ideal, but since we don't have any classes to attach styles to.. we have this mess

cernopendata/modules/theme/assets/semantic-ui/js/search/components/FacetItem.js
audrium: Indeed, I also spent some time trying to find another way, but with no luck

cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
mvidalgarcia: I see, thanks for explaining, it tricked my mind...

> Definitely not ideal, but since we don't have any classes to attach styles to.. we have this mess

That's so true.. it's a bit of a mess not being able to add custom classes easily. I'll ask about this problem in Invenio chat to see if there's any other way than overriding the whole component.

mvidalgarcia (approved):
LGTM! 👨‍🎨 Just needs rebase.

PR #3075, opened 2021-02-16
records: fix schema.org JSON-LD export

### LINKED ISSUE #3065: records: fix schema.org JSON-LD export
Record HTML pages have embedded schema.org JSON-LD inside, e.g. view source of https://opendata.cern.ch/record/1:

```
<script type='application/ld+json'>
{
  "@context": "https://schema.org/", 
  "@id": "https://doi.org/10.7483/OPENDATA.CMS.A342.9982", 
  "@type": "Dataset", 
  "creator": {
    "@type": "Organization", 
    "name": "CMS collaboration"
  }, 
  "dateCreated": "2010", 
  "datePublished": "2014", 
  "description": "<p>BTau primary dataset in AOD format from RunB of 2010</p> <p>This dataset contains all runs from 2010 RunB. The list of validated runs, which must be applied to all analyses, can be found in</p>", 
  "identifier": "https://doi.org/10.7483/OPENDATA.CMS.A342.9982", 
  "name": "/BTau/Run2010B-Apr21ReReco-v1/AOD", 
  "publisher": {
    "@type": "Organization", 
    "name": "CERN Open Data Portal"
  }, 
  "url": "http://opendata.cern.ch/record/1"
}
</script>
```

On DEV, this is printed in one line, and note the different keys:

```
<script type='application/ld+json'>
{"context_":"https://schema.org/","creator":{"@type":"Organization","name":"CMS collaboration"},"date_created":"2010","date_published":"2014","description":"<p>BTau primary dataset in AOD format from RunB of 2010</p> <p>This dataset contains all runs from 2010 RunB. The list of validated runs, which must be applied to all analyses, can be found in</p>","id_":"https://doi.org/10.7483/OPENDATA.CMS.A342.9982","identifier":"https://doi.org/10.7483/OPENDATA.CMS.A342.9982","name":"/BTau/Run2010B-Apr21ReReco-v1/AOD","publisher":{"@type":"Organization","name":"CERN Open Data Portal"},"type_":"Dataset","url":"http://opendata.cern.ch/record/1"}
</script>
```

We should add newlines for clearness, and check keys e.g. `type_` instead of `@type`, etc.

This is governed by the JSON-LD serialiser, which probably needs amendments as per upgrading the Invenio base packages.

audrium:
closes  #3065

- fixed Marshmallow v2 to v3 migration issues
- added custom JSONSerializer to dump it with indentation

### DIFF
--- a/cernopendata/modules/records/serializers/schemaorg.py
+++ b/cernopendata/modules/records/serializers/schemaorg.py
@@ -23,6 +23,7 @@
 # as an Intergovernmental Organization or submit itself to any jurisdiction.
 """CERN Opendata schema.org JSON-LD serializer."""
 
+import json
 
 from invenio_records_rest.serializers.json import JSONSerializer
 
@@ -30,7 +31,24 @@ from cernopendata.modules.records.serializers.schemas \
     import schemaorg_schemas as schemas
 
 
-class CODSchemaorgSerializer(JSONSerializer):
+class BasicJSONSerializer(JSONSerializer):
+    """Basic JSON serializer."""
+
+    def serialize(self, pid, record, links_factory=None, **kwargs):
+        """Serialize a single record and persistent identifier.
+
+        :param pid: Persistent identifier instance.
+        :param record: Record instance.
+        :param links_factory: Factory function for record links.
+        """
+        return json.dumps(
+            self.transform_record(pid, record, links_factory, **kwargs),
+            indent=2,
+            separators=(", ", ": "),
+        )
+
+
+class CODSchemaorgSerializer(BasicJSONSerializer):
     """CERN Open Data schema.org serializer.
 
     Serializes a Record based on it's type (Dataset, Software, etc.) to
--- a/cernopendata/modules/records/serializers/schemas/schemaorg_schemas.py
+++ b/cernopendata/modules/records/serializers/schemas/schemaorg_schemas.py
@@ -34,29 +34,29 @@ class RecordSchemaorgSchema(Schema):
     """schema.org JSON-LD schema for CERN Opendata Record."""
 
     # JSON-LD minimum requirements
-    context_ = fields.Field(default='https://schema.org/', dump_to='@context')
-    id_ = fields.Method('get_identifier', dump_to='@id')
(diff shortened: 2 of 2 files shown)


cernopendata/modules/records/serializers/schemaorg.py
```diff
@@ -23,14 +23,31 @@
...
+
+        :param pid: Persistent identifier instance.
+        :param record: Record instance.
+        :param links_factory: Factory function for record links.
+        """
+        return json.dumps(
+            self.transform_record(pid, record, links_factory, **kwargs),
+            indent=2,
+            separators=(", ", ": "),
+        )
```
mvidalgarcia, line 48: Do we even need this class? It's basically replicating [this functionality](https://github.com/inveniosoftware/invenio-records-rest/blob/c4a3717afcf9b08b6e42f3529addecc64bb2e47c/invenio_records_rest/serializers/json.py#L29-L32). If one passes `?prettyprint=1` the schema would get indented automatically. I guess for Google robots it doesn't matter that is indented or not and we can always force it. WDYT? cc/ @tiborsimko

mvidalgarcia (approved):
It works nicely! Just one minor observation that could save us some lines of code.

PS. Beware the [pycodestyle error](https://github.com/cernopendata/opendata.cern.ch/pull/3075/checks?check_run_id=1912498791)

cernopendata/modules/records/serializers/schemaorg.py
tiborsimko: It may be useful for "data curators" to see the JSON nicely printed, with new lines and everything, when doing "view page source" in the browsers.  That was the only motivation.

That said, we could enrich our unit test case and check whether we produce wanted output there. This would alleviate need fro pretty formatting. We currently under-use the cernopendata test suite, and in order to catch problems like these early, it would be easily to add some seriarisation test cases, since we have all the fixtures present in the source code for CI to see as well.  

Anyway, pretty-printing JSON by default in our HTML output of detailed record pages may be useful for anyone accessing only the web site still, so I guess let's keep it?

PR #3092, opened 2021-02-26
records: fix JSON export

### LINKED ISSUE #3064: records: fix JSON export
This works: https://opendata.cern.ch/record/451/export/json

This does not:  https://opendata-dev.cern.ch/record/451/export/json

audrium:
closes  #3064

In order to test:
 - populate some records with files (`/code/scripts/populate-instance.sh` without any parameters)
 - open any record and add `/export/json` to url
 - compare with production instance


mvidalgarcia (approved):
Added comment explaining why we need to override the `dump()` function.

PR #3093, opened 2021-02-26
ui: fix react children problem

### LINKED ISSUE #3082: ui: minified React error
I have observed the minified react error, similarly to what @ParthS007 reported a few days ago.

Here is the full recipe how to reproduce the problem on current master (`d7e4432d`).

**1.** First, load the site and a few test records:

```script
docker-compose down -v
docker-compose -f docker-compose-dev.yml down -v
docker-compose -f docker-compose-dev.yml build
docker-compose -f docker-compose-dev.yml up -d
sleep 20
docker exec -i -t opendatacernch_web_1 /code/scripts/populate-instance.sh --skip-records
docker exec -i -t opendatacernch_web_1 bash -c "cernopendata fixtures records -f /code/cernopendata/modules/fixtures/data/records/cms-tools-higgsexample20112012.json --mode insert-or-replace"
```

**2.** Second, do an empty search:

```console
$ firefox http://0.0.0.0:5000/search
```

**3.** Third, load some more records:

```script
docker exec -i -t opendatacernch_web_1 bash -c "cernopendata fixtures records -f /code/cernopendata/modules/fixtures/data/records/cms-derived-Run1-datascience.json -f cernopendata/modules/fixtures/data/records/cms-l1-trigger-information-Run2011A.json -f cernopendata/modules/fixtures/data/records/cms-trigger-information-2010.json -f cernopendata/modules/fixtures/data/records/cms-trigger-information-2011.json -f cernopendata/modules/fixtures/data/records/cms-trigger-information-2012.json --mode insert-or-replace"
```

**4.** Fourth, in the same Firefox tab, type `cms` in the search box to search only for CMS. This will lead to "empty results" page:

react-error

```

### PR DESCRIPTION
**Pull request by mvidalgarcia**
- root cause: `metadata.experiment` is an array of objects for
glossary records whereas it's an string for the rest of records.

closes #3082

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/js/search/components/ListItemLabels.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/search/components/ListItemLabels.js
@@ -1,5 +1,27 @@
+import PropTypes from "prop-types";
 import React from "react";
 import { Label } from "semantic-ui-react";
+
+const ExperimentLabel = ({ name, withKey }) => {
+  return (
+    <Label
+      className="tiny navy"
+      href={`/search?f=experiment:${name}`}
+      {...(withKey && { key: name })}
+    >
+      {name}
+    </Label>
+  );
+};
+
+ExperimentLabel.propTypes = {
+  name: PropTypes.string.isRequired,
+  withKey: PropTypes.bool,
+};
+
+ExperimentLabel.defaultProps = {
+  withKey: false,
+};
 
 const ListItemLabels = ({ metadata }) => {
   return (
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: cernopendata/modules/theme/assets/semantic-ui/js/search/components/ListItemLabels.js [hunk 1]
```diff
@@ -1,6 +1,28 @@
+import PropTypes from "prop-types";
 import React from "react";
 import { Label } from "semantic-ui-react";
 
+const ExperimentLabel = ({ name, withKey }) => {
```
ParthS007, line 5: We can create a separate file `ExperimentLabel` and import it here!

ParthS007 (commented):
Thanks for the PR. I tested with updated mappings, works fine.

cernopendata/modules/theme/assets/semantic-ui/js/search/components/ListItemLabels.js
mvidalgarcia: It's only being used in this file and I don't think it'll be used elsewhere anytime soon, so I think we can leave it for now, what do you think?

ParthS007: okay, we can leave it here for now 👍

ParthS007 (approved):

PR #3097, opened 2021-03-04
search: Hide ondemand dataset by default

### LINKED ISSUE #3089: Should on-demand datasets by included default?
The on-demand choice is now down in the left bar and not very visible, and on-demand datasets are included as default.

The default search on CMS simulated datasets:

https://opendata-dev.cern.ch/search?q=&f=experiment%3ACMS&f=type%3ADataset%2Bsubtype%3ASimulated&l=list&order=asc&p=1&s=10&sort=mostrecent

has the on-demand datasets right on the top. This gives a bad first impression as it looks like no datasets are actually available. 

Can the default be brought back to not including the on-demand datasets? Or not bringing them on top?

ParthS007:
closes #3089 

Screenshot 2021-03-09 at 11 14 24 AM

### DIFF
--- a/requirements-production.txt
+++ b/requirements-production.txt
@@ -63,8 +63,8 @@ idna==2.10                # via email-validator, requests
 importlib-metadata==3.3.0  # via jsonschema, kombu, markdown, pluggy
 invenio-accounts==1.4.2   # via invenio-files-rest
 invenio-app==1.3.0        # via cernopendata (setup.py)
-invenio-assets==1.2.5     # via cernopendata (setup.py), invenio-previewer, invenio-search-ui, invenio-theme
-invenio-base==1.2.3       # via cernopendata (setup.py), invenio-accounts, invenio-app, invenio-assets, invenio-cache, invenio-celery, invenio-db, invenio-files-rest, invenio-i18n, invenio-jsonschemas, invenio-logging, invenio-pidstore, invenio-previewer, invenio-records, invenio-records-rest, invenio-records-ui, invenio-rest, invenio-search, invenio-search-ui, invenio-theme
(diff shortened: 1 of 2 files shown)


cernopendata/templates/cernopendata_theme/header.html
```diff
@@ -18,6 +18,15 @@
         </div>
       </form>
   </div>
+  <div class="checkbox-ondemand">
+    <form>
+        {% set checked_value = request.args.get('f','').split(':')[-1] %}
+        <input onChange="this.form.submit()" type="checkbox" name="availability" value="{{checked_value}}"
+            {% if checked_value == 'ondemand' %}checked{% endif %}
+            aria-label="Include on-demand datasets">
```
tiborsimko, line 26: - "include" with lowercase "I"
- please also put a space after search box so that the `[ ] include...` would be a bit more separated

cernopendata/views.py
```diff
@@ -126,7 +126,10 @@ def _build_agg_sub_agg_qs(subagg_agg_mapping, args):
     for arg, arg_values in args.items():
         if arg in aggs.keys():
             for arg_val in arg_values:
-                qs_values["f"].append(f"{arg}:{arg_val}")
+                if arg == "availability":
```
tiborsimko, line 129: Beware of the behaviour when people check/uncheck the box repetitively.  For example, try the following scenarios: _(quoting search result numbers for a partially loaded site)_

- search for something e.g. CMS, giving 70 results 
- click on the checkbox, the page is reloaded, the "cms" search string is gone, and you see "no results found!" message
- uncheck the box again, and you are back at a search not containing "cms" string, with 110 results

The same applies for facets:

- make an empty query
- click on Documentation > Guide facet, obtaining 22 results
- click on on-demand checkbox, page is reloaded, "No results found!" message displayed, and the facet selection is gone

PR #3105, opened 2021-04-14
records: ATLAS data policy update

### LINKED ISSUE #3101: ATLAS Policy record update
Dear All,

We at ATLAS want to update the document (and some of the text)  in record http://opendata.cern.ch/record/413

I want to kindly
[ATLAS-Data-Access-Policy-Feb2021.pdf](https://github.com/cernopendata/opendata.cern.ch/files/6137880/ATLAS-Data-Access-Policy-Feb2021.pdf)
 ask if it is possible to simply change the PDF file for the newest one (attached here for reference)?
Or
If a need record is needed?

Thanks and cheers,
Arturo

PS: I will also update the JSON to update the ext, but before, I would like to ask if this will update the existing record (it seems more natural) or a new one.

tiborsimko:
Closes #3101.

@artfisica Here is how the updated page looks like:

atlas-policy

### DIFF
--- a/cernopendata/modules/fixtures/data/records/data-policies.json
+++ b/cernopendata/modules/fixtures/data/records/data-policies.json
@@ -190,7 +190,7 @@
   },
   {
     "abstract": {
-      "description": "This document contains the policy document regarding the access to ATLAS data by non-ATLAS members which was endorsed by the ATLAS Collaboration Board in June 2014."
+      "description": "This document contains the policy document regarding the access to ATLAS data by non-ATLAS members, which was endorsed by the ATLAS Collaboration Board in June 2014. The current revised version below was endorsed in February 2021."
     },
     "accelerator": "CERN LHC",
     "collaboration": {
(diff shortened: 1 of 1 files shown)


artfisica:
Hi @tiborsimko,

Yes, it looks fine, thank you!

I just cross-checked with management, and we would like to kindly ask you if the name of the old file can be changed from `ATLAS-Data-Policy.pdf` to `ATLAS-Data-Policy-2014.pdf`?

In this case, any new update (e.g. 2023) will be added to the same record, and it will be easy to differentiate *which is which* by the name.

All the rest is fine, and we thank you for this again,
Arturo

tiborsimko:
We cannot easily rename the old file, as that would break old URLs. Somebody might have relied on that due to minted DOI. Granted, this is important for datasets and less here, but still, the DOI/URL persistency principle guides against that. (This is probably why CMS opted for that other option with adding one new record, I guess.)

artfisica:
Oh, yes, I totally see it. Thanks, I did not think about the perennial property of the DOI when writing my last update. Let me come back to you on this

artfisica:
Hi @tiborsimko,
(the magic of been on-site at CERN)

We understand, and at this point, we are happy with the proposal you mentioned [one hour ago](https://github.com/cernopendata/opendata.cern.ch/pull/3105#issue-615090188).

So, please go ahead with that one, and thank you one more time :)
Arturo

PR #3117, opened 2021-06-23
global: upgrade invenio-assets and invenio-base

ParthS007:
invenio-assets -> 1.2.6
invenio-base -> 1.2.4


tiborsimko (approved):
Nicely fixes the Flask app runtime issues observed lately :+1:

PR #3124, opened 2021-06-28
dependency upgrade

tiborsimko:
- documentation: simplify development instructions
- installation: upgrade of all Python dependencies
- installation: pin XRootD
- installation: pin SQLAlchemy
- requirements: new pip-compile format


ParthS007 (commented):
Screenshot 2021-06-28 at 3 11 16 PM
 
I get this error in celery worker.  Related to upgrade of `flask-caching` I think.

tiborsimko:
Thanks, I see the error as well... Since we don't really use `worker`, it does not prevent from the site fully working... But I'll take care of fixing it

tiborsimko:
Hmm, I tested a bunch of dependencies, and at the end it was Celery itself that was causing the troubles... Pinned that one.

```console
$ docker-compose -f docker-compose-dev.yml logs worker
...
worker_1         | [2021-06-28 16:47:54,084: INFO/MainProcess] mingle: searching for neighbors
worker_1         | [2021-06-28 16:47:55,130: INFO/MainProcess] mingle: all alone
worker_1         | [2021-06-28 16:47:55,174: INFO/MainProcess] celery@3c00db0604f7 ready.
```

ParthS007 (approved):
Thanks for the PR. Tested, Works fine! 🚀

PR #3126, opened 2021-07-07
search: hide ondemand dataset by default

### LINKED ISSUE #3089: Should on-demand datasets by included default?

ParthS007:
closes #3089 

Screenshot 2021-07-09 at 12 20 08 PM


tiborsimko:
Note one failed test:

```
=========================== short test summary info ============================
FAILED tests/test_cernopendata_query_parser.py::test_cernopendata_query_parser
================== 1 failed, 12 passed, 25 warnings in 2.64s ===================
```

Please also rebase against latest `master`.

cernopendata/modules/theme/assets/semantic-ui/js/search/components/Facets.js
```diff
@@ -0,0 +1,50 @@
...
+        label="View ondemand datasets"
+        filterValue={["ondemand", "true"]}
+      />
+      {aggs.map((agg) => {
+        return (
+          <div className="ui accordion facet-margin" key={agg.title}>
+            <BucketAggregation title={agg.title} agg={agg.agg} />
+          </div>
+        );
+      })}
```
audrium, line 45: Why it is needed to wrap it with these extra css classes? I've tested
```
  {aggs.map((agg) => (
    <BucketAggregation key={agg.title} title={agg.title} agg={agg.agg} />
  ))}
  ```
  and it works exactly the same for me, margin between the facets is identical. Maybe I'm missing something?

File: cernopendata/modules/theme/assets/semantic-ui/js/search/components/Facets.js [hunk 2]
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * In applying this license, CERN does not
+ * waive the privileges and immunities granted to it by virtue of its status
+ * as an Intergovernmental Organization or submit itself to any jurisdiction.
+ */
+
+import React from "react";
+import { BucketAggregation, Toggle } from "react-searchkit";
+
+
```
audrium, line 30: Very minor: there is an extra line which could be removed

cernopendata/modules/theme/assets/semantic-ui/js/search/components/Facets.js
```diff
@@ -0,0 +1,50 @@
...
+const CODFacets = ({ aggs }) => {
+  return (
+    <>
+      <Toggle
+        title="Availability"
+        label="View ondemand datasets"
+        filterValue={["ondemand", "true"]}
+      />
```
audrium, line 38: Screenshot 2021-07-09 at 13-14-30 CERN Open Data Portal
For me the label of this component is displayed in two lines. Could be easily fixed with some css

cernopendata/modules/theme/assets/semantic-ui/js/search/components/Facets.js
ParthS007: I took the component from RDM and now following the native component. Removed the extra div. Thanks for the catch.

cernopendata/modules/theme/assets/semantic-ui/js/search/components/Facets.js
ParthS007: What about just keeping `Ondemand datasets`, is it intuitive?

cernopendata/modules/theme/assets/semantic-ui/js/search/components/Facets.js
```diff
@@ -0,0 +1,47 @@
...
+const CODFacets = ({ aggs }) => {
+  return (
+    <>
+      <Toggle
+        title="Availability"
+        label="Ondemand datasets"
+        filterValue={["ondemand", "true"]}
+      />
+      {aggs.map((agg) => {
+        return (
```
audrium, line 39: Very minor: `return` is not really needed for one liners, you can do it like this:
```js
{aggs.map((agg) => (
  <BucketAggregation key={agg.title} title={agg.title} agg={agg.agg} />
))}
```

cernopendata/modules/theme/assets/semantic-ui/js/search/components/Facets.js
tiborsimko: The original text, i.e. "include on-demand datasets", would be the clearest.  

Otherwise people may wonder why there is no "availability: online" for example. This might actually be the shortest option:

```
Availability:

[x] online
[ ] ondemand
```

I'll have a comment about the order of facets, e.g. they don't match the current order... So we could perhaps muse and fix it in a separate ticket...

tiborsimko (approved):
Works well :+1: However see a comment about facet order etc.  We can also address this afterwards in another issue.

PR #3135, opened 2021-07-14
facets: use filter of filtered query

### LINKED ISSUE #3098: search: n records displayed after each search term
For the search, I would expect the numbers after search terms to reflect the actual selections.

Indeed, the production search http://opendata.cern.ch/search?page=1&size=20&type=Dataset&subtype=Simulated&year=2011&experiment=CMS
shows the following:
image

The dev search (excluding on-demand) https://opendata-dev.cern.ch/search?q=&f=experiment%3ACMS&f=type%3ADataset%2Bsubtype%3ASimulated&f=year%3A2011&l=list&order=asc&p=1&s=10&sort=mostrecent
shows:

image

i.e. it shows the total number even if selections have been applied. The search results are correct.

Is showing the total the intended behaviour?

ParthS007:
closes #3098 

I did some investigation on how numbers are displayed in front of facets and found out that number of aggregation after every query effects the count shown in front of every facet.

>  As far as search hits are concerned, they are the same thing, i.e. the hits will be correctly filtered according to either the filter in a `filtered`  query or the filter in the `post_filter`.
   However, as far as `aggregations` are concerned, the end result will not be the same. The difference between both comes down to what document set the `aggregations` will be computed on.
    - If the filter is in a `filtered` query, then `aggregations` will be computed on the document set selected by our query(ies) and the filter(s) in the filtered query, i.e. the same set of documents that we will get in the response.
    - If the filter is in a `post_filter`, then the `aggregations` will be computed on the document set selected by our various query(ies). Once aggregations have been computed on that document set, the latter is further filtered by the filter(s) in our `post_filter` before returning the matching documents.

To sum it up:

- a `filtered query` affects both search results and aggregations
- while a `post_filter` only affects the search results but NOT the aggregations

That's the reason I went for `filter` in current PR.

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -386,7 +386,7 @@ RECORDS_REST_FACETS = {
                 field='keywords.keyword',
                 order=dict(_key='asc'))),
         ),
-        'post_filters': dict(
+        'filters': dict(
             type=terms_filter('type.primary.keyword'),
             subtype=terms_filter('type.secondary.keyword'),
             experiment=terms_filter('experiment.keyword'),


cernopendata/config.py
```diff
@@ -386,7 +386,7 @@
                 }
             }
         ),
-        'post_filters': dict(
+        'filters': dict(
```
tiborsimko, line 389: There is a problem with this approach: do an empty search, click on "Datasets", and while the facet numbres are well updated, it is no longer possible to select "Software" in addition to "Datasets", say, because the other facet categories that are on Datasets level -- i.e. Documentation, Environment, etc -- disappear and are no longer visible.

It is important to be able to select multiple values, e.g. years 2010, 2011, 2012 from the choice.

Can you please look and amend so that multiple-value selection will behave as in PROD?

ParthS007: While investigating, there are two different components `Pagination` and `ResultsPerPage` which shows the pagination bar in the bottom and X result per page respectively. The issue requires to access one prop of `ResultsPerPage` component i.e `currentSize` just outside the Pagination. Tried with state hook and passing of currentQueryState but no success so far.

PR #3136, opened 2021-07-15
search: hide paginator when results less than results per page

### LINKED ISSUE #3083: search: unnecessary paginator
**Current behavioun**

When a search returns low numbers of results, such as 3, we are still showing the pagination element:

search-pages

This is not necessary.

**Expected behaviour**

- When there are less than N hits, where N is the wanted number of results per page set by the user, the pagination element can stay hidden.

- On a similar cosmetic tune, `3 result(s) found` reads a bit hard. Can we do a quick `if` to detect singular/plural and display `1 result found`, or `N results found` if N is greater than 1?

ParthS007:
closes #3083

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/js/search/app.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/search/app.js
@@ -25,12 +25,13 @@
  */
 
 import { createSearchAppInit } from "@js/invenio_search_ui";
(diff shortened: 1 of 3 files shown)


ParthS007:
While investigating, there are two different components `Pagination` and `ResultsPerPage` which shows the pagination bar in the bottom and X result per page respectively. The issue requires to access one prop of `ResultsPerPage` component i.e `currentSize` just outside the Pagination. Tried with state hook and passing of currentQueryState but no success so far.

PR #3138, opened 2021-08-05
search: use native search bar from react-searchkit

### LINKED ISSUE #3130: facets: better interplay witth search box
Search for something, e.g. "cms". Get 4011 hits.

Click on a facet, e.g. "Derived datasets". Obtain 136 hits.

Refine search, e.g. "cms event" and click on enter.

Current behaviour: the previous fact selection is lost and you get 5039 hits.

Expected behaviour: you get a "cms event" query plus "derived datasets" facet, so the number of hits should be less than 136.

### LINKED ISSUE #3133: facet: unselecting daugther facets unselects parent facets
As an example UX issues behind #3132, try the following:

- Do an empty search.
- Select "Documentation" facet by ticking the checkbox.
- Select "Authors" facet by ticking the checkbox.
- Select "Guide" facet by ticking the checkbox.
- Unselect "Guide" facet by unticking the checkbox.

Current behaviour: all the facet selection is gone and one gets to see full 7624 hits.

Expected behaviour: we would be at the state where Documentation and Guide are selected... Basically going back in the stack if the last fact selection was undone.

ParthS007:
https://user-images.githubusercontent.com/24358501/128357555-25e5b989-d024-4699-8ea4-cf543b22a451.mov

https://user-images.githubusercontent.com/24358501/128518247-b28dea58-6585-474e-a477-4e2f24afa794.png
Screenshot 2021-08-06 at 4 51 43 PM
closes #3130 
closes #3133

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/js/search/app.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/search/app.js
@@ -25,13 +25,14 @@
  */
 
 import { createSearchAppInit } from "@js/invenio_search_ui";
-import { CODLayoutSwitcher, ResultsCount, CODFacets, CODFacetItem } from "./components";
+import { CODLayoutSwitcher, ResultsCount, CODFacets, CODFacetItem, CODSearchBarContainer, CODSearchBarElement } from "./components";
 
 const initSearchApp = createSearchAppInit({
   "LayoutSwitcher.element": CODLayoutSwitcher,
   "ResultsGrid.container": null,
   "Count.element": ResultsCount,
-  "SearchApp.searchbarContainer": () => null,
   "SearchApp.facets": CODFacets,
+  "SearchApp.searchbarContainer": CODSearchBarContainer,
+  "SearchBar.element": CODSearchBarElement,
   "BucketAggregationValues.element": CODFacetItem,
 });
--- a/cernopendata/modules/theme/assets/semantic-ui/js/search/components/SearchBarContainer.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/search/components/SearchBarContainer.js
@@ -24,15 +24,14 @@
  * as an Intergovernmental Organization or submit itself to any jurisdiction.
  */
 
+import React from "react";
+import { SearchBar } from "react-searchkit";
 
-a:hover {
-    text-decoration: underline;
-}
+export const CODSearchBarContainer = ({
+  }) => {
+    return (
+        <SearchBar />
+    );
+  };
 
-.navbar-item {
-    display: block;
-}
-
-.search-bar {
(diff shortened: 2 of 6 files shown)


cernopendata/modules/theme/assets/semantic-ui/js/search/app.js
```diff
@@ -31,7 +31,6 @@ const initSearchApp = createSearchAppInit({
   "LayoutSwitcher.element": CODLayoutSwitcher,
   "ResultsGrid.container": null,
   "Count.element": ResultsCount,
-  "SearchApp.searchbarContainer": () => null,
```
tiborsimko, line 34: This seems to fix the problem indeed.

Two comments:

- Functionally, when I go to the home page do an empty search, I get _"No results found!"_, while on opendata-dev.cern.ch (i.e. in `master` branch) I get well all the hits.

- Cosmetically, it is OK to move to native RSK bar if it improves functionality, but can we amend the presentation a bit?  E.g. put the bar in banner-like position, or keep it where it is but move it more to the left to be more aligned with where facets start?

ParthS007: - I checked the functionality again, I am getting all the hits locally. Did webpack recompiled again? Can you please check once again after recompiling the JS.

- The search bar is highly nested, I am currently checking if I can override some css to change the alignment.

tiborsimko: > I checked the functionality again, I am getting all the hits locally. Did webpack recompiled again?

I confirm that it works :+1: Perhaps I must have cleaned the instance yesterday without loading records anew?!

```console
$ docker-compose down -v
$ docker-compose build
$ docker-compose up -d
$ docker exec -i -t opendatacernch_web_1 scripts/populate-instance.sh --skip-records
$ docker exec -i -t opendatacernch_web_1 cernopendata fixtures records --mode insert-or-replace -f cernopendata/modules/fixtures/data/records/cms-primary-datasets.json
$ docker exec opendatacernch_nginx_1 find /var/cache/nginx -type f -delete
$ firefox http://localhost/search
```

tiborsimko: > The search bar is highly nested, I am currently checking if I can override some css to change the alignment.

Another small cosmetic thing you can fix: (a) on the home page, the search box has "Start typing..."; (b) on the search results page, the search box has "Type something".  We should unify the text to say the same thing.  Perhaps "Start typing..." bu I would simply erase the hint and not display anything.  Because each user knows how to search... We don't need to hint at anything! The search boxes can remain clean and zen.

tiborsimko (approved):

PR #3139, opened 2021-08-09
es: map orcid using keyword

### LINKED ISSUE #3085: search: better ORCID searches
**Current behaviour**

Currently there are too many results found for ORCID searches on the DEV server:

- "0000-0002-2634-4958" ... 1 hit
- 0000-0002-2634-4958 ... 201 hits, where the "proper hit" is on the 4th place only

Compare to production, where there are 200 hits, and the "proper hit" is on the 1st place.

**Expected behaviour**

We can improve the situation so that DEV would resemble PROD more.  Two options:

- option a: study why the proper hit is only 4th, and amend boosting etc to make it appear on top.  IOW, this will improve the ordering issue, but keep the too-many-hits issue.

- option b: amend tokenisation so that ORCID field would not be split into terms, but index as a whole, which will presumably also reduce the total number of hits. 
should amend ORCID tokenisation process.  IOW, this could fix both the search order and reduce the number of "wrong" hits.

P.S. The treatment of `-` characters in ORCID values resembles the treatment of `/` characters in CMS dataset titles, which can also be solved at index-time vs at search-time.

_Originally posted by @tiborsimko in https://github.com/cernopendata/opendata.cern.ch/pull/3049#discussion_r579323109_

ParthS007:
Screenshot 2021-08-10 at 1 34 25 PM
Screenshot 2021-08-10 at 1 34 44 PM
closes #3085

### DIFF
--- a/cernopendata/mappings/v7/records/record-v1.0.0.json
+++ b/cernopendata/mappings/v7/records/record-v1.0.0.json
@@ -51,12 +51,7 @@
               }
             },
             "orcid": {
-              "type": "text",
-              "fields": {
-                "exact": {
-                  "type": "keyword"
-                }
-              }
+              "type": "keyword"
             },
             "name": {
               "type": "text",
--- a/cernopendata/modules/records/search/query.py
+++ b/cernopendata/modules/records/search/query.py
@@ -42,6 +42,8 @@ def cernopendata_query_parser(query_string=None, show_ondemand=None):
     for index, _query_term in enumerate(_query_string):
         if "/" in _query_term and '"' not in _query_term:
             _query_string[index] = '"' + _query_term + '"'
+        if "-" in _query_term and '"' not in _query_term:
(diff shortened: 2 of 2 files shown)


cernopendata/mappings/v7/records/record-v1.0.0.json
```diff
@@ -51,12 +51,7 @@
               }
             },
             "orcid": {
-              "type": "text",
-              "fields": {
-                "exact": {
-                  "type": "keyword"
-                }
-              }
+              "type": "keyword"
```
tiborsimko, line 54: I get 10 results when I try from scratch, so it's even better than your screenshot :wink:

tiborsimko (approved):
Works nicely :+1: after changing the parser, an ORCID search gives one hit. Otherwise the response is OK for queries like `high-energy` as well.

PR #3142, opened 2021-08-17
es-index: update mappings to match with production

ParthS007:


tiborsimko:
Have you tried in the production mode? It does not seem to work, the numbers don't update for me, and sometimes I get even "oops", as in this case: first do:

```console
$ docker exec opendatacernch_nginx_1 find /var/cache/nginx -type f -delete
$ firefox https://localhost/search?f=type%3ADataset
```

... and now click on "Guide" and you'll see "Oops! Something went wrong while fetching results."

P.S. I tested the above afer building the production-like site like this:

```console
docker-compose down -v
docker-compose build
docker-compose up -d
docker exec -t opendatacernch_web_1 scripts/populate-instance.sh
```

PR #3180, opened 2021-11-22
Insert replace with files

### LINKED ISSUE #2150: fixtures: insert/replace scenario with files
PR  #2046 imrproved the behaviour of insert, replace, insert-or-replace mode to load records with or without files. It addressed well the main use cases behind #2039.

However, one corner case issue remains with respect to deleting previously uploaded files:

```console
$ ./scripts/clean-instance.sh 
$ ./scripts/populate-instance.sh --skip-records
    -f ./cernopendata/modules/fixtures/data/records/cms-tools-vm-image.json \
    --mode insert
$ curl http://0.0.0.0:5000/record/249/files/cms-user-data.txt 
$ curl http://0.0.0.0:5000/record/249/files/cms-user-data.txt 
    -f ./cernopendata/modules/fixtures/data/records/cms-tools-vm-image.json \
    --mode replace --skip-files
$ curl http://0.0.0.0:5000/record/249/files/cms-user-data.txt
```

The detailed record page shows well that there are no files attached anymore, but the file download URL still serves the file content. One would expect to get a 404 not found in this scenario.

Reminder of [wanted behaviour](https://github.com/cernopendata/opendata.cern.ch/pull/1887#issuecomment-348214957):

**mode** | **behaviour when PID exists** | **behaviour when PID does not exist** | 
--- | --- | --- |
insert | fail | upload as new | 
replace | update as existing | fail | 
insert or replace | update existing | upload as new | 

Both when including files or when excluding files (if `-skip-files` option was used).

wahaj:
* Deletes the files and returns a 404 message upon insert or replace
  files when using the `--skip-files` flag
* Fixes #2150


wahaj:
Hi @tiborsimko ,
Could you have a look at this?

PR #3209, opened 2021-12-19
release text

### LINKED ISSUE #3197: CMS: create record for the release announcement
🥳

katilp:
(closes #3197 )


cernopendata/modules/fixtures/data/docs/cms-releases-first-run2-data/cms-releases-first-run2-data.json
```diff
@@ -0,0 +1,27 @@
...
+      {
+        "primary": "News"
+      }
+    ],
+    "created": "2021-12-20",
+    "experiment": "CMS",
+    "short_description": {
+      "content": "As the experiments at the Large Hadron Collider (LHC) brace for the start of Run 3 of the accelerator’s programme in 2022, the CMS collaboration has released a new batch of research-quality open data recorded by the CMS detector in 2015, the first year of Run 2. The new datasets are now available on the CERN Open Data portal."
+    },
+    "slug": "first-cms-open-data-from-lhc-run-2-released",
```
tiborsimko, line 21: Will the title be "First CMS open data from LHC Run 2 released"? 

With this title and this slug, the URL would be `http://opendata.cern.ch/docs/first-cms-open-data-from-lhc-run-2-released`, so I think the directory and the files should not be named `cms-releases-first-run2-data.md`, but rather `first-cms-open-data-from-lhc-run-2-released.md`.

BTW note that the title diverges from the past announcements that used to use the present tense ("releases") instead of past tense ("released"). Perhaps the word "released" could even be removed from the slug and the title... Just some cosmetics observations.

boxes

cernopendata/modules/fixtures/data/docs/cms-releases-first-run2-data/cms-releases-first-run2-data.json
```diff
@@ -0,0 +1,28 @@
...
+      {
+        "experiment": "CMS"
+      },
+      {
+        "primary": "News"
+      }
+    ],
+    "created": "2021-12-20",
+    "experiment": "CMS",
+    "featured": 1,
```
tiborsimko, line 18: Thanks, I've added "featured" here and amended the old news to have the latest news box order correct.

Please see below a question about slug; I can rename the files accordingly depending on what's chosen as the final title and final slug.

cernopendata/modules/fixtures/data/docs/cms-releases-first-run2-data/cms-releases-first-run2-data.json
katilp: Yes, that's the title. OK for me for removing "released" and having the same name for the directory and the title. I would not change the title too much (if that's for what is used by the CMS announcement) but removing "released" keeps it close enough.

tiborsimko: OK, kep the title "as is", and changed the file names to be equal to slug.

tiborsimko (approved):

PR #3222, opened 2022-01-17
records: fix CERN open data policy authorship

tiborsimko:
Fixes DOI registration by adding CERN as the corporate author of the
"CERN Open Data Policy for the LHC Experiments" record.

Adds a link to the CERN Document Server record as well.


DEVELOPING.rst
```diff
@@ -267,9 +267,9 @@ Here is example:
 
    $ git checkout master
    $ git checkout -b improve-event-display-icons
-   $ emacsclient some_file.py
```
mvidalgarcia, line 270: Oh wow 👀

Dockerfile
```diff
@@ -25,17 +25,22 @@
 # Use CentOS7
 FROM centos:7
 
+# Install Nodejs 6 from Nodesource early. Doing so after installing EPEL7 makes
```
mvidalgarcia, line 28: I think Node.js is the official name.
```suggestion
# Install Node.js 6 from Nodesource early. Doing so after installing EPEL7 makes
```

mvidalgarcia (approved):
LGTM.

DEVELOPING.rst
tiborsimko: Hehe, it's just a documentation :) Changing to `$EDITOR` will make it copy-paste-friendly.

PR #3254, opened 2022-07-28
WIP: LHCb Records for testing

MindaugasSarpis:
This is the initial pull request where LHCb open data records, documentation pages and terms were added to the open data portal.


tiborsimko:
Comments on the stripping line pages such as http://localhost/docs/stripping21-stdalllooseelectrons

- Link style: For outside links, you use `[Stripping Project homepage]` style. For inside links, the same style is used to point to `[stripping21 lines]`. So that users may think they will also be leaving the site and going somewhere else. Could we use a different linking style? E.g. just call the second link "Stripping21 lines" without brackets.

- External link: Do we actually need a link to the "Stripping Project homepage" at the top of each page? It drives users away, and it's the same URL link for all the `/docs/xxxx` pages, so it does not really bring any useful information there? We could print the link just in the docs, and not the same one for each of these docs pages.

- Year: The docs records are missing year information. So when you search, you get 2157 documentation hits, but "filter by year" is not really applicable. Could we add year field to docs stripping lines, if the mapping to collision taking is relatively straightforward? Or perhaps it does not make sense?

tiborsimko:
Comments on faceting:

- You can start using the new facet-oriented stripping field already now. The system is not recognising it, but it won't prevent you from uploading those files. As discussed, you can use something like:

```json
"lhcb_stripping": {
    "version": "v21",
    "stream": "leptonic"
  },
```

- Looking at titles, we already have a field for beam energy (and corresponding facet), so the only thing we would be lacking is MagDown MagUp information. Do you wish to introduce a facet for this? If so, we may need a new field.

tiborsimko:
Also, the dataset record titles say `LHCb 2011 Beam3500GeV MagDownEW stream stripping21r1p2` for http://localhost/record/24518. Note that there should probably be an empty space after "MagDown" and before "EW", and that "stripping21r1p2" should perhaps start with uppercase "S" as the other title items?

MindaugasSarpis:
TODO list for myself to track the progress. 

- [ ] Record file name, LFNs file name and index files should have the year of data taking added to them. (Title of a record and the file name should be generated more uniformly) 
- [ ] ConfigVersion (the type of collision or MC) should be lowercase (Collision -> collision) 
- [ ] Add introduction to Production Steps. 
- [ ] Should hide `preview` in `Production Steps` scripts.
- [ ] Refine text in "How to use this data" section.
- [ ] Use different style for outside links (or maybe don't have them at all) 
- [ ] Add year field to docs jsons
- [ ] Write stripping fields to the records/docs jsons. 
- [ ]  MagUp/MagDown is quite instrumental for LHCb analyses (especially CP violation studies) it's best if we can make a facet for it. 
- [ ] Space in the title of records 
- [ ] Capitalize stripping in the title

MindaugasSarpis:
RECID: should point to the authors list. We can export it for 2011, 2012 (any month, so say, December). Will have to check how to fit it into ODP formatting.

tiborsimko:
> RECID: should point to the authors list. 

You can use record ID 455 for linking to the author list instead of the `FIXME1`. The ID 455 will be reserved to represent the LHCb author list record for this releas. I'll make a skeleton and push it here.

MindaugasSarpis:
Yes, the idea is to have the authors list for each year of the data taking. (so for now and for a while it will be 2011, 2012) but later on the other years would be added. (2015,2016,2017,2018)

tiborsimko (approved):

PR #3268, opened 2022-09-12
Stripping lines pages

MindaugasSarpis:
Fixed bugs related to creating an incorrect slug for small portion of the stripping line doc pages. 
Unified stripping line indices (they are now produced automatically).


tiborsimko:
@MindaugasSarpis I have rebased this against latest `qa'`, and it is good to go, however do we need to amend the collision dataset records as well? E.g. http://localhost/record/24521 points to http://localhost/docs/stripping21r1p2-ew which does not exist, and ditto http://localhost/record/24508 points to http://localhost/docs/stripping21-radiative which does not exist, etc.

tiborsimko (approved):

PR #3271, opened 2022-09-14
Terms with short descriptions

MindaugasSarpis:
- Added short description field for glossary items (from memdoc)


tiborsimko (approved):
Thanks, slightly edited to keep the ones without HTML markup.

PR #3277, opened 2022-09-23
added spaces between badges in news and updates closes #3232

### LINKED ISSUE #3232: Badges are not separated by spaces in the 'News & Updates'
This is how a record looks like in the Search results:
immagine
However, in the cards rendered in the "News & Updates" from the landing page, the badges/labels are not separated between each other:
immagine

sksDonni:
Added spaces between badges in the news and updates section to keep in line with the general site theme.
Requesting a PR directly into production as the master branch has a lot of diff with the production branch


tiborsimko (approved):
Thanks for the update, rebased and squashed

PR #3302, opened 2022-11-21
Lhcb open data release

MindaugasSarpis:
Some fixes to index pages, lhcb-about, lhcb-getting-started, lhcb-masterclass pages separated.

PR #3303, opened 2022-11-22
docs: LHCb stripping index page updates

tiborsimko:


tiborsimko (commented):
Taken from #3302 as a part that is ready-to-go.

The docs updates will be worked on by @MindaugasSarpis still.

PR #3317, opened 2022-12-01
records: update cms lumi record text

### LINKED ISSUE #3282: CMS: Record 1052 should point to 1054 more clearly
Record [1052](http://opendata.cern.ch/record/1052) describes 2012 luminosity, but it's an old version of the luminosity estimation. There's an updated one in record [1054](http://opendata.cern.ch/record/1054) that people should be pointed to at the very top of the record, in particular since there's no connection to 2012 in the first paragraph. 1052 should be considered as deprecated.

katilp:
(closes #3282)

- adds an explicit pointer to the updated records in the description of the old records
- in the description of the updated records, removes the reference to the outdated record for the prescaleYYYY.txt file

To do:
- attach prescale2011.txt to 1053
- attach prescale2012.txt to 1054


tiborsimko:
> attach prescale2011.txt to 1053
> attach prescale2012.txt to 1054

Since these files are already attached to past records, that are referenced by other persistent identifiers, we cannot "reuse" the same name and the same path location for another persistent identifier. We would need to copy the file and user another file name and/or path, for example `2011lumi_prescales.txt` in the new record. Would that be OK?

katilp:
Yes, new names are fine. Would you be able to change the then names in the description on merge?

tiborsimko:
> Yes, new names are fine. Would you be able to change the then names in the description on merge?

Yes, I can changed them everywhere. 

Going for `YYYYlumi_prescales.txt` file name then...

tiborsimko (approved):
Thanks, new files added, links updated.

PR #3321, opened 2022-12-02
docs: cms announcement

katilp:


tiborsimko (approved):
Thanks, added featured news item ordering.

PR #3323, opened 2022-12-03
records: CMS POET software tarballs

### LINKED ISSUE #3278: CMS: prepare SW records for POET
Prepare the SW records for POET code, with the json fields as in [trigger sw example records](https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/modules/fixtures/data/records/cms-trigger-examples.json) with the file name `cms-tools-poet.json`.

There will be three records
- 2011 with https://github.com/cms-opendata-analyses/PhysObjectExtractorTool/tree/2011
- 2012 with https://github.com/cms-opendata-analyses/PhysObjectExtractorTool/tree/2012
- 2015 with https://github.com/cms-opendata-analyses/PhysObjectExtractorTool/tree/odws2022-ttbaljets-prod

unless there's a good reason to go for a single record with these three versions inside.

Verify the different metadata fields (e.g. a pointer to the OD guide for the usage?) and foresee the usage with the derived data produced for the workshop (at least for the 2015, maybe also for 2012)

tiborsimko:
Closes #3278.

### DIFF
--- a/cernopendata/modules/fixtures/data/records/cms-tools-poet.json
+++ b/cernopendata/modules/fixtures/data/records/cms-tools-poet.json
@@ -16,7 +16,21 @@
       "2012"
     ],
     "date_published": "2022",
+    "distribution": {
+      "formats": [
+        "zip"
+      ],
+      "number_files": 1,
+      "size": 3749090
+    },
     "experiment": "CMS",
+    "files": [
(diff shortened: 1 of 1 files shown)


katilp (approved):
Looks good, thanks.

PR #3346, opened 2023-02-13
ui: amend brief format for glossary terms

### LINKED ISSUE #3345: ui: amend brief format for glossary terms
**Current behaviour**

On `master` branch, load glossary terms as follows:

```console
$ docker exec -i -t opendatacernch_web_1 cernopendata fixtures glossary --mode insert-or-replace
```

Then search in the web UI for LHCb glossary terms containing memdoc:

- http://localhost:5000/search?q=memdoc&l=list&order=asc&p=1&s=10&sort=bestmatch

brief-master

Compare to the same search on `qa` or `production` branches:

- https://opendata.cern.ch/search?page=1&size=20&q=memdoc

brief-production

Basically, unlike on `qa` or `production`, the `master` branch does not "filter out" HTML in the brief formats of glossary records. This is a consequence of Angular -> React Semantic UI change.

**Expected behaviour**

Amend brief output format templates to filter away HTML in LHCb glossary records so that the output on `master` is nicely readable as on `qa` or `production`.

DaanRosendal:
closes #3345


tiborsimko (dismissed):

cernopendata/modules/theme/assets/semantic-ui/js/search/components/TermListItem.js
```diff
@@ -37,7 +38,9 @@ const CODTermListItem = ({ result: { metadata }, index }) => {
       <Item.Content>
         <Item.Header href={link}>{metadata.anchor}</Item.Header>
         <Item.Description href={link}>
-          <p>{truncate(metadata.definition, { length: 200 })}</p>
+          <p>
```
tiborsimko, line 41: This solves the described problem for the LHCb terms, but it breaks the display of general glossary terms such as "Muon", that now have no description.

Basically, we need to check whether the term has the "short_description" field defined, and if yes, then we use it (and strip HTML); otherwise we have to use the regular "description" field (where there may not be the necessity to strip HTML).

Please see the display logic for glossary definition fields in the `qa` branch in `cernopendata/static/templates/search/briefs/terms.html`.

DaanRosendal: Nice catch! I solved it by checking if `short_description` exists. If it doesn't, it uses the standard definition. Stripping the HTML is not necessary anymore in any case.

tiborsimko (approved):

PR #3361, opened 2023-04-20
records: CMS 2015 HI ppref

tiborsimko:
- CMS 2015 HI ppref collision datasets
- CMS 2015 HI ppref trigger information record
- CMS 2015 HI ppref HLT trigger path records
- CMS 2015 HI ppref RECO configuration file records


tiborsimko:
FYI depends on #3418 so let's review and merge first that one.

katilp (approved):

PR #3371, opened 2023-06-12
cms-validated-runs_HIRun1.json: add HI2013 related records

katilp:
Creates records for validated runs for 2013 HI-related records using
https://github.com/cernopendata/data-curation/tree/master/cms-YYYY-validated-runs

My docker-compose build has trouble so I could not view the records locally.


tiborsimko (approved):
Added files and mutually corrected the wrong energy information in the free text.

Issue #3386, opened 2023-07-04
CMS: add image names as metadata fields 


katilp:
The image names are currently in the documentation page http://opendata.cern.ch/docs/cms-guide-docker
but not in the records.
Enrich the metadata with the image names

PR #3391, opened 2023-07-11
records: update json schema to include container images

nancyhamdan:
fixes [#3386](https://github.com/cernopendata/opendata.cern.ch/issues/3386)
updated records' JSON schema to include container images metadata fields

### DIFF
--- a/cernopendata/jsonschemas/records/record-v1.0.0.json
+++ b/cernopendata/jsonschemas/records/record-v1.0.0.json
@@ -79,10 +79,6 @@
             "description": "The affiliation of the author. Multiple affiliations are separated by a semicolon (;)",
             "type": "string"
           },
-          "rorid": {
-            "description": "The ROR ID for the author's affiliation. Multiple ROR IDs are separated by a semicolon (;)",
-            "type": "string"
-          },
           "ccid": {
             "description": "The author's CCID",
             "type": "string"
@@ -91,12 +87,16 @@
             "description": "The author's INSPIRE ID",
             "type": "string"
           },
+          "name": {
+            "description": "The name of the author",
+            "type": "string"
+          },
           "orcid": {
             "description": "The author's ORCID iD",
             "type": "string"
           },
-          "name": {
-            "description": "The name of the author",
+          "rorid": {
+            "description": "The ROR ID for the author's affiliation. Multiple ROR IDs are separated by a semicolon (;)",
             "type": "string"
           }
         },
@@ -204,8 +204,11 @@
       "properties": {
         "availability": {
           "description": "Specifies if dataset is on-demand, online in future might be also tape, disk",
-          "type": "string",
-          "enum": ["ondemand", "online"]
+          "enum": [
+            "ondemand",
+            "online"
+          ],
+          "type": "string"
         },
         "formats": {
           "items": {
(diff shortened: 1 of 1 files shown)


cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -523,6 +526,30 @@
     },
     "system_details": {
       "properties": {
+        "container_images": {
+          "items": {
+            "properties": {
+              "description": {
+                "description": "A brief description of the container image",
```
tiborsimko, line 533: It may be useful to specify that the image is used for analysing the data, for example:

- Description of the container image that is recommended for analysing these data

(Alternatively, if we could have several images linked here, one recommended for analysis, one how the data was born, we could even introduce a new property called "purpose" or some such... But at this moment it's probably an overkill, since we plan to store only the images useful for data analyses. Unless there could be several images in the future for different purposes... CC @katilp )

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -523,6 +526,30 @@
...
+                "type": "string"
+              },
+              "recid": {
+                "description": "The internal ID of the container image, if it is another Open Data record",
```
tiborsimko, line 537: Let's start again by the same word as the field is called (`recid`) so that readers can form better mental association:

- Record ID of the container image (if it exists as another Open Data record)

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -523,6 +526,30 @@
...
+                "type": "string"
+              },
+              "registry": {
+                "description": "The type of registry the image can be found at",
```
tiborsimko, line 541: It may be useful to provide example values in the documentation:

- Registry type where the image can be found (e.g. dockerhub, cerngitlab)

cernopendata/jsonschemas/records/record-v1.0.0.json
```diff
@@ -523,6 +526,30 @@
...
+                "type": "string"
+              },
+              "uri": {
+                "description": "The URI for finding the container image",
```
tiborsimko, line 545: Again here:

- URI where the container image is located

However, we may perhaps be better off using a common term FQIN here to denote fully-qualified image name concept, i.e. the field would be called "fqin" instead of "uri" and the description would explain it as follows:

- Fully Qualified Image Name (FQIN) where the container image is located (repository/name:tag)

We can discuss live also with @katilp before merging

tiborsimko: P.S. BTW it the "fqin" field name risks looking a bit weird, we could use some longer name variant such as "fully_qualified_name", or just "name", and document that the name should be FQIN in the schema.

cernopendata/jsonschemas/records/record-v1.0.0.json
katilp: Yes, for now, we do not have plans to link to other images than that to be used for analysing the data.

(I could only imagine a separate image for MC to generate a new similar MC dataset with some changes in the input but we are not providing those images and I'd rather not link them to the records in any case)

tiborsimko (approved):
Thanks, rebased against latest QA.

After merge, please prepare another PR with a small example file containing the data as well as the output format Jinja snippet displaying it, as we had talked about IRL.

PR #3418, opened 2023-08-22
records: new field run_numbers

### LINKED ISSUE #3383: CMS: run range indication in the dataset
The text of the run range has become e.g.

"PPJet primary dataset in RECO format from RunA of 2013. Run period from run number 211739 to 211831."

This is however not quite correct as the run range is taken from the run range of the data-taking period but this specific dataset may have different range.

Better: Data taking in Run2013A goes from run number 211739 to 211831.

tiborsimko:
Shows run numbers in the detailed record page for CMS datasets.

Adds run numbers for CMS datasets. Closes #3383.

Cleans other CMS fixture files fixing cosmetic field sorting and white space issues.

Adds new field run_numbers into record JSON Schema. Closes #3383.

Adds forgotton collision_information.type field into the record JSON schema.


cernopendata/templates/cernopendata_records_ui/records/detail.html
```diff
@@ -74,6 +74,14 @@ <h6 class="card-title">
                                 <h2>Description</h2>
                                 <p>{{record.abstract.description | safe}}</p>
                                 {{ link_list(record.abstract.links) }}
+                                {% if record.run_numbers %}
```
tiborsimko, line 77: @katilp We can see live about the best visual presentation of the run number information.  Currently it looks like:

run_numbers

katilp: I'd rather hide it by default and keep it in the metadata JSON only for now. 

It somehow takes importance from the validated run filter (the previous link) and - while it is valid information and certainly something we want to keep track of - it does not add information that users could or would actually use.

tiborsimko: OK, removed.  Just for information, this was the corresponding Jinja snippet:

```diff
diff --git a/cernopendata/templates/cernopendata_records_ui/records/detail.html b/cernopendata/templates/cernopendata_records_ui/records/detail.html
index 84ebd54f..1ba3bd02 100644
--- a/cernopendata/templates/cernopendata_records_ui/records/detail.html
+++ b/cernopendata/templates/cernopendata_records_ui/records/detail.html
@@ -74,6 +74,14 @@
                                 <h2>Description</h2>
                                 <p>{{record.abstract.description | safe}}</p>
                                 {{ link_list(record.abstract.links) }}
+                                {% if record.run_numbers %}
+                                  <p>This dataset was taken during the following run numbers:
+                                  {% for run_number in record.run_numbers %}
+                                     {{run_number}}
+                                     {{ "," if not loop.last else "" }}
+                                  {% endfor %}
+                                  </p>
+                                {% endif %}
                             </div>
                         </div>
                         {% endif %}
```

katilp (approved):

Issue #3425, opened 2023-09-12
records: remove collision information and run period information from CMS HLT path snippet records


tiborsimko:
Similarly as we did for 2015, remove collision information and run period information from CMS HLT path snippet records from all past CMS YYYY releases.

PR #3426, opened 2023-09-13
records: CMS HLT 2011 2012 trigger path updates

tiborsimko:
Removes collision information and run period information from CMS 2011 2012 HLT trigger path records. Closes cernopendata/opendata.cern.ch#3425.


tiborsimko:
P.S. Note that 2015 trigger path snippets were updated as part of the HI release pull request. #3361

katilp (approved):

PR #3444, opened 2023-09-26
Introduce JADE into the OpenData

andriish:
Introduce JADE into the OpenData:
- Add Jade logos and create simple documents with basic info on JADE

@tiborsimko, please comment on the parts that should be removed from the initial MR.


andriish:
Hi @tiborsimko, many thanks! It looks great to me !

tiborsimko (approved):
Thanks, amended as we spoken "live", kept some FIXMEs, merging and deploying on QA...

PR #3461, opened 2023-11-13
cms_records: Fixed the energy of a record

psaiz:


tiborsimko (commented):
This is an interfill dataset, i.e. when no beam is present in the tunnel, hence it is correct that the energy is 0 TeV.

PR #3472, opened 2023-11-29
records: enrich jade-logbooks.json

rdebrand:


tiborsimko (approved):
Thanks, added also Andrii to the list of authors for his previous commit.

PR #3481, opened 2023-11-30
search: show more year values in the year facet

### LINKED ISSUE #3480: facets: show more years
**Curent behaviour**

With the addition of JADE records, the default year facet for empty searches is covering only early years:

years

**Expected behaviour**

More years should be shown by the facet all the way up to the recent years.

tiborsimko:
After the addition of old JADE records, we have much larger year span, and so have to show more year values in the facet. Closes #3480.

### DIFF
--- a/cernopendata/config.py
+++ b/cernopendata/config.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of CERN Open Data Portal.
-# Copyright (C) 2017, 2018, 2022 CERN.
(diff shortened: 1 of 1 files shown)


cernopendata/config.py
```diff
@@ -293,6 +293,7 @@
                 order=dict(_term='asc'))),
             year=dict(terms=dict(
                 field='date_created.keyword',
+                size=70,
```
tiborsimko, line 296: Before:

years

After:

years-new

rdebrand (approved):
Looks good!

Issue #3095, opened 2021-03-01
search: most recent news


tiborsimko:
The most recent news link is displayed on the home page. It is currently not giving the most recent order, see:

https://opendata-dev.cern.ch/search?q=&f=type%3ANews&l=list&order=asc&p=1&s=10&sort=mostrecent

What we should do is to order the news by the `created` field that all records have:

```console
$ grep created cernopendata/modules/fixtures/data/docs/*/*.json                                                                                
cernopendata/modules/fixtures/data/docs/alice-releases-educational-datasets-2014/alice-releases-educational-datasets-2014.json:    "created": "2014-11-20",
cernopendata/modules/fixtures/data/docs/atlas-13tev-open-data-release-2020/atlas-13tev-open-data-release-2020.json:      "created": "2020-02-11",
cernopendata/modules/fixtures/data/docs/atlas-higgs-machine-learning-challenge-2015/atlas-higgs-machine-learning-challenge-2015.json:    "created": "2015-02-17",
cernopendata/modules/fixtures/data/docs/atlas-open-data-resources-2016/atlas-open-data-resources-2016.json:    "created": "2016-07-29",
cernopendata/modules/fixtures/data/docs/cern-open-data-policy-for-lhc-experiments/cern-open-data-policy-for-lhc-experiments.json:    "created": "2020-12-11",
cernopendata/modules/fixtures/data/docs/cms-completes-2010-2011-pp-data/cms-completes-2010-2011-pp-data.json:    "created": "2020-08-27",
cernopendata/modules/fixtures/data/docs/cms-observing-higgs-2017/cms-observing-higgs-2017.json:    "created": "2017-12-20",
cernopendata/modules/fixtures/data/docs/cms-releases-first-batch-of-data-2014/cms-releases-first-batch-of-data-2014.json:    "created": "2014-11-20",
cernopendata/modules/fixtures/data/docs/cms-releases-heavy-ion-data/cms-releases-heavy-ion-data.json:    "created": "2020-12-21",
cernopendata/modules/fixtures/data/docs/cms-releases-new-batch-of-data-2016/cms-releases-new-batch-of-data-2016.json:    "created": "2016-04-22",
cernopendata/modules/fixtures/data/docs/cms-releases-open-data-for-machine-learning-2019/cms-releases-open-data-for-machine-learning-2019.json:    "created": "2019-07-18",
[...]

PR #3484, opened 2023-12-05
search: order most recent news by created field

psaiz:
Addresses https://github.com/cernopendata/opendata.cern.ch/issues/3095


cernopendata/config.py
```diff
@@ -301,7 +301,7 @@
             order=1,
         ),
         'mostrecent': dict(
-            fields=['recid'],
+            fields=['created', 'recid'],
```
zzacharo, line 304: do we need the recid? I guess this is the field to run the sorting upon right?

zzacharo (approved):

PR #3490, opened 2023-12-06
LHCb Open Data Release Full Run I 

MindaugasSarpis:
Added new lines at the end of the json files for records and docs. Capitalized stream names in docs.


tiborsimko (approved):
Thanks, I rebased against latest "qa" and kept only the docs changes.

PR #3504, opened 2023-12-20
Merge branch 'qa'

tiborsimko:
* qa:
      docs: LHCb release announcement
      records: add DOIs to new LHCb streams
      records: add file information to new LHCb streams
      templates: announce 4 PB on the home page
      docs: fix Podman build instructions
      docs: add information about cleaning workspace
      docs: fix LHCb stream name case
      search: show more year values in the year facet
      records: complete JADE logbook descriptions
      records: additional LHCb Run1 open data streams
      docs: additional LHCb Run1 stripping information
      records: update JADE notebook 01
      docs: update DEVELOPING guide about SELinux
      docs: initial JADE open data skeleton
      records: CMS 2016 luminosity files
      records: CMS 2016 luminosity record
      records: CMS 2016 validate runs files
      records: CMS 2016 validated runs records


psaiz:
I don't seem to be able to put the comments in the `Files Changed view`, so I'll post them here:
1. For the facets of years, the aggregation now will display up to 70 different years. Looking in qa, at the moment there are 19. If we were to display 70, that would take a huge space. If we expect that many values, it might be better to put a slide, 
Screenshot 2023-12-20 at 16 28 02, similar to the one in videos:
Screenshot 2023-12-20 at 16 29 58

2. Some existing entries are modified. There is a CMS one for instance where a field is removed. Does that work correctly with the`fixtures --mode create-or-update`, or does it require to delete all the entries and reload them

tiborsimko:
>     1. For the facets of years, the aggregation now will display up to 70 different years. Looking in qa, at the moment there are 19. If we were to display 70, that would take a huge space. If we expect that many values, it might be better to put a slide

Yes, we may need to do this eventually, e.g. if one day we shall have the LEP data on board, then we shall have more time span. For the time being I have put the same same value as in "qa", since this is an inter-branch merge. If we'd like to make a year facet slider, that would be best done via a new feature branch. Anyway I think there won't be much more years any time soon until we get some LEP data on board, so I guess we should be mostly OK for some time to come.

>     2. Some existing entries are modified. There is a CMS one for instance where a field is removed. Does that work correctly with the`fixtures --mode create-or-update`, or does it require to delete all the entries and reload them

Hmm, there were no field removals in the merge request, only additions. For example, if you do: 

```console
$ git log master.. --numstat | grep 'fixtures/data/records' | grep -v LHCb
41      41      cernopendata/modules/fixtures/data/records/jade-logbooks.json
1       1       cernopendata/modules/fixtures/data/records/jade-logbooks.json
547     0       cernopendata/modules/fixtures/data/records/jade-author-list.json
926     0       cernopendata/modules/fixtures/data/records/jade-logbooks.json
74      0       cernopendata/modules/fixtures/data/records/jade-preservation-author-list.json
39      0       cernopendata/modules/fixtures/data/records/jade-tools.json
41      0       cernopendata/modules/fixtures/data/records/cms-luminosity-information-detailed.json
54      0       cernopendata/modules/fixtures/data/records/cms-luminosity-information-detailed.json
24      0       cernopendata/modules/fixtures/data/records/cms-validated-runs-ppRun2.json
112     0       cernopendata/modules/fixtures/data/records/cms-validated-runs-ppRun2.json

$ git log master.. -p -- cernopendata/modules/fixtures/data/records/cms-luminosity-information-detailed.json
$ git log master.. -p -- cernopendata/modules/fixtures/data/records/cms-validated-runs-ppRun2.json 
```

You'll see that only additions have been done for CMS records.

psaiz:
Thanks Tibor. Check for instance:
Screenshot 2023-12-21 at 10 59 37

That field is removed, isn't it?

tiborsimko:
> That field is removed, isn't it?

Ah, yes, this is "docs", I checked the CMS "data records" only. The removal of the feature field works OK, it only influences the position at which  the news items are announced on the home page. That is to say, we "accumulate" them on PROD when deploying without cleaning the DB first: https://opendata.cern.ch/#news and when someone installs from scratch, we display only three news items. I consider this acceptable since the featured news items are ordered chronologically anyway and are only shown on the bottom of the page, hence not influencing the display much.

psaiz (approved):

PR #3511, opened 2024-01-10
Legacy

psaiz:
Support for legacy url. This PR depends on a PR on invenio-search-ui. It should be merged only after that one


scripts/get_facet_hierarchy.py
```diff
@@ -1,29 +0,0 @@
-#!/usr/bin/env python
```
psaiz, line 1: The functionality of this file was moved to the views.py. Instead of having to run it, and then copy the output into a hardcoded list of facets into the config, the facets will be read from opensearch the first time that they are needed.

zzacharo (approved):

PR #3513, opened 2024-01-12
dependencies: Set range versions for the dependencies

psaiz:
Set the version for the invenio dependencies as a range
Increase the minimum dependency on invenio-search-ui
Increase version number of xrootd to be compatible with the Dockerfile

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,36 +67,36 @@ setup_requires = [
 
 install_requires = [
     # General Invenio dependencies
-    "invenio-app==1.3.0",
-    "invenio-base==1.3.0",
-    "invenio-config==1.0.3",
+    "invenio-app>=1.3.0,<2.0.0",
+    "invenio-base>=1.3.0,<2.0.0",
+    "invenio-config>=1.0.3,<2.0.0",
     # Custom Invenio `base` bundle
-    "invenio-assets==3.0.0",
-    "invenio-accounts==3.0.0",
     "importlib-metadata==4.13.0",
-    "invenio-logging[sentry]==1.3.0",
+    "invenio-accounts==3.1.0",
+    "invenio-assets>=3.0.0,<4.0.0",
+    "invenio-logging[sentry]>=1.3.0,<2.0.0",
     "invenio-rest==1.2.8",
-    "invenio-theme==2.5.7",
+    "invenio-theme>=2.5.7,<3.0.0",
     # Custom Invenio `metadata` bundle
-    "invenio-indexer==2.1.0",
-    "invenio-jsonschemas==1.1.3",
-    "invenio-pidstore==1.3.1",
+    "invenio-indexer>=2.1.0,<3.0.0",
+    "invenio-jsonschemas>=1.1.4,<2.0.0",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -67,36 +67,36 @@
...
-    "invenio-records-files==1.2.1",
     # Custom Invenio `postgresql` bundle
-    "invenio-db[versioning,postgresql]==1.1.0",
+    "invenio-db[versioning,postgresql]>=1.1.0,<2",
     # Custom Invenio `opensearch` bundle
-    "invenio-search[opensearch2]==2.1.0",
+    "invenio-search[opensearch2]>=2.1.0,<3",
     # Specific Invenio dependencies
-    "invenio-xrootd>=1.0.0a6",
+    "invenio-xrootd>=1.0.0a6,<2",
```
zzacharo, line 99: Maybe because this is an alpha I would be stricter and pin it....wdyt?

setup.py
```diff
@@ -67,36 +67,36 @@
 
 install_requires = [
     # General Invenio dependencies
-    "invenio-app==1.3.0",
-    "invenio-base==1.3.0",
-    "invenio-config==1.0.3",
+    "invenio-app>=1.3.0,<2",
```
zzacharo, line 70: cosmetic: 
```suggestion
    "invenio-app>=1.3.0,<2.0.0",
```

setup.py
psaiz: Yes, sure.

setup.py
psaiz: Ok, changed all of them

setup.py
```diff
@@ -67,36 +67,36 @@
...
+    "invenio-config>=1.0.3,<2.0.0",
     # Custom Invenio `base` bundle
-    "invenio-assets==3.0.0",
-    "invenio-accounts==3.0.0",
     "importlib-metadata==4.13.0",
-    "invenio-logging[sentry]==1.3.0",
+    "invenio-accounts>=3.0.0,<4.0.0",
+    "invenio-assets>=3.0.0,<4.0.0",
+    "invenio-logging[sentry]>=2.1.0,<3.0.0",
     "invenio-rest==1.2.8",
```
psaiz, line 78: This one was not changed. The latest version requires a new mayor of invenio-logging, and the sentry dependencies are different

setup.py
```diff
@@ -67,36 +67,36 @@
...
+    "invenio-search-ui>=2.8.3,<3.0.0",
     # Custom Invenio `files` bundle
-    "invenio-previewer==2.0.1",
+    "invenio-previewer>=2.0.1,<3.0.0",
+    "invenio-records-files>=1.2.1,<3.0.0",
     "jupyter-client==7.1.0",
     "pluggy==0.13.1",
-    "invenio-records-files==1.2.1",
     # Custom Invenio `postgresql` bundle
     "invenio-db[versioning,postgresql]==1.1.0",
```
psaiz, line 95: This one was not changed. The newer versions require a new mayor of Flask-SQLAlchemy, and I got scared by that one

zzacharo (approved):

PR #3516, opened 2024-01-16
facets: display the facets on the mobile application

psaiz:
Right now, the button to display the facets on a mobile display cannot be clicked. There is a ` div` overlapping. This PR ensures that there is no overlap between the different elements

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
+++ b/cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
@@ -36,10 +36,6 @@
   min-height: 75px;
 }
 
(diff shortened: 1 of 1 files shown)


zzacharo (commented):
Are you sure that this css rule is used only on mobile rendering?

psaiz:
The `ui relaxed` seems to identify the central part. Then, it takes out the `results` div, and the only four wide column that I see there is the `class="mobile only tablet only four wide mobile four wide tablet column"`. On top of that, I'm following Javier's suggestion of avoiding negative margins, that can mess up the display.

Issue #3469, opened 2023-11-22
Conflicts in the attributes of the opensearch schemas


psaiz:
The schema of opendata defines 3 objects (records, glossary and docs). All these obejcts are stored under indices like `record-<type>`. The searches are done in all  the indices.

There are some conflicts in the attributes defined for those objects:

* Experiment:
```
Type | Index names
-- | --
keyword | records-docs-v1.0.0-1700664727, records-record-v1.0.0-1700664727
object | records-glossary-term-v1.0.0-1700664727
```
* Collections:
```
keyword | records-record-v1.0.0-1700664727
-- | --
object | records-docs-v1.0.0-1700664727, records-glossary-term-v1.0.0-1700664727
```

It would be better to be consistent, and store the attribute with the same type for all the objects

psaiz:
Most of the issues have been addressed. The one left is for `collections`. That one is a bit trickier, so it deserves its own PR

psaiz:
For the record, the one of `experiment` has also been fixed (it used to be an array of dictionaries for glossaries, and an array for docs and records)

PR #3528, opened 2024-01-22
mappings: Store the title as tokens as well. Remove duplicate

psaiz:
Addresses #3469

### DIFF
--- a/cernopendata/mappings/os-v2/records/record-v1.0.0.json
+++ b/cernopendata/mappings/os-v2/records/record-v1.0.0.json
@@ -12,7 +12,8 @@
           "type":"keyword"
         },
         "title": {
-            "type": "keyword"
+            "type": "keyword",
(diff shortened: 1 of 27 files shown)


cernopendata/mappings/os-v2/records/record-v1.0.0.json
```diff
@@ -123,26 +127,5 @@
           }
         }
       }
-    },
-    "settings": {
```
zzacharo, line 127: any benefit from removing the analyzer over the `tokens` syntax out of curiosity?

zzacharo (approved):

cernopendata/mappings/os-v2/records/record-v1.0.0.json
psaiz: The default analzyer was already doing that. I wanted to keep thigns simple whenever possible

Issue #3496, opened 2023-12-11
CMS: prepare the SW record for PFNanoProducerTool


katilp:
Prepare the SW records for[ PFNanoProducer](https://github.com/cms-opendata-analyses/PFNanoProducerTool) code, with the json fields as in [trigger sw example records](https://github.com/cernopendata/opendata.cern.ch/blob/master/cernopendata/modules/fixtures/data/records/cms-trigger-examples.json) with the file name cms-tools-pfnanoproducer.json.

Similar to #3495

The derived data records (see https://github.com/cernopendata/data-curation/issues/212) will refer to this record

jmhogan:
@katilp @tiborsimko I've made a release of the tool: https://github.com/cms-opendata-analyses/PFNanoProducerTool/releases/tag/v1.0

katilp:
@tiborsimko Can you attach the zip to the [record](https://opendata.cern.ch/record/12504) so that we can close this issue? Thanks!

PR #3531, opened 2024-01-23
records: CMS pfnano/nanoaod updates

### LINKED ISSUE #3349: CMS: change the current "nanoaod" tags to custom-nanoaod or similar
As mentioned in #2822 the current nanoaod files which are not official nanoaod's should have a different tag.
Remember that POET-nano (#3279) and Run1-nano (#3280) will be uploaded to the derived data.

Possibilities:
- custom-nanoaod for all of them
- differentiate for poet-nanoaod and run1-nanoaod

It would be possible to distinguish them also without as they are of `type` "Derived" and not "Collision" or "Simulated" but still best to be explicit. The official nano will not be "Derived" but either "Collision" or "Simulated".

A side remark, whether derived data are collision or simulated, needs to be extracted from the parent dataset, not really practical but probably no use case so far.

katilp:
Close #3349
Addresses #3496 

For the SW record, to do:
- tag the repository
- upload a zip file from that tag


tiborsimko (approved):
Thanks, works nicely, as discussed merging without PFNanoProducerTool tarball which will be added later once a release is created on GitHub.

Issue #3546, opened 2024-01-30
Style: Add some space between the search bar and the top


psaiz:
The search bar touches theheader of the page. It will be better to have some space there.

![Screenshot 2024-01-30 at 12 01 30](https://github.com/cernopendata/opendata.cern.ch/assets/12909623/cdca42f7-d648-42d6-8507-d8165330a860)

PR #3548, opened 2024-01-30
style: Add a line between the header and the search bar

psaiz:
Solves #3546

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
+++ b/cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
@@ -15,6 +15,10 @@
   }
 }
 
+#main_search_div {
+    padding-top: 20px;
+}
+
 .item {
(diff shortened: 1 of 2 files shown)


psaiz:
Screenshot 2024-01-30 at 17 51 10

cernopendata/modules/theme/assets/semantic-ui/scss/search.scss
```diff
@@ -15,6 +15,10 @@
   }
 }
 
+.ui.fluid.container {
```
zzacharo, line 18: maybe give a specific `id` as well so it doesn't affect all fluid containers unless you intend to!

psaiz: Thanks for the hint. Is this better?

zzacharo (approved):

PR #3549, opened 2024-01-30
schema: Ensure that the field 'experiment' is an array for all entries

psaiz:


psaiz:
Screenshot 2024-01-30 at 15 33 50

Now, a document can have the tags of multiple experiments. Note that the empty entry in the experiment filter has dissappeared

zzacharo (approved):

tiborsimko:
Seems you have forgotten to update the badge display part: note the array:

badges

Issue #3537, opened 2024-01-24
CMS: add records for the updated VM


katilp:
(From https://github.com/cernopendata/data-curation/issues/124)

A new VM is provided with slc7 "CMS Shell". Build new records for 
Image:
- similar to https://opendata-qa.cern.ch/record/257
- recid 258
- image version 1.6.0
Contextualization files
- similar to https://opendata-qa.cern.ch/record/257
- recid 259

katilp:
Contextualization files to be copied to

/eos/opendata/cms/environment/2016

katilp:
Closed through #3552

Issue #3495, opened 2023-12-11
CMS: prepare the SW record for NanoAODRun1 usage examples


katilp:
Information and code currently in https://twiki.cern.ch/twiki/bin/view/CMSPublic/NanoAODRun1Examples
It should be brought to a repository under https://github.com/cms-opendata-analyses/
To follow the naming scheme, the repository should be called `NanoAODRun1ExampleAnalysis`

Then, prepare a SW record to refer to this repository. 
The extensive information can stay on the GitHub repository readme, and the portal record can brief explanation to avoid duplication.

jmhogan:
Updating this one: github repository exists in our area now and has been tested. I've got a record in the works but not pushed yet

katilp:
@jmhogan do you already recid for this? Asking because of [this](https://github.com/cernopendata/data-curation/blob/cms-nancy-derived-datasets/cms-derived-data/config.yaml#L42)

jmhogan:
@katilp I will plan to use 12506 for this and 12507 for nanoAODtools

katilp:
@jmhogan We might want to make https://opendata-qa.cern.ch/docs/cms-getting-started-nanoaod to point to 12507 then. In particular, if we make a cms-opendata-analyses repo out of https://github.com/cms-nanoAOD/nanoAOD-tools/ 
If we do so, we need to handle eventual updates if the original repo gets updated.

jmhogan:
@katilp sounds good, I like that plan for the "how to use" section.

And yes, I will relink in the nano instructions.

katilp:
Closing, the [record](https://opendata.cern.ch/record/12507) exists and has the zip of repo attached

Issue #3465, opened 2023-11-20
CMS - github repositories deleted for 2010 examples recids 101 and 200


katilp:
As reported by @jmhogan the github repositories for http://opendata.cern.ch/record/101 and http://opendata.cern.ch/record/101 have been deleted.
For 101, it is most likely just a naming change w.r.t the 2011 example, see https://github.com/cms-opendata-analyses/OutreachExercise2011/pull/1/files t.b.c
For 200, it is most liley recoverable from https://github.com/sreev/pattuples2010

The other SW records not pointing to cms-opendata-analyses or cms-opendata-validation:
- http://opendata.cern.ch/record/552 : https://github.com/tpmccauley/dimuon-filter
- http://opendata.cern.ch/record/553: https://github.com/jmduarte/SUSYBSMAnalysis-RazorFilter
- http://opendata.cern.ch/record/550: https://github.com/cms-outreach/ispy-analyzers/tree/Run2010B
- http://opendata.cern.ch/record/551: https://github.com/cms-outreach/ispy-analyzers/tree/Run2011A
- http://opendata.cern.ch/record/212: points to http://opendata.cern.ch/record/50 that points to a missing IPPOG page https://ippog.org/resources/2012/cms-hep-tutorial
- http://opendata.cern.ch/record/556: https://github.com/cms-outreach/ispy-analyzers/tree/Run2012

Issue #3452, opened 2023-10-17
CMS running documentation to-do list for 2023-2024.


jmhogan:
CMS will do a big documentation update in Autumn '23 -- Winter '24. This encompasses pages on the Portal, the Open Data Guide, and public CMS TWikis

Open Data Portal pages
- [ ] [Guide to research use](https://opendata.cern.ch/docs/cms-guide-for-research): update format (mostly "exploring" sections) from years --> file types, adding 2015 MiniAOD.
- [ ] [Guide to trigger system](https://opendata.cern.ch/docs/cms-guide-trigger-system): update with clear methods of accessing trigger info in AOD, MiniAOD. 
- [ ] [Guide for event production](https://opendata.cern.ch/docs/cms-guide-event-production): broaden to include examples for MiniAOD production
- [ ] Getting Started pages:
   - [ ] Morph [2010](https://opendata.cern.ch/docs/cms-getting-started-2010), [2011-2012](https://opendata.cern.ch/docs/cms-getting-started-2011), and [2015](https://opendata.cern.ch/docs/cms-getting-started-2015) pages into a "Getting Started with CMS (AOD Data 2010-2012)(MiniAOD Data 2015-on)" (and adapt to future file types). Pages should include:
      - Docker container instructions for each year on the page, hopefully in tabs
      - Data contents and how to list them (differs by file type). Link to Open Data Guide
      - EDAnalyzer example (differs by file type)
      - POET example for filtering and analyzing (for AOD and MiniAOD pages, using relevant branches)
      - Links and references.  
   - [ ] Test POET 2011 branch with 2010 container and file to look for issues using this analyzer.
   - [ ] Get tab support or find another solution #3451 

Pages needing only link checks and small updates with any future releases: 
- [ ] https://opendata.cern.ch/docs/cms-guide-docker
- [ ] https://opendata.cern.ch/docs/about-cms
- [ ] https://opendata.cern.ch/docs/cms-guide-pileup-simulation

Duplicate records of documentation pages to remove after checking for any links that point to them:
- [ ] https://opendata.cern.ch/record/70 --> https://opendata.cern.ch/docs/cms-guide-pileup-simulation
- [ ] https://opendata.cern.ch/record/71 --> https://opendata.cern.ch/docs/cms-simulated-dataset-names
- [ ] https://opendata.cern.ch/docs/cms-guide-troubleshooting --> Make this point to the CMS Open Data Guide FAQ for easier maintenance.

[...]

PR #3551, opened 2024-01-31
Add CMS Getting Started for AOD, MiniAOD, NanoAOD

jmhogan:
This PR adds docs pages for CMS AOD, MiniAOD, and NanoAOD formats, addressing several tasks in #3452.

* cernopendata/modules/fixtures/data/docs/cms-getting-started-aod/ will replace the cms-getting-started-2010 and cms-getting-started-2011 pages
* cernopendata/modules/fixtures/data/docs/cms-getting-started-miniaod/ will replace the cms-getting-started-2015 page
* cernopendata/modules/fixtures/data/docs/cms-getting-started-nanoaod/ is a new addition

Please forgive the inevitable first-time-contributor gaffs! 

Some of the contents of these pages refer to records that will be made in the future. Hopefully all links are currently valid, but several should updated with new records before new CMS documentation are records are put into production.  (#3495, #3496, #3537, #3281, #3279)

This slightly addresses #3465 by no longer linking to these older records as examples.


cernopendata/modules/fixtures/data/docs/cms-getting-started-miniaod/cms-getting-started-miniaod.json
katilp, line 1: If I'm not wrong the standard mkedanlz does not produce code that works in CMSSW_10_6_30. It is not the container feature but the release feature: it includes objects that are not in the standard miniaod (and also does not produce the config file). A mention of that could be added. A work-through attempt is reported in the GitLab issue in https://gitlab.cern.ch/cms-cloud/cmssw-docker-opendata/-/issue, I'll add a link, I can't connect there now)

The current MiniAOD mkedanlz instructions work in 7_6_7 container, so that correct.

katilp (approved):
Looks great! After a quick reading, the only point I have is to add a mention that the 2016 MiniAOD mkedanlz  does not work as such.

jmhogan:
Ok, @katilp do we not have the same "hack" in that container that we do in 767? If not, it would be useful to add it in the upcoming refreshes...

katilp:
> Ok, @katilp do we not have the same "hack" in that container that we do in 767? If not, it would be useful to add it in the upcoming refreshes...

This a different problem, it is not just to make it work: it works but provides a non-working example
So some heavier hacking would be needed. I'd go for a git repository with a basic equivalent skeleton

jmhogan:
Ah, ok I remember now -- will check tonight and make a working EDAnalyzer
in a repo

And of course try to make the checks pass

jmhogan:
@katilp new commits change up the EDAnalyzer example in MiniAOD: 
* 2015 container pointed to mkedanlzr
* all others (2016 container, either VM) pointed to clone this: https://github.com/cms-opendata-analyses/MiniAnalyzer_CMSSW_10_6_30
Tested in the 2016 container -- the analyzer itself is a direct copy of mkedanlzr in the 7_6_7 container. I haven't tested the VMs but would be surprised if it doesn't work as stated. 

I also (for now) have removed the note that cmsenv should not be done in the container. 

I need to triple check the rendering tomorrow on my desktop where I've worked up until now. I'm getting HTTPS Connection refused messages on my laptop when I run any of the instance cleaning or populating commands, though build, up, and down commands are successful.

To be seen whether I can pass the checks...

jmhogan:
Ok, ready to go on my side.

katilp:
Perfect, thanks!

tiborsimko:
@jmhogan @Ari-mu-l Thanks, I'll rebase the branch and squash all the O(300) commits together into one and declare you both as co-authors. Hope that's OK? (CC @katilp)

jmhogan:
Thanks @tiborsimko that sounds good -- it's too bad that the Github Preview is so helpful for markdown editing compared to rendering the site, very tempting to just hit that "commit" button. We discussed how to avoid so many commits in the future.

tiborsimko:
> Thanks @tiborsimko that sounds good -- it's too bad that the Github Preview is so helpful for markdown editing compared to rendering the site, very tempting to just hit that "commit" button. We discussed how to avoid so many commits in the future.

Thanks, I'll write over your PR then with everything squashed. BTW no problem in having many small commits whilst developing! It can be very useful. We are simply cleaning branches before merging as the last step, just to get rid of some intermediate commits when they aren't fully necessary to preserve. This facilitates future updates or bug fix hunting. (See also the [developing guide](https://github.com/cernopendata/opendata.cern.ch/blob/master/DEVELOPING.rst#updating-pull-requests).)

tiborsimko (approved):
Thanks, squashed as discussed, renamed the location of supporting images, added you and Xiaohe as co-authors.

PR #3571, opened 2024-02-07
index: moving from templates to index_templates

psaiz:
This one depends on https://github.com/inveniosoftware/invenio-search/pull/225

The dependencies willl be changed when that one is merged, so that we require the latest version of invenio-search


psaiz:
This is in the infrastructure now. Closing

Issue #3563, opened 2024-02-02
Give higher weight to the title on the searches


psaiz:
At the moment, the searches go through all the fields of the records. It would be better that the title field gets a higher weight than the body, for instance.

katilp:
An example:
I would like to get the "Getting Started with YYYY CMS Open Data" first with [this search](https://opendata.cern.ch/search?q=Getting%20started&f=experiment%3ACMS&f=type%3ADocumentation%2Bsubtype%3AGuide&l=list&order=asc&p=1&s=10&sort=bestmatch)

psaiz:
![Image](https://github.com/cernopendata/opendata.cern.ch/assets/12909623/76d305d1-1808-42fb-81c1-a448bf953595)

psaiz:
Deployed!

PR #3578, opened 2024-02-12
search: Increase the weight of the title in the results of a search.

psaiz:
Solves #3563


psaiz:
Moving to cernopendata-portal

PR #3579, opened 2024-02-12
layout: Rename the Download to Download index.

### LINKED ISSUE #2618: templates: download index
When offering download for index files, the button should say "Download index" rather than "Download" in order to avoid misunderstandings.  This would give "List files - Download index".

Example: http://opendata-dev.cern.ch/record/12100

psaiz:
Closes #2618

### DIFF
--- a/cernopendata/modules/theme/assets/semantic-ui/js/records/components/FileTable.js
+++ b/cernopendata/modules/theme/assets/semantic-ui/js/records/components/FileTable.js
@@ -60,6 +60,7 @@ export default function FileTable({ items }) {
                   },
                 }
               : { href: getFileUri(file.key) };
+          const label= file.type=='index.txt' ? 'Download index' : 'Download';
           return (
             <Table.Row key={file.version_id}>
               <Table.Cell className="filename-cell">{file.key}</Table.Cell>
@@ -80,7 +81,7 @@ export default function FileTable({ items }) {
                   </Button>
                 )}
                 <Button as="a" icon size="mini" primary {...downloadProp}>
-                  <Icon name="download" /> Download
+                  <Icon name="download" /> {label}
                 </Button>
               </Table.Cell>
             </Table.Row>


cernopendata/modules/theme/assets/semantic-ui/js/records/components/FileTable.js
```diff
@@ -80,7 +80,7 @@ export default function FileTable({ items }) {
                   </Button>
                 )}
                 <Button as="a" icon size="mini" primary {...downloadProp}>
-                  <Icon name="download" /> Download
```
tiborsimko, line 83: The button label should be dynamic:

- for data files attached via file indexes, such as https://opendata.cern.ch/record/1, the button should say "Download index";

- for data files attached directly, such as https://opendata.cern.ch/record/5500, the button should say "Download".

(The index files can be recognised by having `files.type` field value "index.json" or "index.txt" in the record JSON.)

katilp: @psaiz The dynamic button label would be indeed needed: we have had to explain this separately to the OD users as they got confused. Now we will have more files that are small enough to be downloaded so this is even more important.

cernopendata/modules/theme/assets/semantic-ui/js/records/components/FileTable.js
```diff
@@ -60,6 +60,7 @@ export default function FileTable({ items }) {
                   },
                 }
               : { href: getFileUri(file.key) };
+          const label= file.type=='index.txt' ? 'Download index' : 'Download';
```
zzacharo, line 63: could that be also `index.json`?

psaiz: THe index.json entries seem to be filtered out at some point (I haven't figure it out where exactly).  See for instance [this entry](https://opendata.cern/record/28003), which contains a normal file and a  file set:
Screenshot 2024-03-27 at 16 27 14

If we look at the entry on opensearch, it has three files:
Screenshot 2024-03-27 at 16 28 00

psaiz:
This has been addressed in the cernopendata_portal repo

Issue #3454, opened 2023-10-23
CMS - Adding cross-section values to the datasets


katilp:
Add the cross-section values to 2015 MC datasets, as computed with [XGenAnalyzer](https://github.com/Ari-mu-l/OpenData/tree/main/GenXSecAnalyzer) by @Ari-mu-I 

The values are in the logs, e.g. https://github.com/Ari-mu-l/OpenData/blob/main/GenXSecAnalyzer/logs/Drell-Yan/xsec_16414.log#L1629-L1643

Follow the guidelines at https://cms-opendata-releaseguide.docs.cern.ch/adding_metadata/

- [ ] Agree with @jmhogan and @Ari-mu-I what are the values to be displayed

- [ ] Extract them from the logs
- [ ] Add them to the metadata, agree with @tiborsimko on the structure
      - (see the [earlier draft in 2015 script](https://github.com/cernopendata/data-curation/blob/master/cms-2015-simulated-datasets/code/dataset_records.py#L339-L348) for the metadata structure, but not that the actual values are not reliable in the location where this code would read them - that's why the new logs have been generated)

jmhogan:
Here's an idea for text on the record pages:

"For pp collisions at X TeV, this sample has a cross section of (TOTAL +/- UNCERTAINTY) pb, calculated using the method described HERE (link to #3455). 

This cross section takes into account a matching efficiency of MATCH and a filtering efficiency of FILTER, based on generator settings and/or filters. If this sample was generated at NLO, it has FRACTION% events with negative weights." 


The README of this Github repository explains what you'll find in the json files: https://github.com/Ari-mu-l/OpenData/tree/main. There is more information available than I think we really need to put on the record pages right now, but we can iterate.

And the jsons are in CERNbox here: /eos/user/s/sxiaohe/OpenData/MC2015/   (Xiaohe is making sure this is public)

jmhogan:
Here's a public CERNbox link for the json files: https://cernbox.cern.ch/s/EHpyrdJet939vGy

nancyhamdan:
I developed the script to extract the cross-section values and add them to the metadata of their corresponding datasets (see [#210)](https://github.com/cernopendata/data-curation/pull/210 in `data-curation` repo) and I discovered the following using the script:
```
Total number of cross-section values json files: 544, Total number of amended datasets: 544
Total number of datasets amended using Format 1: 1
Total number of datasets amended using Format 2: 371
[...]

PR #3581, opened 2024-02-16
records: add xsec values for 2015 MC and update record template

katilp:
Addresses #3454 

- adds cross-section values to 2015 MC records from MC2015 in  https://cernbox.cern.ch/s/EHpyrdJet939vGy 
- modifies the record template `templates/cernopendata_records_ui/records/record_detail.html` so that the values get displayed with the text suggested in #3454

### DIFF
--- a/cernopendata/templates/cernopendata_records_ui/records/record_detail.html
+++ b/cernopendata/templates/cernopendata_records_ui/records/record_detail.html
@@ -103,6 +103,25 @@
         {%include "cernopendata_ispy/ispy_standalone.html"%}
     {% endif %}
 
+    {% if record.cross_section %}
+    <h2>Cross section</h2>
+        {% if record.cross_section.description %}
+            <span>{{record.cross_section.description | safe}}</span>
+            <br>
+        {% else %}
+            <p>For {{record.collision_information.type }} collisions at {{record.collision_information.energy }}, this sample has a cross section of (<span><strong>{{record.cross_section.total_value }}</strong></span> +/- <span><strong>{{record.cross_section.total_value_uncertainty }}</strong></span>) pb, calculated using the method described in <a href="/docs/cms-guide-cross-sections#run2">CMS guide for cross sections</a>.</p>
+        {% endif %}
+        {% if record.cross_section.matching_efficiency %}
+            <p>This cross section takes into account a matching efficiency of <span><strong>{{record.cross_section.matching_efficiency }}</strong></span> and a filtering efficiency of <span><strong>{{record.cross_section.filter_efficiency }}</strong></span>, based on generator settings and/or filters.</p>
+        {% else %}
+            <p>This cross section takes into account a filtering efficiency of <span><strong>{{record.cross_section.filter_efficiency }}</strong></span>, based on generator settings and/or filters.</p>
+        {% endif %}
+        {% if record.cross_section.neg_weight_fraction %}
+            <p>If this sample was generated at NLO, it has <span><strong>{{record.cross_section.neg_weight_fraction }}</strong></span>% events with negative weights.</p>
+        {% endif %}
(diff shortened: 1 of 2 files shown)


katilp:
Note that 
- the matching efficiency is `"matching_efficiency": "1.0" (one occasion), else empty.
- the filter efficiency has a value different from 1 for 58 datasets, else  `"filter_efficiency": "1.000e+00"`

The values are displayed as:

image

jmhogan (approved):
The text and method of choosing which elements are displayed looks good to me.

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
```diff
@@ -103,6 +103,24 @@ <h2 id="previewer">Preview</h2>
         {%include "cernopendata_ispy/ispy_standalone.html"%}
     {% endif %}
 
+    {% if record.cross_section %}
+    <h2>Cross section</h2>
+        {% if record.cross_section.description %}
+            <span>{{record.cross_section.description | safe}}</span>
+            <br>
+        {% else %}
+            <p>For {{record.collision_information.type | safe}} collisions at {{record.collision_information.energy | safe}}, this sample has a cross section of (<span><strong>{{record.cross_section.total_value | safe}}</strong></span> +/- <span><strong>{{record.cross_section.total_value_uncertainty | safe}}</strong></span>) pb, calculated using the method described in <a href="/docs/cms-guide-cross-sections">CMS guide for cross sections</a>.</p>
```
tiborsimko, line 112: Some comments for the template:

- The `| safe` part should not be necessary in many places here, since we won't have HTML in the values. We should probably keep it for the free-text description subfield, but for other subfields it may be probably removed. 
- If some of the subfields may not be always present (say filter efficiency?), we could change the template to the style of `record.get('cross_section', {}).get(''filter_efficiency', "") ` to make sure we have some default value even if a subfield is not present. Would this be useful?

BTW we don't have cross_sections defined in the metadata model. I guess it may be useful? If so, we could add it during rebase.

katilp: We can link to
`<a href="/docs/cms-guide-cross-sections#run2">CMS guide for cross sections</a>`

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
```diff
@@ -103,6 +103,24 @@ <h2 id="previewer">Preview</h2>
...
+        {% endif %}
+        {% if record.cross_section.matching_efficiency %}
+            <p>This cross section takes into account a matching efficiency of <span><strong>{{record.cross_section.matching_efficiency | safe}}</strong></span> and a filtering efficiency of <span><strong>{{record.cross_section.filter_efficiency | safe}}</strong></span>, based on generator settings and/or filters.</p>
+        {% else %}
+            <p>This cross section takes into account a filtering efficiency of <span><strong>{{record.cross_section.filter_efficiency | safe}}</strong></span>, based on generator settings and/or filters.</p>
+        {% endif %}
+        {% if record.cross_section.neg_weight_fraction %}
+            <p>If this sample was generated at NLO, it has <span><strong>{{record.cross_section.neg_weight_fraction | safe}}</strong></span>% events with negative weights.</p>
+        {% endif %}
```
katilp, line 121: Add
<p>This is a leading-order value, and cross sections evaluated at next-to-leading order (NLO) or next-to-next-to-leading order (NNLO) are preferred, if available.</p>

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
tiborsimko: Added.

cernopendata/templates/cernopendata_records_ui/records/record_detail.html
tiborsimko: Added.

tiborsimko (approved):

PR #3552, opened 2024-01-31
records: cms 2016 vm

katilp:
(Closes #3537 )

Add records for
- the new VM image
- the contextualization scripts to build it

To do:

Copy from `/eos/opendata/cms/upload/vm160:`

For the VM record 258

```
CMS-OpenData-1.6.0.ova
```

For the contextualization scripts

```
ucernvm-v4prod.2020.07-1.cernvm.x86_64.hdd
cernvm-script
cms-user-data.txt
opendata-desktop-settings
```

PR #3596, opened 2024-03-05
Cms 2016 vm guide

katilp:
Adds the guide for VM. Has reduced content from the previous versions.

Requires updates 
- in https://github.com/cernopendata/opendata.cern.ch/pull/3552
- in several records and docs for naming:  all `cms-virtual-machine-2016` or `cms-virtual-machine-2016-2018` should become **`cms-virtual-machine-cc7`** (can be replaced all in one, no separate updates needed)


jmhogan (approved):
Looks all good to me

tiborsimko (approved):
Thanks, fixed image URL, rebased and squashed.

PR #3602, opened 2024-03-13
docker: move to PostgreSQL 14.10

tiborsimko:
Changes database version to PostgreSQL 14.10, keeping parity with the recommended DBoD deployment version.


psaiz:
This is part of the infrastructure now

PR #3605, opened 2024-03-21
Add author first names

rdebrand:


rdebrand:
Left the dots in there as there were double initials (e.g. J.J.), also left the et al. after the name.

cernopendata/modules/fixtures/data/records/jade-computing-notes.json
```diff
@@ -98,13 +98,13 @@
     "accelerator": "DESY-PETRA",
     "authors": [
       {
-        "name": "Hagermann"
+        "name": "Hagermann, J."
```
tiborsimko, line 101: The bibliographic standard is to leave the initial without the dots, i.e. "Hagermann, J" format.  This will better compare to what CDS, INSPIRE etc repositories are doing, and will also play more nicely with facets here when the first name is fully known. 

Can you please amend the names to leave out trailing dots everywhere?

cernopendata/modules/fixtures/data/records/jade-computing-notes.json
```diff
@@ -929,7 +929,7 @@
     "accelerator": "DESY-PETRA",
     "authors": [
       {
-        "name": "Allison et al."
+        "name": "Allison, John et al."
```
tiborsimko, line 932: The "et al" part in the name would not be correct. We should have only the full author name there so that the name is searchable uniquely across records regardless of collaborators.

For the "et al" parts, ideally we should expand them, e.g. note 18 shows the full list of authors, so the record here should ideally say:

```yaml
    "authors": [
      {
        "name": "Allison, John"
      },
      {
        "name": "Dittmann, P"
      },
      {
        "name": "Elsen, E"
      },
      {
        "name": "O'Neill, L"
      },
      {
        "name": "Steffen, P"
      },
      {
        "name": "Yamada, S"
      },
    ]
```

Can you please expand all the four instances of "et al" correspondingly?

cernopendata/modules/fixtures/data/records/jade-computing-notes.json
```diff
@@ -206,10 +206,10 @@
     "accelerator": "DESY-PETRA",
     "authors": [
       {
-        "name": "Bowdery"
+        "name": "Bowdery, C.K."
       },
       {
-        "name": "Pryce"
+        "name": "Pryce, J.J."
```
tiborsimko, line 212: In case of two initials, we should use space and leave out the dots, e.g. "Pryce, J J".

Can you please amend this here and elsewhere for people having two initials?

PR #3622, opened 2024-04-03
records: CMS 2016 PFNano file information

tiborsimko:


katilp (approved):
Looks good for the file indexes, thanks!
Would you add processedLumis.json for DoubleEG and JetHT or does it go separately?

katilp (approved):

PR #3623, opened 2024-04-04
scripts: add script to fix records with multiple record buckets

zzacharo:
* the record bucket is deleted only if it empty
* the script assumes that the second bucket is fine


psaiz:
Before adding this one, what do you think about checking to see if we can fix/reproduce the issue that causes the multiple buckets? I'm hoping that [this](https://github.com/cernopendata/cernopendata-portal/pull/14) will help.

If we end up adding the file, I would prefer to put it in the cernopendata-portal repo.

psaiz:
Removing from the repo of the content

PR #3645, opened 2024-06-07
records: ATLAS CERN-EP-2024-132 OmniFold data

tiborsimko:


zlmarshall:
These data are already on Zenodo:
http://doi.org/10.5281/zenodo.11507450

So we should avoid duplication and _not_ also publish them here.

Cheers,
Zach

tiborsimko:
> So we should avoid duplication and _not_ also publish them here.

Fully agreed! I'll add you to the email conversation with Dag about this.

PR #3651, opened 2024-06-26
add basic terms for LEP

schwicke:


cernopendata/modules/fixtures/data/terms/terms.json
```diff
@@ -12746,5 +12746,186 @@
...
+      {
+        "term": "L3"
+      },
+      {
+        "term": "OPAL"
+      }
+    ],
+    "term": [
+      "LEP"
+    ],
```
psaiz, line 12801: In this one, I would add "experiment" : ["ALEPH", "DELPHI", "L3", "OPAL"]. That way, we would have the small boxes with the names of the experiments, like in this screenshot.

It might be good to add them also to any other glossary term that applies to them


Screenshot 2024-06-28 at 17 17 26

cernopendata/modules/fixtures/data/terms/terms.json
```diff
@@ -12746,5 +12746,328 @@
     "type": {
       "primary": "Glossary"
     }
+  },
+  {
+    "anchor": "CERNLIB",
+    "category": "generic",
+    "definition": "CERN Program library. The CERN Program Library is a large collection of general purpose libraries and modules maintained and offered in both source and object code form on the CERN central computers. The two most popular applications based on CERNLIB are PAW and GEANT 3.21. Most of these programs were developed at CERN and are therefore oriented towards the needs of a physics research laboratory that is general mathematics, data analysis, detectors simulation, data-handling etc... applicable to a wide range of problems. ",
```
psaiz, line 12753: For consistency with the reference to PAW, what about dropping the`3.21` ( or at least the `.21`) ?

cernopendata/modules/fixtures/data/terms/terms.json
```diff
@@ -12746,5 +12746,328 @@
...
+    ],
+    "experiment": ["ALEPH", "DELPHI", "L3", "OPAL"],
+    "type": {
+      "primary": "Glossary"
+    }
+  },
+  {
+    "anchor": "ALEPH",
+    "category": "generic",
+    "definition": "ALEPH was one out of four experiments of the LEP collider",
```
psaiz, line 12814: Including the definition of the acronym?

cernopendata/modules/fixtures/data/terms/terms.json
```diff
@@ -12746,5 +12746,328 @@
...
+    ],
+    "experiment": ["ALEPH", "DELPHI", "L3", "OPAL"],
+    "type": {
+      "primary": "Glossary"
+    }
+  },
+  {
+    "anchor": "L3",
+    "category": "generic",
+    "definition": "L3 was one out of four experiments of the LEP collider",
```
psaiz, line 12910: Is L3 an acronym? If so, it would be nice to explain it (for the record, I did look at https://home.cern/science/experiments/l3 , and I couldn't find out why it was called L3 :( )

cernopendata/modules/fixtures/data/terms/terms.json
```diff
@@ -12746,5 +12746,328 @@
...
+    "term": [
+      "ZEBRA"
+    ],
+      "experiment": ["DELPHI", "L3", "OPAL", "JADE"],
+    "type": {
+      "primary": "Glossary"
+    }
+  },
+  {
+    "anchor": "PAW",
```
psaiz, line 12972: For consistency, since PAW is described, what about adding also GEANT?

psaiz (approved):
Couple of minor changes. This could be merged as is, and put the rest of the things later on

cernopendata/modules/fixtures/data/terms/terms.json
schwicke: Hmm, these are two quite different pairs of shoes :) DELPHI didn't use GEANT (to the best I know at least). We can add it once OPAL joins the club I'd say ?

schwicke: OK, I see why you want it. Because I quoted it myself above. Right. Good point.

PR #3655, opened 2024-06-27
add DELPHI manuals

schwicke:


cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
```diff
@@ -0,0 +1,217 @@
...
+  {
+    "abstract": {
+      "description": "Detailed description of the DELPHI conditions database contents, and the tools coming with it."
+    },
+    "accelerator": "CERN-LEP",
+      "collaboration": {
+      "name": "DELPHI collaboration",
+      "recid": "27050"
+    },
+    "collections": [
```
psaiz, line 11: Would you mind removing this field, please (from all the entries)?

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
```diff
@@ -0,0 +1,205 @@
...
+    },
+    "date_created": [
+      "1993"
+    ],
+    "date_published": "2024",
+    "distribution": {
+      "formats": [
+        "pdf"
+      ],
+      "number_files": 3
```
tiborsimko, line 19: If we list the number of files, then it would be good to list also "size" (total in bytes for all attached files) as well, for consistency with other records.

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
```diff
@@ -0,0 +1,205 @@
+[
+  {
+    "abstract": {
+      "description": "Detailed description of the DELPHI conditions database contents, and the tools coming with it."
+    },
+    "accelerator": "CERN-LEP",
+      "collaboration": {
+      "name": "DELPHI collaboration",
+      "recid": "27050"
```
tiborsimko, line 9: The DELPHI collaboration record ID is 80050, here and elsewhere.

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
```diff
@@ -0,0 +1,205 @@
...
+    "license": {
+      "attribution": "CC0"
+    },
+    "publisher": "CERN Open Data Portal",
+    "recid": "27001",
+    "title": "DELPHI database documentation",
+    "type": {
+      "primary": "Supplementaries",
+      "secondary": [
+        "DELPHI manuals"
```
tiborsimko, line 50: Let's call the `type.secondary` value simply "Manual", i.e. without leading DELPHI. This is because the facet values are usually experiment-agnostic, such as "Dataset > Collision" etc. The experiment-specific search filtering is provided by combining two facets, the type facet and the experiment facet. (Note that this is what we did for JADE logbooks for example, they have the value `Supplementaries.Logbook`.) In this way other experiments could add manuals more easily.

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
```diff
@@ -0,0 +1,205 @@
...
+        "checksum": "adler32:63ea0076",
+        "size": 557435,
+        "uri": "root://eospublic.cern.ch//eos/opendata/delphi/documentation/ddb/dplot.pdf"
+      }
+    ],
+    "license": {
+      "attribution": "CC0"
+    },
+    "publisher": "CERN Open Data Portal",
+    "recid": "27001",
```
tiborsimko, line 45: We can use values around 80500 so that we don't fill "old holes" with the new records.  I.e. from 80501 to 80504.

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
```diff
@@ -0,0 +1,205 @@
...
+    "files": [
+      {
+        "checksum": "adler32:9a26470d",
+        "size": 174717,
+        "uri": "root://eospublic.cern.ch//eos/opendata/delphi/documentation/skelana/99_175_prog_239.pdf"
+      },
+      {
+        "checksum": "adler32:4bac9f7a",
+        "size": 264402,
+        "uri": "root://eospublic.cern.ch//eos/opendata/delphi/documentation/skelana/skelana.pdf"
```
tiborsimko, line 85: When I see a file named "skelana.pdf", I consider it to be the canonical version... Whilst "99_175_prog_239" is named a bit opaquely and seems to provide later updates.

What about explaining this in the description?  Currently, the description does not provide much more information than the title:

skelana


It may be good to expand the description phase to say something like: "The Skelana analysis framework version 1 is documented in the file `skelana.pdf`. The later updates for Skelana version 2 are described in file 99_...".

... or perhaps consider renaming files?

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
```diff
@@ -0,0 +1,205 @@
...
+    "type": {
+      "primary": "Supplementaries",
+      "secondary": [
+        "DELPHI manuals"
+      ]
+    }
+  },
+  {
+    "abstract": {
+      "description": "The DELPHI event display manual."
```
tiborsimko, line 103: Cosmetics: the abstract is the same as the title:

delgra

Perhaps there is something more to say for better readability?

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
```diff
@@ -0,0 +1,205 @@
+[
+  {
+    "abstract": {
+      "description": "Detailed description of the DELPHI conditions database contents, and the tools coming with it."
```
tiborsimko, line 4: Cosmetics: Perhaps you may want to enrich the abstract to say that CARGO describes the database management package, DDAP describes the detection description application package, and DPLOT describes the database graphics?

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
schwicke: fixed

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
schwicke: Fixed

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
schwicke: I kept the names are they are since these are consistent with the manuals elsewhere. It's now described in the abstract, as suggested.

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
schwicke: Valid point. Done.

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
schwicke: Done.

cernopendata/modules/fixtures/data/records/delphi-documentation-manuals.json
schwicke: We can do that. Not sure what the added value is though.

tiborsimko: OK, I add them during merge, and fix the record links from the docs pages.

The value is basically to be consistent with data record where we display total size. This could be useful for information (how big the thing is), for size-based search facets, for enriching records with umbrella-checksum-like information...

tiborsimko (approved):
Thanks for all the amendments, enriched as discussed, merging...

Issue #3659, opened 2024-07-01
Automatic citation inspire link for open data records


zlmarshall:
Hi there! 

I wonder if it might be possible to add to all records an automatically generated link to the inspire search for papers referencing them? I remember seeing this in one of the CMS Open Data talks:

https://inspirehep.net/literature?sort=mostrecent&size=25&page=1&q=references.reference.dois%3A10.7483%2FOPENDATA.CMS%2A

And thinking it's a fantastic thing. My notion is that for example on the DOI line of a record could be a link to "Citations"; e.g. on

https://opendata.cern.ch/record/15012

a link to:

https://inspirehep.net/literature?sort=mostrecent&size=25&page=1&q=references.reference.dois%3A10.7483%2FOPENDATA.ATLAS.UXKX.TXBN

Thanks,
Zach

tiborsimko:
Thanks for the great suggestion! 

One concern that comes to mind is that many datasets may not get cited directly, so offering a "static" link may lead to many "zero citation" pages after a user clicks on those, which may lead to some frustrating user experience. For example, we have 57k simulated datasets overall, and I think most of them will have zero direct citations, since they may be of particular niche interest only. 

I may therefore be advantageous to offer the citation link only when we know for sure that the data was cited by a paper. 
We could do this if we periodically synchronise the citation counts from INSPIRE, for example by a background daily sync process, and store this information also on the Open Data portal side. We would then be able to offer "dynamic" links targeted to each record, such as "Cited by 4 papers", "Cited by 17 papers", etc.  Additionally, knowing the citation counts could also be useful in order to offer some additional search interface capabilities, such as "sort by most cited". (Or to display some aggregated citation information by experiment, for example in the "About ATLAS" pages.)

This dynamic way of linking would be probably better than offering static citation links to INSPIRE for the users, but it would come at the price of some technical development, syncing complexity and maintenance. We'll try to think it over in the team and come back, perhaps with a few mock-ups?

psaiz:
Hi Zach, 
Thanks for this suggestion. I'm linking a pull request that will add the link only for those datasets that have something pointing to them.

Let me know what you think about this approach. 

Best,
pablo

zlmarshall:
Hi @psaiz ,

Thanks! You'll notice one addition in my version above: the search is both on doi and page URL. We've found that some folks directly cite the webpage instead of using the URL (even with the DOI clearly visible on the page). Would that be easy for you to add?

[...]

PR #3663, opened 2024-07-14
Adding inspire links to records

zlmarshall:
This should add a citations link with the inspire logo to the records.

Not sure if it's desirable to check if there are citations before providing the link - because of this formatting, it's no overhead to just have the link, but checking for citations would create overhead in generating the pages.

Using the inspire favored icon for the image, happy to change if there's a preferred icon or logo that anyone is aware of.

This is a quick attempt to address https://github.com/cernopendata/opendata.cern.ch/issues/3659


psaiz:
Hey Zach, I'm going to close this one, since it was addressed in https://github.com/cernopendata/cernopendata-portal/pull/44

PR #3679, opened 2024-09-20
docs(developing): adapt documentation after infrastructure code split

### LINKED ISSUE #3678: docs: update DEVELOPING guide following infrastructure code split
Update the DEVELOPING guide following the split of the infrastructure code from the content.

See review notes done in <https://github.com/cernopendata/opendata.cern.ch/pull/3590>

tiborsimko:
Adapts documentation for data curators following up the split of the 
 portal's infrastructure code from the content in two repositories.

 Implements comments from the #3590 pull request review.

 Closes #3678.

### DIFF
--- a/DEVELOPING.rst
+++ b/DEVELOPING.rst
@@ -5,98 +5,308 @@
 .. contents::
    :backlinks: none
 
+This document describes how you can run a local instance of the CERN Open Data
+portal in order to work with the content records and associated documentation.
+
+Prerequisites
+=============
+
+You will need to fork and clone two repositories:
+
+- `opendata.cern.ch <https://github.com/cernopendata/opendata.cern.ch>`_ which
+  contains the open data content;
+
+- `cernopendata-portal <https://github.com/cernopendata/cernopendata-portal>`_
+  which contains the portal infrastructure code.
+
+Please make sure to also install `Docker
+<https://docs.docker.com/get-started/get-docker/>`_ and `Docker Compose
+<https://docs.docker.com/compose/install/>`_ (version 2) that is used for local
+developments.
+
+You will also need to pull Docker container images from `CERN Harbor registry
+<https://registry.cern.ch/>`_. If this is your first time using CERN Harbor,
+you will need to login using a command-line secret that you can obtain at
+`registry.cern.ch <https://registry.cern.ch/>`_ in the right-hand-side menu
+"johndoe > User Profile". Once you obtain and copy your CLI secret, you can
+login to the CERN Harbor registry as follows:
+
+.. code-block:: console
+
+   $ docker login registry.cern.ch -u johndoe
+
 Installation
 ============
 
-This module contains the content of the CERN Open Data instance. It does not contain the code of the portal itself.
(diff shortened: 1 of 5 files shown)


docker-compose-override.yml
```diff
@@ -30,7 +30,7 @@ services:
       - TEMPLATES_AUTO_RELOAD=True
     command: bash -c "/content/scripts/start-server-debug.sh"
     restart: "unless-stopped"
-    image: registry.cern.ch/cernopendata/cernopendata-portal:0.2.5
+    image: registry.cern.ch/cernopendata/cernopendata-portal:0.2.9
```
psaiz, line 33:
```suggestion
    image: registry.cern.ch/cernopendata/cernopendata-portal:prod
```

README.rst
```diff
@@ -8,34 +8,36 @@
...
 About
 -----
 
-This is the source code behind CERN Open Data portal. You can access the portal
-at `http://opendata.cern.ch/ <http://opendata.cern.ch>`_.  The source code uses
-`Invenio <https://inveniosoftware.org/>`_ digital repository framework.
+This repository is part of for the `CERN Open Data portal
+<https://opendata.cern.ch>`_, hosting the source versions of the content,
+notably data records and associated documentation pages. The CERN Open Data
+portal's infrastructure code is living in the sister `cernopendata-portal
```
psaiz, line 17:
```suggestion
portal's infrastructure code is living in the `cernopendata-portal
```

DEVELOPING.rst
```diff
@@ -5,120 +5,383 @@
...
+.. code-block:: console
+
+   $ vim .git/config
+   $ cat .git/config
+   [remote "upstream"]
+       url = git@github.com:cernopendata/opendata.cern.ch
+       fetch = +refs/heads/*:refs/remotes/upstream/*
+       fetch = +refs/pull/*/head:refs/remotes/upstream/pr/*
 
 Understanding repository branches
```
psaiz, line 261: Before the section of the repository branches, it would be nice to have a section with the repository structure, explaining the difference between the different folders (in particular the one of the `skeletons`).

docker-compose-override.yml
tiborsimko: Perhaps there is a container image tag called `prod`, but there is no such git tag in the `cernopendata-portal` repository. That could be problematic; recall that we have been using `latest` in the past, and had to stop it due to local installation being broken. This is what lead to using 0.1.11 docker image in the sources here, with occasional update to later stable versions.

I therefore prefer to use concrete versions (which exist as both git tags and docker image versions, and which are immutable) over something like `latest` or `prod` (which is mutable and which does not exist as a tag or a branch).

README.rst
tiborsimko: Removed here and in two other places in the developing guide.

DEVELOPING.rst
tiborsimko: Added Appendix A explaining the repository structure, and also added Appendix C explaining the git conventional commit style (that is now being checked by the CI).

PR #3682, opened 2024-09-23
add reference to DELPHI detector paper

schwicke:
Add reference to Nuclear Instruments and Methods, A303(1991), 233-276


data/docs/delphi-about/delphi-about.json
```diff
@@ -8,7 +8,7 @@
       "DELPHI"
     ],
     "short_description": {
-      "content": "DELPHI was a Particle Physics experiment at the CERN laboratory in Geneva, Switzerland. It studied the products of electron-positron collisions at the LEP circular accelerator, working at the then highest energies in the world. The main parts are still kept in the original state in the cavern where it was operating and can be visited."
+	"content": "DELPHI was a Particle Physics experiment at the CERN laboratory in Geneva, Switzerland. It studied the products of electron-positron collisions at the LEP circular accelerator, working at the then highest energies in the world. The main parts are still kept in the original state in the cavern where it was operating and can be visited. A description of the detector as it was in 1990 can be found in [this preprint](https://cds.cern.ch/record/212026/files/cer-000124113.pdf), which has been published in [Nuclear Instruments and Methods, A303(1991), 233-276](https://doi.org/10.1016/0168-9002(91)90793-P)."
```
tiborsimko, line 11: Note that "short description" is used only for the search pages, and here Markdown is not supported. This would give:

about

Even if we fix the markup, it seems a bit too long text for the short search page result, and moreover a link there would "compete"  against the usual liking from the short search result hit to the long detailed page.

Rather you may want to move the new paragraph to the detailed page content itself, i.e. https://opendata-qa.cern.ch/docs/about-delphi and its source `data/docs/delphi-about/delphi-about.md`  ?

tiborsimko (approved):
Thanks, changed `(91)` to `%2891%29` in the URL just to be on the safe side...

PR #3684, opened 2024-09-27
fix(records): add JADE VM image as file

tiborsimko:
Attaches JADE VM image to the record as a proper file.for

Removes `eospublichttp.cern.ch` links correspondingly.for

Fixes link to the VM installation documentation page.

Fixes link to the JADE primary data.


data/records/jade-tools.json
```diff
@@ -16,7 +16,7 @@
     ],
     "links": [
       {
-        "url": "https://eospublichttp.cern.ch/eos/opendata/jade/environment/MPP-JADE-CentOS-7-x86_64-DVD-1511.ova"
```
tiborsimko, line 19: @andriish @rdebrand If the VM snapshot image is "final" and won't change, we could attach the file directly to the record, instead of just using a link. This has several advantages, e.g. file backups, file exposure via `cernopendata-client` , etc. Please let me know whether I can proceed with "freezing" this file and attaching it directly?

andriish: Uh, that is an old VM.  With old CERNLIB. I would say "good for now", but I would replace it with a docker in the nearest months/weeks.

andriish:
LGTM! Thank you!

PR #3686, opened 2024-09-30
Create jade-analysis-example.json

rdebrand:
Add an example analysis of jade data with variable names and descriptions.

### DIFF
--- a/cernopendata/modules/fixtures/data/records/jade-analysis-example.json
+++ b/cernopendata/modules/fixtures/data/records/jade-analysis-example.json
@@ -1,0 +1,1071 @@
+[
+    {
+      "abstract": {
+        "description": ""
+      },
+      "accelerator": "DESY-PETRA",
+      "collaboration": {
+      "name": "JADE collaboration",
+      "recid": "26050"
+      },
+      "collections": [
+        "JADE-Analysis"
+      ],
+      "dataset_semantics": [
+        {
+          "description": "Run number.",
+          "type": "UShort_t",
+          "variable": "Irun"
+        },
+        {
+          "description": "Event number",
+          "type": "Int_t",
+          "variable": "Ievnt"
+        },
+        {
+          "description": "Tokio multihadron selection flag, see primer , flag ITKMH=1 indicates event is TKMH, =0 otherwise",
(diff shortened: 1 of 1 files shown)


data/records/jade-analysis-example.json
```diff
@@ -0,0 +1,1063 @@
+[
+  {
+    "abstract": {
+      "description": ""
```
tiborsimko, line 4: The description seems missing.

data/records/jade-analysis-example.json
```diff
@@ -0,0 +1,1063 @@
...
+      {
+        "description": "",
+        "type": "Char_t",
+        "variable": "Ichgh[2004]"
+      }
+    ],
+    "date_published": "2021",
+    "distribution": {
+      "formats": [
+        "C",
```
tiborsimko, line 1039: @andriish @rdebrand Where can I find the 9 files to attach to this record?

data/records/jade-analysis-example.json
```diff
@@ -0,0 +1,1063 @@
...
+      "size": 73681875
+    },
+    "experiment": [
+      "JADE"
+    ],
+    "publisher": "CERN Open Data Portal",
+    "recid": "26998",
+    "title": "Examples of basic analysis techniques for the JADE detector",
+    "type": {
+      "primary": "Software",
```
tiborsimko, line 1054: I have changed this to `Software -> Analysis` rather than `Dataset -> Derived`, based on the record title and the C file format of the example. Do you plan to release the code together with the derived dataset? Could there be more code examples for the given derived data? If so, we could  make two separate records, one for software one for data.  (This is what we usually do.)

data/records/jade-analysis-example.json
andriish: LEt us release things as they are and add the files later.
(We can optionally drop the record till that time)

tiborsimko: @andriish OK, thanks! We'll release JADE data on QA and this PR can therefore stay here in the draft mode. (We can update it and/or close it.) Please get in touch later WRT new examples and a suitable release announcement.

PR #3691, opened 2024-10-18
alice: new dataset LHC10h/000137135

psaiz:


data/records/alice-2010-pbpb.json
```diff
@@ -0,0 +1,69 @@
...
+    ],
+    "files": [
+      {
+        "checksum": "adler32:7ac174b9",
+        "size": 27508,
+        "type": "index.json",
+        "uri": "root://eospublic.cern.ch//eos/opendata/alice/upload/2010/LHC10h/000139510/file-indexes/LHC10h_000139510_PbPb_2.76TeV_205656_file_index.json"
+      }
+    ],
+    "license": {
```
tiborsimko, line 41: For heavy-ion physics data, we should add also the keyword information:

```json
    "keywords": [
      "heavy-ion physics"
    ],
```

data/records/alice-2010-pbpb.json
```diff
@@ -0,0 +1,69 @@
...
+      "attribution": "CC0-1.0"
+    },
+    "publisher": "CERN Open Data Portal",
+    "recid": "1114",
+    "title": "LHC10h_000137135_PbPb_2.76TeV_213750",
```
tiborsimko, line 46: Note that this does not look consistent with the old existing titles such as "LHC2010h_PbPb_ESD_138275". Shall we harmonise them?

data/records/alice-2010-pbpb.json
```diff
@@ -0,0 +1,69 @@
...
+    "title_additional": "Lead-Lead data sample at the collision energy of 2.76TeV from run number 137135 of the LHC10h data taking period",
```
tiborsimko, line 47: For the former ALICE records, the description used to say "Pb-Pb data sample" instead of "lead". Another cosmetic observation for harmonising.

data/records/alice-2010-pbpb.json
```diff
@@ -0,0 +1,69 @@
...
+    "type": {
+      "primary": "Dataset",
+      "secondary": [
+        "Collision"
+      ]
+    },
+    "updated": "2025-02-11T15:53:44.819508",
```
tiborsimko, line 54: The `updated` field should not be present in the record JSON.

data/records/alice-2010-pbpb.json
```diff
@@ -0,0 +1,69 @@
...
+    "usage": {
+      "description": "You can access and analyse these data through the ALICE Virtual Machine. Please follow the instructions for getting started and setting up the Virtual Machine:",
```
tiborsimko, line 56: Note that ALICE VM is not working well due to those microCERNVM format changes. (The container is still fine.) Will there be a new VM (that will also work for the past data)?

data/records/alice-2010-pbpb.json
```diff
@@ -0,0 +1,69 @@
...
+      "name": "ALICE Collaboration"
+    },
+    "collections": [
+      "ALICE-Reconstructed-Data"
+    ],
+    "collision_information": {
+      "energy": "2.76TeV",
+      "type": "PbPb"
+    },
+    "created": "2025-02-11T15:53:44.819508",
```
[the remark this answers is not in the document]
adriansev: ack

data/records/alice-2010-pbpb.json
```diff
@@ -0,0 +1,69 @@
+[
```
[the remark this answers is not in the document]
adriansev: The current name is `LHC10h_000139510_PbPb_7.0TeV_205656_record.json`
I've been told that there will be a single run per record.
Let me know it there is something to change.

data/records/alice-2010-pbpb.json
adriansev: what would be the algorithm behind this? anything that is not pp then it's heavy-ion?
how will be the treatment for Xe-Xe, O-O, or p-Xe or p-O?

data/records/alice-2010-pbpb.json
adriansev: I was pointed to such a name: `{PERIOD}_{RUN}_{BEAMTYPE}_{ENERGY}_{COLLISIONS_NR}`
This data is converted data to O2 ALICE framework, there are no longer ESD/AOD present, only AO2D.root files. Let me know if different name is required and with what content.

data/records/alice-2010-pbpb.json
adriansev: Ok, let's get back to symbolic name. How should pp look in this phrase? `Pp`, `PP`, or `pp` ?

adriansev:
this is no longer a target

PR #3693, opened 2024-10-25
records: add TOTEM 2011 dataset and analysis example

tiborsimko:


data/records/totem-datasets-2011.json
```diff
@@ -0,0 +1,195 @@
...
+      "2011"
+    ],
+    "date_published": "2024",
+    "distribution": {
+      "formats": [
+        "root",
+        "NTuple"
+      ],
+      "number_events": 12606699,
+      "number_files": 1,
```
tiborsimko, line 155: @grzanka Note that your record mentioned 2 files, but actually there was only one `prepared_data.root` file, so I have changed this to 1.

data/records/totem-datasets-2011.json
```diff
@@ -0,0 +1,195 @@
...
+    "publisher": "CERN Open Data Portal",
+    "recid": "84000",
+    "title": "2011 Proton scattering data from Roman Pots at energy sqrt(s) = 7 TeV",
+    "type": {
+      "primary": "Dataset",
+      "secondary": [
+        "Derived"
+      ]
+    },
+    "usage": {
```
tiborsimko, line 181: I have put all the `use_with` part together with `usage` which is the default practice for dataset records.

data/records/totem-datasets-2011.json
grzanka: Yes, thats correct, we have one file.

data/records/totem-datasets-2011.json
grzanka: Fine with me

grzanka (approved):
Looks good to me

PR #3696, opened 2024-11-07
WIP: GT db record for 2017 MC

### LINKED ISSUE #3695: CMS: Add a record for the 2017 MC condition GT db
The GT is the same for the 2017 collision data. There is a different GT for MC data. Therefore a new record is needed.

tpmccauley:
* Create a record for the 2017 MC GT db
* Additionally create a record for the 2017 collision GT db
* Fix a typo in the 2016 MC GT db record
* Closes #3695

One question to resolve is whether or not to have a separate record for the 2016 and 2017 collision GT db when they point to the same file. The other option would be to make a new record for the GT MC db, modify the GT for collision data to include references to 2017, adding a new link. 

The XXXX and YYYY in the json file refer to the as-yet-undefined record ids

### DIFF
--- a/data/records/cms-condition-data-2017.json
+++ b/data/records/cms-condition-data-2017.json
@@ -1,10 +1,10 @@
 [
   {
     "abstract": {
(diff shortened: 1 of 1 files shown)


jmhogan (approved):
Looks good to me!

Issue #3111, opened 2021-05-07
records: standardise license strings


tiborsimko:
Amend record fixtures to standardise on some widely-used license strings. Example: [SPDX license list](https://spdx.org/licenses/) that is used by [F-UJI](https://f-uji.net/) would help to improve F-UJI's FAIR assessment findings.

tiborsimko:
The current distribution of license values seems to be:

```
   12420 CC0
    6052 null
     46 GNU General Public License (GPL) version 3
      2 Apache License, Version 2.0
      1 MIT Open Source license
      1 MIT License (MIT)
      1 CC0 license
      1 BSD License
```

After we update the values to SPDX, we'd also need to amend the portal infrastructure repository's output templates to display the CC0 copyright message accordingly.

Biscgit:
Created adjustments for the Portal already regarding the upcoming license changes. Please follow PR in cernopendata/cernopendata-portal#83 on that.

psaiz:
The PR on the portal is there. The one of opendata.cern.ch is still open

PR #3697, opened 2024-11-07
consistency: adjusted licenses to consistent names

Biscgit:


data/records/cms-learning-resources.json
```diff
@@ -144,7 +144,7 @@
       "education"
     ],
     "license": {
-      "attribution": "MIT Open Source license"
+      "attribution": "MIT License"
```
tiborsimko, line 147: It would be good to take the occasion of licence value clean-up and move towards the SPDX standard license values everywhere. This will help to improve FAIR score in tools like FUJ-I. We'll need small amendments in the  Jinja templates accordingly as well. Please see #3111 for all details.

Biscgit: Ah, I just saw your issue. Do you have something like this in mind on how to improve the annotation on the license attributions:
```
"license": {
    "identifier": ...
    "full_name": ...
}
```

tiborsimko: For simplicity, it would be good to stick to the current data model, using the same `license.attribution` field, b with values containing standardised license identifiers (such as `CC0-1.0`). And, in order to display the full name, we could have a Jinja helper function on the infrastructure side translating identifiers to the standardised full license names (such as `Creative Commons Zero v1.0 Universal`) following the SPDX nomenclature.  (We could perhaps even make a link to the license in question.) And, to be on the safe side, for any new licence string we may introduce in thefuture that the Jinja helper function does not recognise, we would display just the text.

Since we are using only a handful of licenses, this approach should do, and has the merit of keeping the same data model.

Biscgit:
It made more sense to move development focus to the standardizing directly. Follow #3111 for further updates.

PR #3703, opened 2024-12-10
feat(docs): Ten years of CERN Open Data portal

tiborsimko:


psaiz:
This PR would make more sense on the cernopendata_portal repo. Note that the cernopendata_portal contains the banner (which points to this document). If we keep those two things in separate repos, that means the cernopendata_portal is no longer self-contained. 

Note also that the other documents about the portal are already in the cernopendata_portal repo.

tiborsimko:
Moved to https://github.com/cernopendata/cernopendata-portal/pull/112

data/docs/ten-years-of-cern-open-data-portal/ten-years-of-cern-open-data-portal.json
```diff
@@ -0,0 +1,19 @@
...
+    "author": "CERN Open Data team",
+    "body": {
+      "content": "ten-years-of-cern-open-data-portal.md",
+      "format": "md"
+    },
+    "date_published": "2024-12-10",
+    "short_description": {
+      "content": "The CERN Open Data portal celebrates its ten year birthday! Find out about its journey and today's challenges."
+    },
+    "featured": 1,
```
tiborsimko, line 12: Do we need `"featured"`` still? 

Some time ago, we have discussed that we don't really feature old articles, so the portal can simply display last 3 (or last 6) news. This should perfectly do, and we could remove "featured" from the content repository articles too. Was this implemented already?

PR #3709, opened 2024-12-13
Adding Heavy Ion Physics tag to HI records

zlmarshall:
For the two new HI records, we should make sure the Heavy Ion tag is included :)

### DIFF
--- a/data/records/atlas-hi-2024-hi-2015-data.json
+++ b/data/records/atlas-hi-2024-hi-2015-data.json
@@ -4,9 +4,6 @@
       "description": "Run 2 2015 Pb-Pb collision data from the ATLAS experiment"
     },
     "accelerator": "CERN-LHC",
-    "categories": {
-      "source": "ATLAS Collaboration"
-    },
     "collaboration": {
       "name": "ATLAS collaboration"
     },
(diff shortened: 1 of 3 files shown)


data/records/atlas-hi-2024-hi-2015-data.json
```diff
@@ -5,6 +5,7 @@
     },
     "accelerator": "CERN-LHC",
     "categories": {
+      "primary": "Heavy-Ion Physics",
```
tiborsimko, line 8: Note that the "categories" field is used with the simulated data only:
https://github.com/cernopendata/cernopendata-portal/blob/main/cernopendata/jsonschemas/records/record-v1.0.0.json#L107

Hence this edit should rather go to the MC record, not to the data record.

tiborsimko: Also, we have the data-neutral keywoard "heavy-ion physics" that people select in the faceted search:  https://opendata.cern.ch/search?q=&f=keywords%3Aheavy-ion%20physics&l=list&order=desc&p=1&s=10&sort=mostrecent

So that we could add to all three records the following snippet:

```json
    "keywords": [
      "heavy-ion physics"
    ],
```

If it's OK with you, I can make these changes whilst merging.

zlmarshall: It is definitely ok with me — thanks!

PR #3711, opened 2024-12-19
feat(records): add DOIs to DELPHI datasets and manuals

tiborsimko:


schwicke (approved):
LGTM. 
Maybe ask Dietrich as well for another pair of eyes.

Issue #3717, opened 2025-01-07
DELPHI: Fix entries with the same base file name


psaiz:
Look at entries like 91396. It uses the same base file name several times: 
```
sh-5.1$ cat /tmp/my_delphi  |grep uri |grep wphact_206.5_map.xsdst.99
        "uri": "root://eospublic.cern.ch//eos/opendata/delphi/simulated-data/live/wphact20/va0u/206.5/job1/wphact_206.5_map.xsdst.99"
        "uri": "root://eospublic.cern.ch//eos/opendata/delphi/simulated-data/live/wphact20/va0u/206.5/job2/wphact_206.5_map.xsdst.99"
        "uri": "root://eospublic.cern.ch//eos/opendata/delphi/simulated-data/live/wphact20/va0u/206.5/job3/wphact_206.5_map.xsdst.99"
        "uri": "root://eospublic.cern.ch//eos/opendata/delphi/simulated-data/live/wphact20/va0u/206.5/job4/wphact_206.5_map.xsdst.99"
        "uri": "root://eospublic.cern.ch//eos/opendata/delphi/simulated-data/live/wphact20/va0u/206.5/job5/wphact_206.5_map.xsdst.99"
```
Note that from the [record url](https://opendata-qa.cern.ch/record/91396), the base filename is used as a key to identify which file should be downloaded:
https://opendata-qa.cern.ch/record/91396/files/wphact_206.5_map.xsdst.99
That means that only one out of the five entries can currently be downloaded from the web page. 

In order to solve this ticket, all the DELPHI entries that have this issue should be solved. The first three options that came to mind are:
1. Move from file to file_indices
2. Rename the files to ensure that they use different base file names
3. Remove the constraint on the basename, and use a different key (this might be tricky)

PR #3718, opened 2025-01-27
fix(records): change DELPHI record 91396 to use file indices

psaiz:
…ries directly attached to the record

Closes [#3717](https://github.com/cernopendata/opendata.cern.ch/issues/3717)


tiborsimko (approved):
Works nicely :+1: 

This change has been checked with @dietrichliko and @schwicke. The motivation being that the given MC record contains many individual files with the same base name, only stored in different subdirectories, which is something that the portal infrastructure currently does not support. And, the file-index based browsing is actually useful for an MC record consisting of so many files anyway.

PR #3721, opened 2025-02-19
ALICE :: add test record for LCH10h/139510

adriansev:
@tiborsimko @psaiz 
Done the PR as we talked about .. let me know what other details are to be fixed


adriansev:
this is no longer a target

PR #3722, opened 2025-03-05
ci(run-tests): fix data JSON checker error reporting

tiborsimko:
Fixes a problem with the CI data JSON checker that was not reporting proper error code when checking multiple files.

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -36,7 +36,7 @@ data_dois() {
 }
 
 data_json() {
-    find . -name "*.json" -exec ./scripts/clean_json_file.py --check {} \;
+    find . -name "*.json" -exec ./scripts/clean_json_file.py --check {} \+
 }
 
 data_licenses() {
--- a/scripts/clean_json_file.py
+++ b/scripts/clean_json_file.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python3
 
-"""Clean JSON file.
+"""
+Clean JSON files.
 
-Cleans JSON file so as to fix indentation, order fields, and remove trailing
-whitespace.
+Clean one or more JSON files so as to fix indentation, order fields, and remove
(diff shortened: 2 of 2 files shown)


run-tests.sh
```diff
@@ -36,7 +36,7 @@ data_dois() {
 }
 
 data_json() {
-    find . -name "*.json" -exec ./scripts/clean_json_file.py --check {} \;
+    find . -name "*.json" -exec ./scripts/clean_json_file.py --check {} \+
```
psaiz, line 39: This will pass all the files in a single command, isn't it? It is better than the previous version. At the same time, wouldn't it be even better if the script accepted a directory, and it would go through all the json files in that directory?

tiborsimko: That could be another option. I went for accepting multiple file names instead, just like `jsonlint` (or `shellcheck`) are doing...

scripts/clean_json_file.py
```diff
@@ -13,46 +14,44 @@
...
-            print(f"[ERROR] File {filename} is badly formatted.")
-            sys.exit(1)
-        else:
-            with open(filename, "w") as fdesc:
-                fdesc.write(new_content)
+
+        if old_content != new_content:
+            if check:
+                print(f"[ERROR] File {filename} is badly formatted.")
+                sys.exit(1)
```
psaiz, line 51: Imagine that there are 10 files failing. The first execution of the script will only show the first file, and exit. Then, the user will fix that one, run the script again and it will fail in the next one. 
Wouldn't it be better if it gave the list of all the badly formatted files on the first execution? That would be consistent with the pytest

tiborsimko: Good point. Fixed.

psaiz (approved):

PR #3728, opened 2025-04-12
Add delphi online logbooks

schwicke:
Create a record for DELPHI logbooks.


data/records/delphi-documentation-logbooks.json
```diff
@@ -0,0 +1,233 @@
...
+      "name": "DELPHI collaboration",
+      "recid": "80050"
+    },
+    "collections": [
+      "DELPHI-documentation-manuals"
+    ],
+    "date_created": [
+      "2000"
+    ],
+    "date_published": "2024",
```
tiborsimko, line 17: 2025

data/records/delphi-documentation-logbooks.json
```diff
@@ -0,0 +1,233 @@
...
+    "distribution": {
+      "formats": [
+        "pdf"
+      ],
+      "number_files": 39,
```
tiborsimko, line 22: Do we want to have one record only for all 39 logbooks? Or would you like to split logbooks into individual records by years? It might be even easier for the search facet functionality, e.g. if people select "DELPHI" and "1991", they will find immediately corresponding data and logbooks for 1991 only.

This is what we did for JADE logbooks: https://opendata-qa.cern.ch/search?q=&f=experiment%3AJADE&f=type%3ASupplementaries%2Bsubtype%3ALogbook&l=list&order=desc&p=1&s=10&sort=mostrecent

data/records/delphi-documentation-logbooks.json
```diff
@@ -0,0 +1,233 @@
...
+    "license": {
+      "attribution": "CC0-1.0"
+    },
+    "publisher": "CERN Open Data Portal",
+    "recid": "80508",
+    "title": "DELPHI online logbooks",
+    "type": {
+      "primary": "Supplementaries",
+      "secondary": [
+        "Manual"
```
tiborsimko, line 229: The field `type` should be `Supplementaries -> Logbook`, for parity with JADE.

data/records/delphi-documentation-logbooks.json
```diff
@@ -0,0 +1,233 @@
...
+    "accelerator": "CERN-LEP",
+    "collaboration": {
+      "name": "DELPHI collaboration",
+      "recid": "80050"
+    },
+    "collections": [
+      "DELPHI-documentation-manuals"
+    ],
+    "date_created": [
+      "2000"
```
tiborsimko, line 15: We should probably include all the years in `date_created`, since this field corresponds to collision data taking. There can be multiple values, so we can list 1991, 1992, etc.  This will be good for search facets.

data/records/delphi-documentation-logbooks.json
schwicke: Changed from manual to Logbook.

data/records/delphi-documentation-logbooks.json
schwicke: Yeah, it may be useful to do so. We would still have 4 logbooks in single record for each year which may be good enough.

schwicke: OK, done. By year, as suggested. This way, searching by year, they will get 1, 4 or 5 logbooks, depending on the year.

psaiz (approved):

PR #3730, opened 2025-04-14
fix(records): add missing file_index.json

psaiz:
WIP:  note that the last one still misses the size and checksum


tiborsimko (approved):
Thanks, rebased against latest `master` and slightly amended log message.

PR #3739, opened 2025-05-12
fix(records): add record for DELPHI conditions database

schwicke:
The conditions database is currently still missing a record entry but needed by some of our tools. This patch adds a reference to it.

### DIFF
--- a/data/records/delphi-conditions-database.json
+++ b/data/records/delphi-conditions-database.json
@@ -1,0 +1,79 @@
+[
+  {
+    "abstract": {
+      "description": "The DELPHI conditions database consists of 7 different files, storing distinct information about the run condition during the different periods. The format is proprietary, and there are tools available to read it both programmatic from Fortran and interactively via the commands `dplot` and applications like `dtobin` or `dlist`.  The database contains information about the run conditions, including background (raw) delivered luminosity, meteorological information, fill information and calibration data for the different detector components.<p>The conditions database is needed for event reconstruction, as provided by the DELPHI event server. It is not needed for data analysis or for simulation. The latter uses yearly snapshot databases instead, which ships with the corresponding simulation version.<p> Manuals, including usage examples, are available from the CERN document server, in particular the [CARGO Database Management Package](https://cds.cern.ch/record/2625473/files/delphi-93-5_ocr.pdf), [DPLOT Database graphics](https://cds.cern.ch/record/2630031/files/96_139_das_180_prog_219.pdf) and the [DELPHI Detector Description Application Package user manual](https://cds.cern.ch/record/2629009/files/delphi-90-37_ocr.pdf)."
+    },
+    "accelerator": "CERN-LEP",
+    "collaboration": {
+      "name": "DELPHI collaboration",
+      "recid": "80050"
+    },
+    "collection": [
+      "DELPHI-Condition-Data"
+    ],
+    "date_created": [
+      "2000"
+    ],
+    "date_published": "2025",
+    "distribution": {
+      "formats": [
+        "dat"
+      ],
+      "number_files": 7,
+      "size": 3205865472
+    },
+    "doi": "10.7483/OPENDATA.DELPHI.18JY.GL8K",
+    "experiment": [
+      "DELPHI"
+    ],
+    "files": [
+      {
+        "checksum": "adler32:18ab9196",
+        "size": 1497251840,
+        "uri": "root://eospublic.cern.ch//eos/opendata/delphi/condition-data/DBcalb.dat"
+      },
+      {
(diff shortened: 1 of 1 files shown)


data/records/delphi-conditions-database.json
```diff
@@ -0,0 +1,75 @@
+[
+  {
+    "abstract": {
+      "description": "The DELPHI conditions database."
```
tiborsimko, line 4: The record looks currently like:

image

The abstract is a bit short, so casual visitors may be perplexed.

Can you please extend the text and perhaps add some usage information and links to docs etc? You could for example explain what `DBcalb.dat` means, what `DBsysf.dat` means, etc.

Example from CMS: https://opendata.cern.ch/record/1818 with the sources in `data/records/cms-condition-data-2016.json`.

data/records/delphi-conditions-database.json
```diff
@@ -0,0 +1,75 @@
...
+    "collections": [
+      "DELPHI"
+    ],
+    "date_created": [
+      "2000"
+    ],
+    "date_published": "2025",
+    "distribution": {
+      "formats": [
+        "zebra"
```
tiborsimko, line 20: The `distribution.formats` field is technical and usually indicates the file extension. I would use `dat` here, and explain about Zebra in the abstract?

data/records/delphi-conditions-database.json
```diff
@@ -0,0 +1,75 @@
...
+    "abstract": {
+      "description": "The DELPHI conditions database."
+    },
+    "accelerator": "CERN-LEP",
+    "collaboration": {
+      "name": "DELPHI collaboration",
+      "recid": "80050"
+    },
+    "collections": [
+      "DELPHI"
```
tiborsimko, line 12: The value should rather be `DELPHI-Condition-Data` to have parity with CMS.

(Note also singular instead of plural.)

data/records/delphi-conditions-database.json
```diff
@@ -0,0 +1,75 @@
...
+        "size": 131072,
+        "uri": "root://eospublic.cern.ch//eos/opendata/delphi/condition-data/DBsysf.dat"
+      }
+    ],
+    "license": {
+      "attribution": "CC0-1.0"
+    },
+    "publisher": "CERN Open Data Portal",
+    "recid": "80509",
+    "title": "DELPHI conditions database",
```
tiborsimko, line 70: Cosmetics: Do you consider the individual files as one single database, or should we use "databases" (in plural)?

data/records/delphi-conditions-database.json
```diff
@@ -0,0 +1,75 @@
...
+    "type": {
```
tiborsimko, line 71: The `type` field should rather be Environment -> Condition, for parity with CMS data:

```json
    "type": {
      "primary": "Environment",
      "secondary": [
        "Condition"
      ]
    },
```

data/records/delphi-conditions-database.json
```diff
@@ -0,0 +1,75 @@
...
+        "checksum": "adler32:1d8e2a3b",
+        "size": 131072,
+        "uri": "root://eospublic.cern.ch//eos/opendata/delphi/condition-data/DBsysf.dat"
+      }
+    ],
+    "license": {
+      "attribution": "CC0-1.0"
+    },
+    "publisher": "CERN Open Data Portal",
+    "recid": "80509",
```
tiborsimko, line 69: If you'd like to give it a DOI, we can use `10.7483/OPENDATA.DELPHI.18JY.GL8K`.

(Assuming one condition data record aggregating all the DAT files is enough. Alternatively, if you'd like to have distinct DOIs for distinct files, to be able to reference them independently, we could also create independent records for each DBsysf etc entity.)

data/records/delphi-conditions-database.json
schwicke: I prefer to use the plural. The code needed to read that stuff usually wants to have all of them.

data/records/delphi-conditions-database.json
schwicke: Makes sense. Fixed. Thanks!

data/records/delphi-conditions-database.json
```diff
@@ -0,0 +1,75 @@
...
+    "title": "DELPHI conditions database",
+    "type": {
+      "primary": "Supplementaries"
+    }
+  }
+]
```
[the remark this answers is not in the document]
schwicke: Looks like I never properly answered this one: I have fixed the commit message a while ago, and squashed the commits (after having resolved the other conversations) to make the history more clean. 

Please take a second look at the final version.

Issue #3745, opened 2025-06-25
enrich CMS RAW data records with run number information


tiborsimko:
### Current behaviour

For CMS collision records, we have `run_numbers` field that lists all the data-taking run numbers for the given dataset.

Example: [record 1](https://opendata.cern.ch/record/1)

```console
$ grep -A 10 '"recid": "1"' data/records/cms-primary-datasets.json
    "recid": "1",
    "relations": [
      {
        "title": "/BTau/Run2010B-v1/RAW",
        "type": "isChildOf"
      }
    ],
    "run_numbers": [
      "146240",
      "146242",
      "146248",
```

However, the CMS RAW data samples do not have this information.

Example: [record 65](https://opendata.cern.ch/record/65)

```console
$ grep -A 10 '"recid": "65"' data/records/cms-raw-datasets-Run2012B.json
    "recid": "65",
    "run_period": [
      "Run2012B"
    ],
    "system_details": {
      "container_images": [
        {
          "name": "docker.io/cmsopendata/cmssw_5_3_32-slc6_amd64_gcc472:latest",
          "registry": "dockerhub"
        },
        {
```

The only information about run numbers is present in the abstract, see "Run range [194439,194691]" there.

### Expected behaviour

It would be good to add the run number information to all CMS RAW sample records as well:

- `data/records/cms-raw-datasets-Run2010B.json`
- `data/records/cms-raw-datasets-Run2011A.json`
- `data/records/cms-raw-datasets-Run2012B.json`

CC @IssaAlBawwab @nasseralbess @katilp @tpmccauley

PR #3752, opened 2025-07-04
enriching: added run_numbers field for the cms datasets (#3752)

nasseralbess:
I retrieved the run numbers for each dataset and updated the fields as mentioned in issue #3745


katilp (commented):
Thanks! Looks good.
Would you just check recid 35? It is a bit unlikely that it only contains events from one run number.

nasseralbess:
Alright, pushed again without run_numbers for record 35.

PR #3763, opened 2025-11-07
fix(records): change the link of the ATLAS software tutorial

psaiz:


zlmarshall (approved):
That's the correct link — thank you for the help!

PR #3772, opened 2025-12-15
build(docker): upgrade cernopendata-portal to 0.13.0

tiborsimko:
Upgrade CERN Open Data portal version to 0.13.0 and change local
development server to use port 8000 in order to avoid conflicts with
port 5000 on macOS.

Additionally, plug the MQ service back to avoid unnecessary slow-downs
when accessing the detailed record pages. The MQ is needed by the
invenio-stats module that came in as part of the upgrade.

### DIFF
--- a/DEVELOPING.rst
+++ b/DEVELOPING.rst
@@ -45,7 +45,7 @@ follows:
 
    $ git clone https://github.com/cernopendata/opendata.cern.ch
    $ git clone https://github.com/cernopendata/cernopendata-portal
-   $ cd cernopendata-portal && git checkout -B stable v0.2.9
+   $ cd cernopendata-portal && git checkout -B release-v0.13.0 v0.13.0
    $ cd ../opendata.cern.ch
    $ docker compose pull
    $ docker compose up -d
@@ -55,7 +55,7 @@ follows:
 
 This will create a running instance of the CERN Open Data portal with a
 relatively empty content. The portal will be accessible locally at
-`http://127.0.0.1:500 <http://127.0.0.1:5000>`_.
+`http://127.0.0.1:8000 <http://127.0.0.1:8000>`_.
 
 If you would like to stop and delete your local instance, you can do:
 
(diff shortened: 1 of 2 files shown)


DEVELOPING.rst
```diff
@@ -45,7 +45,7 @@ follows:
 
    $ git clone https://github.com/cernopendata/opendata.cern.ch
    $ git clone https://github.com/cernopendata/cernopendata-portal
-   $ cd cernopendata-portal && git checkout -B stable v0.2.9
+   $ cd cernopendata-portal && git checkout -B stable v0.10.2
```
tiborsimko: Note: Ideally we should do this only from v0.10.3 (or even better v0.11.0 to indicate a breaking change of sorts) when the portal will have port 8000 changes merged.

DEVELOPING.rst
```diff
@@ -45,7 +45,7 @@ follows:
 
    $ git clone https://github.com/cernopendata/opendata.cern.ch
    $ git clone https://github.com/cernopendata/cernopendata-portal
-   $ cd cernopendata-portal && git checkout -B stable v0.2.9
+   $ cd cernopendata-portal && git checkout -B stable v0.11.0
```
psaiz: What about using the tag `stable` (that I just created on that repo) ? That way, the readme doesn't have to be modified to point to the latest stable release

tiborsimko: The trouble with this approach is that "stable" would be volatile, so if we upgrade portal code in an incompatible way, the user would not know that "stable" needs to be ported to a "new stable"... Recall the troubles we had with the old "prod" tag and why we introduced semantic versioning, see https://github.com/cernopendata/opendata.cern.ch/pull/3679

Using semantic versions don't have this problem, because they are never mutable. See also https://vsupalov.com/docker-latest-tag/

If the terminology "stable" seems off, I'll rather rename the branch name to "v0.13.0" in the developer docs, i.e. equal to the tag... That might be the safest/cleanest.

PR #3787, opened 2026-05-11
add json for lhcb omegac2xick opendata example

pietnogga:
Sceleton for the new lhcb opendata example OmegaC2XiCK. I took the example json from https://github.com/cernopendata/opendata.cern.ch/blob/master/data/records/lhcb-ntuples-run2-b2jpsik.json and changed (with codex's help) some fields to match the new example.

Some fields are still left open, but the number of root files per job should be correct. Please let me know if there is something missing or you still need from my side

### DIFF
--- a/data/records/lhcb-ntuples-run2-omegac2xick.json
+++ b/data/records/lhcb-ntuples-run2-omegac2xick.json
@@ -1,0 +1,831 @@
+[
+  {
+    "abstract": {
+      "description": "<p>Data from proton-proton (pp) collisions collected by the LHCb experiment filtered to produce ntuples for exploring $\\Omega_c^0 \\to \\Xi_c^+ (\\to K^- p \\pi^+) K^-$ and charge-conjugate decays.</p> <p>Ntuples are created using DaVinci version v46r11 and the Analysis Productions batch processing system. Quantities saved to the ntuple are specified during the request phase and detailed in the code configuration files attached below.</p>"
+    },
+    "accelerator": "CERN-LHC",
+    "collaboration": {
+      "name": "LHCb collaboration"
+    },
+    "collections": [
+      "LHCb-Collision-Datasets"
+    ],
+    "date_created": [
+      "2015",
+      "2016",
+      "2017"
+    ],
+    "date_published": "2026",
+    "date_reprocessed": "2026",
+    "distribution": {
+      "formats": [
+        "py",
+        "yaml"
+      ],
+      "number_files": 3,
+      "size": 6081
+    },
+    "doi": "10.7483/OPENDATA.LHCB.MHK4.OQO3",
+    "experiment": [
+      "LHCb"
+    ],
+    "files": [
+      {
+        "checksum": "adler32:62f202c9",
+        "size": 761,
+        "uri": "root://eospublic.cern.ch//eos/opendata/lhcb/CollisionNtuples/OPENDATA.LHCB.MHK4.OQO3/inputs/OmegaCOpenDataTest/DecayTree.py"
+      },
+      {
+        "checksum": "adler32:a4755ea8",
+        "size": 3980,
+        "uri": "root://eospublic.cern.ch//eos/opendata/lhcb/CollisionNtuples/OPENDATA.LHCB.MHK4.OQO3/inputs/OmegaCOpenDataTest/DecayTree.yaml"
+      },
+      {
+        "checksum": "adler32:33c3949f",
+        "size": 1340,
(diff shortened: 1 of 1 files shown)


data/records/lhcb-ntuples-run2-omegac2xick.json
```diff
@@ -0,0 +1,837 @@
...
+    "date_published": "2026",
+    "date_reprocessed": "2026",
+    "distribution": {
+      "formats": [
+        "root"
+      ],
+      "number_files": 3,
+      "size": "TODO_TOTAL_SIZE_BYTES"
+    },
+    "doi": "TODO_2015_MAGDOWN_DOI",
```
tiborsimko: @dillfitz @pietnogga Note that for the b2jpsik example, we decided to introduce DOI only to the parent record, not to daughter records. We should harmonise this one way or another. Do you prefer to have one DOI per request? Or one DOI per request and N DOIs for each ntuple?

data/records/lhcb-ntuples-run2-omegac2xick.json
```diff
@@ -0,0 +1,837 @@
...
+        "recid": "TODO_PARENT_RECID",
+        "type": "isChildOf"
+      }
+    ],
+    "run_period": [
+      "2015"
+    ],
+    "stripping": {
+      "stream": "CHARM.MDST",
+      "version": "Stripping24r2"
```
tiborsimko: Cosmetics: The stripping versions are usually normalised to all-lowercase, even if LHCb uses uppercase S. Since we have a lot of released data, let's amend the scripts to downcase stripping version information.

data/records/lhcb-ntuples-run2-omegac2xick.json
```diff
@@ -0,0 +1,837 @@
...
+        "checksum": "TODO_2015_MAGUP_ROOT_FILE_2_CHECKSUM",
+        "size": "TODO_2015_MAGUP_ROOT_FILE_2_SIZE_BYTES",
+        "uri": "TODO_2015_MAGUP_ROOT_FILE_2_URI"
+      }
+    ],
+    "magnet_polarity": "MagUp",
+    "methodology": {
+      "description": "<p>These ntuples were produced from the LHCb Open Data Ntupling Service user request TODO_REQUEST_ID. Please see the input configuration files <a href=\"/record/TODO_PARENT_RECID\">here</a>.</p>"
+    },
+    "publisher": "CERN Open Data portal",
```
tiborsimko: Cosmetics: the usual capitalisation of the `publisher` field value is "CERN Open Data Portal".

data/records/lhcb-ntuples-run2-omegac2xick.json
```diff
@@ -0,0 +1,837 @@
...
+        {
+          "description": "LHCb Open Data Ntupling Service",
+          "url": "https://opendata-lhcb-ntupling-service.app.cern.ch/"
+        },
+        {
+          "description": "LHCb Open Data Ntupling Service Usage Guide",
+          "url": "https://lhcb-opendata-guide.web.cern.ch/ntupling-service/"
+        },
+        {
+          "description": "methodology",
```
tiborsimko: The `description` field should probably look like: 

```json
         {
            "description": "OmegaC2XicK analysis example",
            "url": "TODO_REANA_ANALYSIS_EXAMPLE_URL"
          }
```

or:

```json
        {
          "description": "Searching for resonant OmegaC+ states in OmegaC2XicK",
          "url": "TODO_REANA_ANALYSIS_EXAMPLE_URL"
        }
```

data/records/lhcb-ntuples-run2-omegac2xick.json
dillfitz: My vote is to keep it consistent and just have one DOI for the parent record. I can't remember if I followed that structure but the same applies for the dijet PR. I think this would be easier for the user.

pietnogga: Good catch. The record now only has one placeholder parent DOI

data/records/lhcb-ntuples-run2-omegac2xick.json
pietnogga: done

data/records/lhcb-ntuples-run2-omegac2xick.json
pietnogga: Ah yes good catch, changed to your option 1

tiborsimko (approved):

PR #3789, opened 2026-06-27
feat(records): add three LHCb author lists for LHC Run 2

dillfitz:
I have converted the 2015-2017 LHCb author lists from December 31 of each year to the expected schema. This relies on querying the inspirehep database to get the inspire ID. 

Currently the queries are only successful if an ORCID is supplied with the author on the author list, in which case the name, inspire ID, and affiliations are gathered from the inspire DB. Otherwise the name and affiliations are extracted directly from the author list. 

Some extra work could be done to make the queries to the inspire DB successful by name in an orcid is not supplied, but this takes a lot of additional effort and can lead to false positives. I am hoping this is "good enough" as it effectively blocks the run 2 metadata release(a PR will be made for this soon).

The record IDs are set to "XYZ" temporarily, but I did confirm that the author lists upload and render properly using the development guide. Once we have firm recids I will need this for the run 2 metadata records to refer to the appropriate author lists. The PR for the run 2 metadata records is here (https://github.com/cernopendata/opendata.cern.ch/pull/3790).

cc @tiborsimko, @pietnogga, @MindaugasSarpis

### DIFF
--- a/data/records/lhcb-author-list-2015.json
+++ b/data/records/lhcb-author-list-2015.json
@@ -1,0 +1,3650 @@
+[
+  {
+    "authors": [
+      {
+        "affiliation": "CERN",
+        "inspireid": "INSPIRE-00258707",
+        "name": "Aaij, Roel"
+      },
+      {
+        "affiliation": "U. Santiago de Compostela (main)",
+        "inspireid": "INSPIRE-00194077",
+        "name": "Adeva, Bernardo"
+      },
+      {
+        "affiliation": "Bristol U.",
+        "inspireid": "INSPIRE-00060950",
+        "name": "Adinolfi, Marco"
+      },
+      {
+        "affiliation": "Liverpool U.",
+        "inspireid": "INSPIRE-00061029",
+        "name": "Affolder, Anthony"
+      },
+      {
+        "affiliation": "Clermont-Ferrand U.",
+        "inspireid": "INSPIRE-00061185",
+        "name": "Ajaltouni, Ziad"
+      },
+      {
+        "affiliation": "Marseille, CPPM",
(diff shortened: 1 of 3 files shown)


dillfitz:
Just a comment that I plan to update the json files with the versions produced by code Mindaugas provided, but if we have 3 recids picked out I can already amend https://github.com/cernopendata/opendata.cern.ch/pull/3790

data/records/lhcb-author-list-2015.json
```diff
@@ -0,0 +1,3390 @@
...
+    "recid": "999",
+    "title": "LHCb author list for the LHCb 2015 open data release",
+    "type": {
+      "primary": "Documentation",
+      "secondary": [
+        "Authors"
+      ]
+    }
+  }
+]
```
tiborsimko: @dillfitz The `data-json` formatting CI test fails only due to trailing newline.

You can fix it by doing:

```console
$ ./scripts/clean_json_file.py data/records/lhcb-author-list-2015.json \
                               data/records/lhcb-author-list-2016.json \
                               data/records/lhcb-author-list-2017.json
```

dillfitz: Thanks! I just pushed a new commit with the updated author lists (using the code Mindaugas provided) and ran this script to clean the author lists before pushing. I think these are all set now aside from the recids, but feel free to let me know if anything needs to be changed.

tiborsimko: Thanks, the record IDs are free, so we can merge "as is". (FYI I have updated commit headline wording in order to pass CI.)

tiborsimko (approved):

Issue #3681, opened 2024-09-21
CMS - wrong units for dataset size in the PFNano dataset records 


katilp:
The [PFNano records](https://opendata.cern.ch/search?q=&f=type%3ADataset&f=file_type%3Ananoaod-pf&l=list&order=desc&p=1&s=10&sort=mostrecent) show dataset size units of Mib/Gib whereas the real size is in Gb/Tb:

```
1.3T    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/BTagCSV
306G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/BTagMu
643G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/Charmonium
215G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/DisplacedJet
853G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/DoubleEG
506G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/DoubleMuon
476G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/DoubleMuonLowMass
384G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/HTMHT
1.4T    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/JetHT
324G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/MET
387G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/MuonEG
362G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/MuOnia
1.7T    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/SingleElectron
1.6T    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/SingleMuon
362G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/SinglePhoton
881G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/Tau
210G    /eos/opendata/cms/derived-data/PFNano/29-Feb-24/ZeroBias
```

Cid-oe:
This issue is addressed in PR #3799 (and #3767), which updates all 17 PFNano records in `data/records/cms-derived-pfnano-2016.json` to their exact byte sizes (matching the file indexes and resolving the factor of 1024 KiB-to-bytes discrepancy).

PR #3795, opened 2026-09-11
fix(records): correct dataset size in bytes for CMS PFNano 2016 records (#3767)

### LINKED ISSUE #3767: Inconsistent dataset size on NanoAOD+ParticleFlow records.
Hello all!

It looks like there's a minor bug in the dataset size for the "NanoAOD format enhanced with Particle Flow candidates" files. I've only checked 2 so far but they both have inconsistent listings for the total dataset size. 

If you look [here](https://opendata.cern.ch/record/31308), under "Dataset characteristics" it says the dataset size is 1.4 Gib (Gigabytes). 

But down where it has the "index description" at the bottom, it says 1.4 Tib (Terabytes), which I think is the correct number.

Cid-oe:
Closes #3767

### Summary
The dataset sizes in `data/records/cms-derived-pfnano-2016.json` for the 17 NanoAOD+ParticleFlow records (`31300` through `31316`) were stored in KiB instead of Bytes (~1,024x smaller than actual total size). As a result, Invenio displayed dataset sizes as ~1.4 GiB instead of ~1.4 TiB.

This PR recalculates and sets the exact byte sum from each record's EOS index JSON (`*root_file_index.json`), matching the file sizes and resolving the inconsistency across all 17 records.

### DIFF
--- a/data/records/cms-derived-pfnano-2016.json
+++ b/data/records/cms-derived-pfnano-2016.json
@@ -39,7 +39,7 @@
       ],
       "number_events": 100826567,
       "number_files": 468,
-      "size": 1297795768
+      "size": 1328942631099
     },
     "experiment": [
       "CMS"
@@ -142,7 +142,7 @@
       ],
       "number_events": 27536603,
       "number_files": 91,
-      "size": 320657946
+      "size": 328353697999
     },
     "experiment": [
       "CMS"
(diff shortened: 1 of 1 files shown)


Cid-oe:
This PR also resolves #3681.

